Issue #24, opened 2022-02-07
Adding 'interactive' for only on:click slows it massively while cursor is hovering


FreddieRa:
I assume this slow down comes from the fact it's doing raytracing every frame, but for `on:click` it would be great if this raytracing was actually only done on mouse press, rather the whole time.

grischaerbe:
There's unfortunately no way I know of to check if a parent component listens to an event, so the moment `interactive` is set on a Mesh, the frameloop is running continuously and raycasting is done on every frame. One way around this would be to set individual flags but I think in reality you quickly run into using pointer events and therefore raycasting on every frame anyway.

One might be tempted to raycast only on interactions of the cursor with the canvas element – let's say on `pointermove`, `pointerenter` and `pointerleave` – but this would not account for your scene changing "underneath" your stationary cursor.

Nevertheless this should not slow down things *massively*. Can you elaborate on that part? You can optimize your scene by adding `ignorePointer` to Meshes that don't need to react to the cursor. This way the raycaster won't raycast on them.

```svelte
<Mesh ignorePointer {geometry} {material} />
```

By the way: I just double checked: It's `ignorePointer` and not `ignorePointerEvents` as the README states, I'll change that.

FreddieRa:
Adding `ignorePointer` to everything else stopped the slow down, thanks for that!

grischaerbe:
Perfect. Be aware that a `<Mesh ignorePointer >` is not only *not* reacting to pointer events, it's also "click-through". So an `interactive` Mesh behind it can actually receive click events. It's much like `pointer-events: none` in a way.

FreddieRa:
Ahh okay, thanks for letting me know. Luckily that's not really a problem in this case.

PR #25, opened 2022-02-08
readme fix

grischaerbe:


grischaerbe:
I noticed that the flat is actually called `ignorePointer` when checking for #24

Issue #34, opened 2022-02-18
More control over the OrthographicCamera frustrum


qwertasyx:
As far as i can tell the frustrum is always linked to the size of the parent container/canvas, which probably is good enough in most cases. I would like to have more control over the frustrum like in vanilla three.js (left,right,top,bottom). 

So could imagine having some sort of switch deciding the size based on the parent container (like now) or a custom frustrum.

I´m not sure which way is better as in extending the $size store in some way 
https://github.com/grischaerbe/threlte/blob/087f26706481188bc159abaffa86f0f76beeabef/src/lib/Canvas.svelte#L76-L77
or adding new reactive props (left,right,top,bottom) just for the OrthographicCamera
https://github.com/grischaerbe/threlte/blob/087f26706481188bc159abaffa86f0f76beeabef/src/lib/cameras/OrthographicCamera.svelte#L22-L26

Some guidance which way would fit best would be appreciated :).

PR #37, opened 2022-02-21
Feature/chooseable camera frustrum

qwertasyx:
related to #34

providing frustrum props to set a custom view. Not using them means resizing to parent Container size as default

    <OrthographicCamera
      position={{ z: 50, x: 50, y: 50 }}
      lookAt={{ x: 0, y: 0, z: 0 }}
      left={-50}
      right={50}
      top={50}
      bottom={-50}
    />

### DIFF
--- a/src/lib/cameras/OrthographicCamera.svelte
+++ b/src/lib/cameras/OrthographicCamera.svelte
@@ -43,10 +43,10 @@
   $: if (useCamera) setCamera(camera)
 
   $: {
-    camera.left = $size.width / -2
-    camera.right = $size.width / 2
-    camera.top = $size.height / 2
-    camera.bottom = $size.height / -2
+    camera.left = left || $size.width / -2
+    camera.right = right || $size.width / 2
+    camera.top = top || $size.height / 2
+    camera.bottom = bottom || $size.height / -2
     camera.updateProjectionMatrix()
     invalidate('OrthographicCamera: onResize')
   }
@@ -55,10 +55,6 @@
     if (near !== undefined) camera.near = near
     if (far !== undefined) camera.far = far
     if (zoom !== undefined) camera.zoom = zoom
-    if (left !== undefined) camera.left = left
-    if (right !== undefined) camera.right = right
-    if (bottom !== undefined) camera.bottom = bottom
-    if (top !== undefined) camera.top = top
     camera.updateProjectionMatrix()
     invalidate('OrthographicCamera: props changed')
   }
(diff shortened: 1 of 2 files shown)


grischaerbe:
Thanks for your PR, I like the idea!
I'm not sure about the individual props as that could lead to unintentional effects. If e.g. one would set only `left` (which doesn't make any sense, but still) `right`, `top` and `bottom` would be set according to the canvas size.
Should we make the frustum a single property? This would also align with three.js types as all frustum properties are required when setting up the `THREE.OrthographicCamera`.

```ts
type FrustumProperty = {
  left: number
  right: number
  top: number
  bottom: number
}
```
Component types:
```ts
export type ThrelteFrustum = {
  left: number
  right: number
  top: number
  bottom: number
}

export type OrthographicCameraProperties = Omit<CameraInstanceProperties, 'camera'> & {
  near?: number
  far?: number
  zoom?: number
  frustum?: ThrelteFrustum
}
```

This way you would be sure to not accidentally leave out a property that silently gets calculated from canvas size dimensions. What do you think?

qwertasyx:
Splitting this up into a new type seems weird to me when i compare it with the way a THREE.OrthographicCamera is set up. So like `near`,`far` and `zoom` ... `left`,`right`,`top`,`bottom` are properties of the camera instance and not like `camera.frustrum.left ... `. The Difference here is that `near` f.e. has sensible defaults.

With that said i also dont really have a better idea atm :D .. i have to think about it for a bit. Adding a check if all 4 values are provided or just set defaults based on the provided values could make sense too.

grischaerbe:
You are right, the frustum is defined by 6 values: `near`, `far`, `left`, `right`, `top` and `bottom`. But I assume most of the times you'd want to set `near` and `far` and let threlte deal with the rest when the OrthographicCamera is the camera you are rendering your entire scene with.

The other main use case is using an OrthographicCamera to render to a renderTexture or rendering anything other than the "main view". In this case I would like to be in control of `left`, `right`, `top` and `bottom`. I'd argue it's rare to only set `left` or `top`. For me the confusion stems from the property name `frustum` that is not needed and therefore not "established".

So I see and agree with your point. It's definitely not as much in line with the regular THREE.OrthographicCamera parameters as other threlte components. I'm willing to make senseful deviations though :)

grischaerbe:
I'll close this PR as the new syntax allows for complete control over the camera frustum:

`<T.OrthographicCamera left={10} right={-10} bottom={10} top={-10} />`

PR #55, opened 2022-04-20
Support for interactivity without re-render

daevid66:
Added a pointerInvalidated to context
Moved animationFrameCast outside renderloop
Added onScroll event that invalidates pointer position
Added console logs in render loop (to show that it is working)
Removed auto rotation from getting started scene (to see functionality)

### DIFF
--- a/src/lib/Canvas.svelte
+++ b/src/lib/Canvas.svelte
@@ -13,8 +13,7 @@
     onContextMenu,
     onPointerDown,
     onPointerMove,
-    onPointerUp,
-    onScroll
+    onPointerUp
   } from './lib/interactivity'
   import {
     createRendererAndComposer,
@@ -102,11 +101,11 @@
 <canvas
   use:parentSizeAction
   bind:this={canvas}
-  on:click={(e) => onClick(getCtx(), getRootCtx(), e)}
-  on:contextmenu={(e) => onContextMenu(getCtx(), getRootCtx(), e)}
-  on:pointerup={(e) => onPointerUp(getCtx(), getRootCtx(), e)}
-  on:pointerdown={(e) => onPointerDown(getCtx(), getRootCtx(), e)}
-  on:pointermove={(e) => onPointerMove(getCtx(), getRootCtx(), e)}
+  on:click={(e) => onClick(getCtx(), getRootCtx(), renderCtx, e)}
+  on:contextmenu={(e) => onContextMenu(getCtx(), getRootCtx(), renderCtx, e)}
+  on:pointerup={(e) => onPointerUp(getCtx(), getRootCtx(), renderCtx, e)}
+  on:pointerdown={(e) => onPointerDown(getCtx(), getRootCtx(), renderCtx, e)}
+  on:pointermove={(e) => onPointerMove(getCtx(), getRootCtx(), renderCtx, e)}
   on:pointerenter={() => getCtx().pointerOverCanvas.set(true)}
   on:pointerleave={() => getCtx().pointerOverCanvas.set(false)}
 >
@@ -114,7 +113,6 @@
     <slot />
   {/if}
 </canvas>
-<svelte:window on:scroll={(e) => onScroll(getCtx(), getRootCtx(), e)} />
 
 <style>
   canvas {
--- a/src/lib/Canvas.svelte
+++ b/src/lib/Canvas.svelte
@@ -61,7 +61,7 @@
     frameloop
   )
 
-  const { renderCtx, getCtx, getRootCtx } = contexts
+  const { getCtx, getRootCtx, getRenderCtx, renderCtx } = contexts
 
   // context bindings
   export const { ctx, rootCtx } = contexts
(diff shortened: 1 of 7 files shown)


daevid66:
Not sure on what context you want the pointerInvalided, added it to the main context because it was available to the interaction event handlers.

We need to remove console logs before a possible merge and possibly switch back the getting started scene to auto-rotation.

grischaerbe:
Hey @daevid66, looks promising, thank you. I'll have a look at it tomorrow!

src/lib/lib/frameloop.ts
```diff
@@ -53,17 +51,21 @@ export const useFrameloop = (
   renderCtx: ThrelteRenderContext
 ): void => {
   useRaf(() => {
```
grischaerbe, line 53: I suggest this for better readability:
```ts
const shouldRender =
renderCtx.frameloop === 'always' ||
renderCtx.frameInvalidated ||
renderCtx.frameHandlers.size > 0

const shouldRaycast = shouldRender || ctx.pointerInvalidated
```
and commence accordingly

src/lib/types/types.ts
```diff
@@ -60,6 +60,7 @@ export type ThrelteContext = {
   size: Readable<Size>
   pointer: Writable<Vector2>
   pointerOverCanvas: Writable<boolean>
+  pointerInvalidated: boolean
```
grischaerbe, line 63: I think this property should go either in the `ThrelteRenderContext` or a new, separate `ThrelteInteractivityContext` where we would extract these properties from the `ThrelteRootContext`:
```ts
addRaycastableObject: (obj: Object3D) => void
removeRaycastableObject: (obj: Object3D) => void
addInteractiveObject: (obj: Object3D) => void
removeInteractiveObject: (obj: Object3D) => void
interactiveObjects: Set<Object3D>
raycastableObjects: Set<Object3D>
raycaster: Raycaster
lastIntersection: Intersection<Object3D<Event>> | null
```
but that's probably something for later.

src/lib/lib/interactivity.ts
```diff
@@ -91,6 +92,12 @@ export const onPointerUp = onEvent
 export const onPointerDown = onEvent
 export const onPointerMove = onEvent
 
+export const onScroll = (ctx: ThrelteContext,
+  rootCtx: ThrelteRootContext,
+  e: Event) => {
```
grischaerbe, line 97: Only one context needs to be passed

src/lib/lib/frameloop.ts
```diff
@@ -53,17 +51,21 @@ export const useFrameloop = (
...
       !renderCtx.frameInvalidated &&
-      rootCtx.interactiveObjects.size === 0 &&
       renderCtx.frameHandlers.size === 0
     ) {
       return
     }
     const camera = get(ctx.camera)
     if (!camera || !ctx.composer || !ctx.renderer) return
-    animationFrameRaycast(ctx, rootCtx)
+    console.log("render")
```
grischaerbe, line 68: I always use this tiny handy function to log to an expression that you can realtime track in Chrome DevTools without your console exploding:
```ts
import { browser } from '$app/env'

export const logToExpression = (name: string, value: any) => {
	if (!browser) return
	;(window as any)[name] = value
}
```

Bildschirmfoto 2022-04-21 um 10 18 50

src/lib/Canvas.svelte
```diff
@@ -114,6 +114,7 @@
     <slot />
   {/if}
 </canvas>
+<svelte:window on:scroll={(e) => onScroll(getCtx(), getRootCtx(), e)} />
```
grischaerbe, line 118: clever! Although I was not able to move the cursor onto an interactive object by scrolling and affecting it by that. Only when I moved the mouse a bit.

grischaerbe: I then realized that this behaviour is the same as it was before, so that's fine.

grischaerbe (changes requested):
Thanks a lot for your PR! I think it solves the problem quite elegantly. I was not able to "upset" the system in a way that the behaviour differs from before.
I would consider this a somewhat breaking change since the frameloop is not continuously running and imperatively changing things around might not work anymore, although that being an edge-case.

Before merging can you maybe set up a test scene where everything is being accounted for so that also future changes to the interactivity system can easily be tested? Manual testing of course, don't bother to implement *actual* tests.

src/lib/Canvas.svelte
daevid66: Hmm I guess the scroll event doesn't contain a new pointerposition we cant do a call to setPointerFromEvent(). Since it's not working, I will remove all scroll related stuff.

src/lib/lib/interactivity.ts
daevid66: Removed scroll functionality

src/lib/types/types.ts
daevid66: Moved it to renderCtx for now. Not sure what you mean with extract from root context.

daevid66:
Thanks for your feedback on the PR. I think I have now made all the required changes. I also removed the console logs so It is a bit hard to see that it is working as expected but I tested before removing the logs.

As for test scene, I used the gettingStarted scene, where I removed the auto-rotation. I think this is a good test case but it's hard to test without some kind of debug mode. i see that you have a debugFrame in frameloop, but I am not sure how this is used.

grischaerbe:
Thanks for making the changes. However before merging I'd like to clean it up a bit and implement a special test page as this is kind of a vital part of the whole thing and making sure it's set up correctly for every commit to come is important to me. Also `npm run check` still shows an error. I'm playing around with Playwright tests however I'm not sure if Playwright is capable of handling mousedown/mouseup/mousemove events in that fashion.
In any case, is it okay for you to grant permission to ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) on this pull request so I can rebase and clean it up a bit?

daevid66:
Sure feel free to implement however it fits best into the project (it's your project). It looks to me like the "Allow edits by maintainers" is already checked. Let me now if you cannot commit to it.

I fixed the ts check error, by the way.

grischaerbe:
I added a simple interaction test on `/tests/interaction` to check if events go through. Looks good to me!

grischaerbe:
I guess we can merge this now, thanks @daevid66 for your effort, I really like it!

PR #63, opened 2022-05-16
feat: 🎸 Arrow,Axes,Box,Camera and DirectionalLight Helpers

takoyaro:
Added some of the helpers. Does not include the grid helpers nor the other lights.

I also added the docs for each of these but intentionally didn't include them in this PR because the docs behave in a strange way in my environment.
image

_This is my first "public" PR. Feedback welcome_

### DIFF
--- a/src/lib/helpers/ArrowHelper.svelte
+++ b/src/lib/helpers/ArrowHelper.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import type { ArrowHelperProperties, Object3DProperties } from '$lib/types/components';
-  import { ArrowHelper as ThreeArrowHelper, type ColorRepresentation, Vector3 } from 'three'
+  import { ArrowHelper as ThreeArrowHelper } from 'three'
   import Object3DInstance from '../instances/Object3DInstance.svelte'
 
   // Object3DInstance
--- a/src/lib/helpers/ArrowHelper.svelte
+++ b/src/lib/helpers/ArrowHelper.svelte
@@ -1,0 +1,19 @@
+<script lang="ts">
+  import { ArrowHelper, Color, ColorRepresentation, Vector3 } from 'three'
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+
+  //self
+  export let dir:Vector3;
+  export let origin:Vector3;
+  export let length:number = 1;
+  export let color:ColorRepresentation = "#ffff00";
+  export let headLength:number = 0.2 * length;
+  export let headWidth:number = 0.2 * headLength;
+
+  let helper: ArrowHelper = new ArrowHelper(dir,origin,length,color,headLength,headWidth);
+
+</script>
+
+{#if helper}
+  <Object3DInstance object={helper} />
+{/if}
--- a/src/lib/helpers/ArrowHelper.svelte
+++ b/src/lib/helpers/ArrowHelper.svelte
@@ -1,19 +1,52 @@
 <script lang="ts">
-  import { ArrowHelper, Color, ColorRepresentation, Vector3 } from 'three'
+  import type { ArrowHelperProperties, Object3DProperties } from '$lib/types/components';
+  import { ArrowHelper as ThreeArrowHelper, type ColorRepresentation, Vector3 } from 'three'
   import Object3DInstance from '../instances/Object3DInstance.svelte'
 
+  // Object3DInstance
+  export let position: Object3DProperties['position'] = undefined
+  export let scale: Object3DProperties['scale'] = undefined
+  export let rotation: Object3DProperties['rotation'] = undefined
+  export let lookAt: Object3DProperties['lookAt'] = undefined
+  export let viewportAware: Object3DProperties['viewportAware'] = false
+  export let inViewport: Object3DProperties['inViewport'] = false
+  export let castShadow: Object3DProperties['castShadow'] = undefined
+  export let receiveShadow: Object3DProperties['receiveShadow'] = undefined
+  export let frustumCulled: Object3DProperties['frustumCulled'] = undefined
+  export let renderOrder: Object3DProperties['renderOrder'] = undefined
+  export let visible: Object3DProperties['visible'] = undefined
+  
   //self
-  export let dir:Vector3;
-  export let origin:Vector3;
-  export let length:number = 1;
-  export let color:ColorRepresentation = "#ffff00";
-  export let headLength:number = 0.2 * length;
-  export let headWidth:number = 0.2 * headLength;
+  export let dir:ArrowHelperProperties['dir'] = undefined;
+  export let origin:ArrowHelperProperties['origin'] = undefined;
+  export let length:ArrowHelperProperties['length'] = undefined;
+  export let color:ArrowHelperProperties['color'] = undefined;
+  export let headLength:ArrowHelperProperties['headLength'] = undefined;
+  export let headWidth:ArrowHelperProperties['headWidth'] = undefined;
 
-  let helper: ArrowHelper = new ArrowHelper(dir,origin,length,color,headLength,headWidth);
+  let helper: ThreeArrowHelper = new ThreeArrowHelper (dir,origin,length,color,headLength,headWidth);
+
+  $:{
+    if(length!==undefined || headLength!==undefined || headWidth!==undefined) helper.setLength(length,headLength,headWidth);
+    if(color!==undefined) helper.setColor(color);
+  }
 
 </script>
 
 {#if helper}
-  <Object3DInstance object={helper} />
+  <Object3DInstance 
+  object={helper}
+  {position}
+  {scale}
+  {rotation}
+  {lookAt}
+  {frustumCulled}
+  {renderOrder}
+  {visible}
+  {castShadow}
+  {receiveShadow}
+  {viewportAware}
+  bind:inViewport
+  on:viewportenter
+  on:viewportleave/>
 {/if}
--- a/src/lib/helpers/AxesHelper.svelte
+++ b/src/lib/helpers/AxesHelper.svelte
@@ -1,0 +1,14 @@
+<script lang="ts">
+  import { AxesHelper } from 'three'
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+
+  //self
+  export let length:number = 1;
+
+  let helper: AxesHelper = new AxesHelper(length);
+
+</script>
+
+{#if helper}
+  <Object3DInstance object={helper} />
+{/if}
--- a/src/lib/helpers/BoxHelper.svelte
+++ b/src/lib/helpers/BoxHelper.svelte
@@ -1,0 +1,23 @@
+<script lang="ts">
+    import { type ColorRepresentation, BoxHelper } from 'three'
+    import Object3DInstance from '../instances/Object3DInstance.svelte'
+    import { getParent } from '../internal/HierarchicalObject.svelte'
+  
+    //self
+    export let color:ColorRepresentation = undefined;
+  
+    const parent = getParent()
+  
+    let helper: BoxHelper
+  
+    if (!($parent instanceof Object3DInstance)) {
+      console.warn('<BoxHelper> can only be used as a direct child of <Mesh>') //This description is somewhat innacurate and needs to be improved.
+    } else {
+      helper = new BoxHelper($parent,color)
+    }
+  </script>
+  
+  {#if helper}
+    <Object3DInstance object={helper} />
+  {/if}
+  
--- a/src/lib/helpers/CameraHelper.svelte
+++ b/src/lib/helpers/CameraHelper.svelte
@@ -1,0 +1,20 @@
+<script lang="ts">
+    import { CameraHelper, Camera } from 'three'
+    import Object3DInstance from '../instances/Object3DInstance.svelte'
+    import { getParent } from '../internal/HierarchicalObject.svelte'
+  
+    const parent = getParent()
+  
+    let helper: CameraHelper
+  
+    if (!($parent instanceof Camera)) {
+      console.warn('<CameraHelper> can only be used as a direct child of <PerspectiveCamera>, <OrthographicCamera> ') //This description is somewhat innacurate and needs to be improved once <StereoCamera> is implemented.
+    } else {
+      helper = new CameraHelper($parent)
+    }
+  </script>
+  
+  {#if helper}
+    <Object3DInstance object={helper} />
+  {/if}
+  
--- a/src/lib/helpers/DirectionalLightHelper.svelte
+++ b/src/lib/helpers/DirectionalLightHelper.svelte
@@ -1,0 +1,23 @@
+<script lang="ts">
+  import { type ColorRepresentation, DirectionalLight, DirectionalLightHelper } from 'three';
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+  import { getParent } from '../internal/HierarchicalObject.svelte'
+
+  //self
+  export let size:number = 1;
+  export let color:ColorRepresentation = undefined;
+
+  const parent = getParent()
+
+  let helper: DirectionalLightHelper
+
+  if (!($parent instanceof DirectionalLight)) {
+    console.warn('<PositionalAudioHelper> can only be used as a direct child of <PositionalAudio>')
+  } else {
+    helper = new DirectionalLightHelper($parent,size,color)
+  }
+</script>
+
+{#if helper}
+  <Object3DInstance object={helper} />
+{/if}
--- a/src/lib/types/components.ts
+++ b/src/lib/types/components.ts
@@ -7,7 +7,8 @@ import type {
   Material,
   Mesh,
   Object3D,
-  PositionalAudio
+  PositionalAudio,
+  Vector3
 } from 'three'
 import type { OrbitControls } from 'three/examples/jsm/controls/OrbitControls'
 import type { Pass } from 'three/examples/jsm/postprocessing/Pass'
@@ -293,4 +294,12 @@ export type PositionalAudioProperties = Omit<AudioInstanceProperties<PositionalA
     coneOuterAngle: number
     coneOuterGain: number
   }
+}
+export type ArrowHelperProperties = Omit<Object3DInstanceProperties, 'object'> & {
+  dir?: Vector3,
+  origin?: Vector3,
+  length?:number,
+  color?:ColorRepresentation,
+  headLength?:number,
+  headWidth?:number
 }


src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  import { ArrowHelper, Color, ColorRepresentation, Vector3 } from 'three'
```
grischaerbe, line 2: ArrowHelper must be imported as `import { ArrowHelper as ThreeArrowHelper } from 'three'` and used as `new ThreeArrowHelper` otherwise compiling the types of this component fails as the components name ("ArrowHelper") and the imports name from `three` ("ArrowHelper") are the same. This applies to all other components.

grischaerbe: Imports that are only used as a type must be imported with `import type { ColorRepresentation } from 'three'`. Generally these errors are also visible when you run `npm run check`.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,19 @@
...
+  export let color:ColorRepresentation = "#ffff00";
+  export let headLength:number = 0.2 * length;
+  export let headWidth:number = 0.2 * headLength;
+
+  let helper: ArrowHelper = new ArrowHelper(dir,origin,length,color,headLength,headWidth);
+
+</script>
+
+{#if helper}
+  <Object3DInstance object={helper} />
```
grischaerbe, line 18: The [source](https://github.com/mrdoob/three.js/blob/master/src/helpers/ArrowHelper.js) states that the ArrowHelper extends THREE.Object3D and as such the usage of the component `<Object3DInstance>` is a good choice.
Pass down all props that also an Object3D can have. See the types of other components for reference. Outsource the types to the components.ts and extend on the types `Object3DInstanceProperties`. It's not a perfect system but it's somewhat coherent.
I still hope for the svelte language-tools to improve the situation as passing down props with {...$$props} with type support would make everything so much easier.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,19 @@
...
+
+  //self
+  export let dir:Vector3;
+  export let origin:Vector3;
+  export let length:number = 1;
+  export let color:ColorRepresentation = "#ffff00";
+  export let headLength:number = 0.2 * length;
+  export let headWidth:number = 0.2 * headLength;
+
+  let helper: ArrowHelper = new ArrowHelper(dir,origin,length,color,headLength,headWidth);
```
grischaerbe, line 13: Account for reactivity: What if any of the properties change?

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  import { ArrowHelper, Color, ColorRepresentation, Vector3 } from 'three'
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+
+  //self
+  export let dir:Vector3;
+  export let origin:Vector3;
+  export let length:number = 1;
```
grischaerbe, line 8: The [constructor](https://github.com/mrdoob/three.js/blob/860af3c012915be534f8a5c37d745e2dff339e04/src/helpers/ArrowHelper.js#L18) provides default values, so the component should not do so as the default values "management" is upstream.

src/lib/helpers/BoxHelper.svelte
```diff
@@ -0,0 +1,23 @@
...
+    import { getParent } from '../internal/HierarchicalObject.svelte'
+  
+    //self
+    export let color:ColorRepresentation = undefined;
+  
+    const parent = getParent()
+  
+    let helper: BoxHelper
+  
+    if (!($parent instanceof Object3DInstance)) {
```
grischaerbe, line 13: `Object3DInstance` is the component, I guess you want to check against `$parent instanceof THREE.Object3D && $parent.type === 'Mesh'`

src/lib/helpers/BoxHelper.svelte
```diff
@@ -0,0 +1,23 @@
...
+      console.warn('<BoxHelper> can only be used as a direct child of <Mesh>') //This description is somewhat innacurate and needs to be improved.
+    } else {
+      helper = new BoxHelper($parent,color)
+    }
+  </script>
+  
+  {#if helper}
+    <Object3DInstance object={helper} />
```
grischaerbe, line 21: In this particular case I think it's okay to not pass down all Object3DInstance props since it's purely a debugging thing and actually can cause confusion. What do you think?

src/lib/helpers/BoxHelper.svelte
```diff
@@ -0,0 +1,23 @@
...
+    export let color:ColorRepresentation = undefined;
+  
+    const parent = getParent()
+  
+    let helper: BoxHelper
+  
+    if (!($parent instanceof Object3DInstance)) {
+      console.warn('<BoxHelper> can only be used as a direct child of <Mesh>') //This description is somewhat innacurate and needs to be improved.
+    } else {
+      helper = new BoxHelper($parent,color)
```
grischaerbe, line 16: This helper poses an interesting problem as it's not auto-updating to match the parent it is trying to visualise.
The [docs](https://threejs.org/docs/index.html?q=boxhel#api/en/helpers/BoxHelper) state you need to call `update()` to ~'update the helper's geometry to match the dimensions of the object, including any children.'.
As this is something that is hard to track with threltes unified render loop and i oppose the use of `useFrame` for a debugging utility I suggest to export the function update. I'd refactor it like this:
```ts
const parent = getParent()
// https://github.com/mrdoob/three.js/blob/860af3c012915be534f8a5c37d745e2dff339e04/src/math/Box3.js#L191 I don't think it's throwing for objects that are not a Mesh, so we don't do that too.
export const boxHelper = new BoxHelper($parent, color) // remember to rename this to ThreeBoxHelper as stated in ArrowHelper comment
export const update = () => boxHelper.update()
```
This way users can update the helper easily by themselves until we find a better approach.

src/lib/helpers/BoxHelper.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+    import { type ColorRepresentation, BoxHelper } from 'three'
+    import Object3DInstance from '../instances/Object3DInstance.svelte'
+    import { getParent } from '../internal/HierarchicalObject.svelte'
+  
+    //self
+    export let color:ColorRepresentation = undefined;
```
grischaerbe, line 7: Account for reactivity as well.

src/lib/helpers/CameraHelper.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+    import { CameraHelper, Camera } from 'three'
+    import Object3DInstance from '../instances/Object3DInstance.svelte'
+    import { getParent } from '../internal/HierarchicalObject.svelte'
+  
+    const parent = getParent()
+  
+    let helper: CameraHelper
+  
+    if (!($parent instanceof Camera)) {
```
grischaerbe, line 10: This helper poses the same problem as the BoxHelper, slightly more complex due to the object needing to be of type `Camera`. I propose:

```ts
const parent = getParent()
if (!($parent instanceof Camera)) throw new Error('…') // throwing is fine in this context as it's also a debugging utility.
export const cameraHelper = new CameraHelper($parent) // rename import, see ArrowHelper comment
export const update = () => cameraHelper.update()
```

src/lib/helpers/CameraHelper.svelte
```diff
@@ -0,0 +1,20 @@
...
+  
+    const parent = getParent()
+  
+    let helper: CameraHelper
+  
+    if (!($parent instanceof Camera)) {
+      console.warn('<CameraHelper> can only be used as a direct child of <PerspectiveCamera>, <OrthographicCamera> ') //This description is somewhat innacurate and needs to be improved once <StereoCamera> is implemented.
+    } else {
+      helper = new CameraHelper($parent)
+    }
```
grischaerbe, line 14: The `CameraHelper`s method [`dispose`](https://github.com/mrdoob/three.js/blob/860af3c012915be534f8a5c37d745e2dff339e04/src/helpers/CameraHelper.js#L180) should be called `onDestroy`.

src/lib/helpers/DirectionalLightHelper.svelte
```diff
@@ -0,0 +1,23 @@
...
+
+  //self
+  export let size:number = 1;
+  export let color:ColorRepresentation = undefined;
+
+  const parent = getParent()
+
+  let helper: DirectionalLightHelper
+
+  if (!($parent instanceof DirectionalLight)) {
```
grischaerbe, line 14: Again I'd suggest to refactor like this:

```ts
const parent = getParent()
if (!($parent instanceof DirectionalLight)) throw new Error('…')
export const directionalLightHelper = new DirectionalLightHelper($parent, size, color) // rename import, see ArrowHelper comment
export const update = () => directionalLightHelper.update()
```

src/lib/helpers/DirectionalLightHelper.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+  import { type ColorRepresentation, DirectionalLight, DirectionalLightHelper } from 'three';
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+  import { getParent } from '../internal/HierarchicalObject.svelte'
+
+  //self
+  export let size:number = 1;
+  export let color:ColorRepresentation = undefined;
```
grischaerbe, line 8: Account for reactivity.

src/lib/helpers/DirectionalLightHelper.svelte
```diff
@@ -0,0 +1,23 @@
...
+
+  const parent = getParent()
+
+  let helper: DirectionalLightHelper
+
+  if (!($parent instanceof DirectionalLight)) {
+    console.warn('<PositionalAudioHelper> can only be used as a direct child of <PositionalAudio>')
+  } else {
+    helper = new DirectionalLightHelper($parent,size,color)
+  }
```
grischaerbe, line 18: `dispose` should be called `onDestroy`, see [docs](https://threejs.org/docs/index.html?q=direction#api/en/helpers/DirectionalLightHelper).

src/lib/helpers/AxesHelper.svelte
```diff
@@ -0,0 +1,14 @@
...
+
+  //self
+  export let length:number = 1;
+
+  let helper: AxesHelper = new AxesHelper(length);
+
+</script>
+
+{#if helper}
+  <Object3DInstance object={helper} />
```
grischaerbe, line 13: I'd say since this one is not bound to an object you might want to transform it. Pass down all props of `<Object3DInstance>`.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  import { ArrowHelper, Color, ColorRepresentation, Vector3 } from 'three'
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+
+  //self
+  export let dir:Vector3;
```
grischaerbe, line 6: Please use prettier for formatting and eslint for linting as well as `npm run check` to check for errors or warnings.

grischaerbe:
Thanks a ton for that PR, great stuff. These helpers go a long way in quickly prototyping scenes. I made some comments which would need to be addressed before merging though. Can you incorporate your docs in this PR so I can give it a testdrive to see whats going wrong?
Again, thanks a lot! ❤️

grischaerbe (changes requested):
Thanks a ton for that PR, great stuff. These helpers go a long way in quickly prototyping scenes. I made some comments which would need to be addressed before merging though. Can you incorporate your docs in this PR so I can give it a testdrive to see whats going wrong?
Again, thanks a lot! ❤️

takoyaro:
@grischaerbe this feedback is very appreciated, thank you for your time..!!

Before diving any deeper, I just made the changes you mentioned to the `ArrowHelper` component and I'd like to know how you feel about it before I apply the changes to the rest of the helpers.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import type { ArrowHelperProperties, Object3DProperties } from '$lib/types/components';
+  import { ArrowHelper as ThreeArrowHelper } from 'three'
+  import Object3DInstance from '../instances/Object3DInstance.svelte'
+
+  // Object3DInstance
+  export let position: Object3DProperties['position'] = undefined
```
grischaerbe, line 7: These props can also be typed from `ArrowHelperProperties` as these types are *extending* the `Object3DInstanceProperties` types.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
...
+  export let viewportAware: Object3DProperties['viewportAware'] = false
+  export let inViewport: Object3DProperties['inViewport'] = false
+  export let castShadow: Object3DProperties['castShadow'] = undefined
+  export let receiveShadow: Object3DProperties['receiveShadow'] = undefined
+  export let frustumCulled: Object3DProperties['frustumCulled'] = undefined
+  export let renderOrder: Object3DProperties['renderOrder'] = undefined
+  export let visible: Object3DProperties['visible'] = undefined
+  
+  //self
+  export let dir:ArrowHelperProperties['dir'] = undefined;
```
grischaerbe, line 20: Style: Just let prettier run over everything.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
...
+  export let length:ArrowHelperProperties['length'] = undefined;
+  export let color:ArrowHelperProperties['color'] = undefined;
+  export let headLength:ArrowHelperProperties['headLength'] = undefined;
+  export let headWidth:ArrowHelperProperties['headWidth'] = undefined;
+
+  let helper: ThreeArrowHelper = new ThreeArrowHelper (dir,origin,length,color,headLength,headWidth);
+
+  $:{
+    if(length!==undefined || headLength!==undefined || headWidth!==undefined) helper.setLength(length,headLength,headWidth);
+    if(color!==undefined) helper.setColor(color);
```
grischaerbe, line 31: The property `color` is of type `THREE.ColorRepresentation` but the method `setColor` takes a `THREE.Color` object as an argument. This might be sufficient: 
```ts
if (color !== undefined) helper.setColor(color instanceOf Color ? color : new Color(color))
```

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
...
+  export let origin:ArrowHelperProperties['origin'] = undefined;
+  export let length:ArrowHelperProperties['length'] = undefined;
+  export let color:ArrowHelperProperties['color'] = undefined;
+  export let headLength:ArrowHelperProperties['headLength'] = undefined;
+  export let headWidth:ArrowHelperProperties['headWidth'] = undefined;
+
+  let helper: ThreeArrowHelper = new ThreeArrowHelper (dir,origin,length,color,headLength,headWidth);
+
+  $:{
+    if(length!==undefined || headLength!==undefined || headWidth!==undefined) helper.setLength(length,headLength,headWidth);
```
grischaerbe, line 30: We might want to warn the user if *some* of these properties were supplied but not all.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
...
+  export let dir:ArrowHelperProperties['dir'] = undefined;
+  export let origin:ArrowHelperProperties['origin'] = undefined;
+  export let length:ArrowHelperProperties['length'] = undefined;
+  export let color:ArrowHelperProperties['color'] = undefined;
+  export let headLength:ArrowHelperProperties['headLength'] = undefined;
+  export let headWidth:ArrowHelperProperties['headWidth'] = undefined;
+
+  let helper: ThreeArrowHelper = new ThreeArrowHelper (dir,origin,length,color,headLength,headWidth);
+
+  $:{
```
grischaerbe, line 29: The direction can change too, account for reactivity.
Currently the property `dir` is of type `THREE.Vector3`. Threlte uses an extension to that type called `Position`

```ts
export type Position =
  | Vector3
  | {
      x?: number
      y?: number
      z?: number
    }
```

so that a user can also just pass *some* values and others are assumed 0. I'd say that the property `dir` is a good fit for that type as well, but we need to account for that of course.
On top of that [the docs](https://threejs.org/docs/index.html?q=arrowhelper#api/en/helpers/ArrowHelper.setDirection) state that the vector passed into `setDirection()` needs to be a unit vector, so it needs to be normalized.

grischaerbe: Another option would be to lay out the component types like this:

```ts
export type ArrowHelperProperties = Omit<Object3DInstanceProperties, 'object'> & {
  target?: Position,
  color?: ColorRepresentation,
  headLength?: number,
  headWidth?: number
}
```
And calculate the property `length` based on the supplied `position` against `target`. Three.js is clear about how they calculate the default values for `headLength` and `headWidth` so I guess we can roll with that too. What do you think?
It's a bit further away from the original three.js implementation and that is something I'm very careful about. In this case it might make setting up the component easier depending on the use case.

src/lib/helpers/ArrowHelper.svelte
```diff
@@ -0,0 +1,52 @@
...
+  
+  //self
+  export let dir:ArrowHelperProperties['dir'] = undefined;
+  export let origin:ArrowHelperProperties['origin'] = undefined;
+  export let length:ArrowHelperProperties['length'] = undefined;
+  export let color:ArrowHelperProperties['color'] = undefined;
+  export let headLength:ArrowHelperProperties['headLength'] = undefined;
+  export let headWidth:ArrowHelperProperties['headWidth'] = undefined;
+
+  let helper: ThreeArrowHelper = new ThreeArrowHelper (dir,origin,length,color,headLength,headWidth);
```
grischaerbe, line 27: helper is never reassigned: use `const` and export it so that users can bind onto it: `export const arrowHelper = new ThreeArrowHelper(…)`.

src/lib/helpers/ArrowHelper.svelte
grischaerbe: Also a very finicky detail but while we're at it:
This reactive block is rerunning on any component prop change that is referenced. If the color is changing, we might not want to call `setLength` or `setDir` as that could be an expensive operation, so make that a separate reactive statement for each function call:

```ts
$: setLength…
$: setColor…
$: setDir…
```

grischaerbe (changes requested):
We're getting there, nice work :)
In general try to see what the three.js docs offer in terms of methods and property setters and check that against the component props.

src/lib/helpers/ArrowHelper.svelte
takoyaro: > We might want to warn the user if _some_ of these properties were supplied but not all.

Would that be done through the `invalidate` function? I've noticed other components use it to send debug messages to the user. Is that what you mean?

src/lib/helpers/ArrowHelper.svelte
takoyaro: I believe Helpers should be able to be set up very straightforward so the implementation you're suggesting here is very appealing.
Honestly, I don't have the 3D/Math knowledge to infer the `dir` unit vector from an `origin` and a `target` Vector but I could implement the length.

src/lib/helpers/ArrowHelper.svelte
grischaerbe: Ah true, i totally missed that part! Every time a prop changes and the component takes that into account e.g. by calling methods like `setLength` we need to invalidate the current frame by calling `invalidate(debugMessage?: string)` you should pass a debugmessage as that makes it easier to spot performance bottlenecks when using `<Canvas debugFrameloop>`.

takoyaro:
Just leaving this here, this PR isn't dead. I'm attending a wedding, will be back home on 30th (JST). Will get this done when I'm back home.

grischaerbe:
Hey @takoyaro! A lot of these helpers are now by nature part of [Threlte v5](https://threlte.xyz/core-transition):

```html
<T.ArrowHelper />
```

PR #70, opened 2022-05-30
Fixed wrong event type bug in OrbitControl events

daevid66:
The OrbitControl end-event is currently not working, see proposed change.


grischaerbe:
Ah, good catch! I guess you spotted it because you depend on that event, so I'll merge it right away 👍

PR #84, opened 2022-07-15
InstancedMesh interactivity bug fix

Klowes:
Fixed broken interactivity on MeshInstance at element 0


grischaerbe (approved):
Ah, good catch! Thanks a lot!

grischaerbe:
I'll release directly since it's quite blocking.

PR #86, opened 2022-07-15
Stricter click events

PontusHorn:
Validate `click` events to make sure the last `pointerdown` event hit the same instance as the `click` event did. This heuristic more closely resembles how the DOM works, and prevents accidental clicks while e.g. using `OrbitControls` or otherwise dragging on the canvas.

### DIFF
--- a/apps/docs/src/routes/concepts/interactivity.md
+++ b/apps/docs/src/routes/concepts/interactivity.md
@@ -4,7 +4,7 @@ title: Interactivity
 
 # Interactivity
 
-[Open the interactivity example in a Svelte REPL](https://svelte.dev/repl/c2b74dcf88ec4b4681b664e202c54274?version=3.46.2)
+[Open the interactivity example in a Svelte REPL](https://svelte.dev/repl/612bfb61f8f84e299fe5083e4767b1e1?version=3.46.2)
 
 Listen to events of a `<Mesh>` and a `<MeshInstance>` as if it would be a regular DOM element:
 
--- a/packages/core/src/lib/lib/interactivity.ts
+++ b/packages/core/src/lib/lib/interactivity.ts
@@ -76,10 +76,13 @@ export const useEventRaycast = (
     setPointerFromEvent(ctx, e)
 
     const closestIntersection = getClosestIntersection(rootCtx, pointer, camera)
-    if (eventType === 'pointerdown') {
-      pointerDownOn = closestIntersection
-        ? { object: closestIntersection.object, instanceId: closestIntersection.instanceId }
-        : null
+    if (eventType === 'pointerdown' && closestIntersection) {
+      // Remember which object was pressed in order to validate the next click event
+      const { object, instanceId } = closestIntersection
+      pointerDownOn = { object, instanceId }
+    } else if (['click', 'pointerup', 'pointerdown'].includes(eventType)) {
+      // Clear pointerDownOn when pointer is released or when pointerdown hits nothing
+      pointerDownOn = null
     }
 
     if (!closestIntersection) return
--- a/packages/core/src/lib/lib/interactivity.ts
+++ b/packages/core/src/lib/lib/interactivity.ts
@@ -1,6 +1,13 @@
 import { onDestroy } from 'svelte'
 import { get } from 'svelte/store'
-import type { Camera, Event, Intersection, Object3D, Vector2 } from 'three'
+import {
+  MOUSE,
+  type Camera,
+  type Event,
+  type Intersection,
+  type Object3D,
+  type Vector2
+} from 'three'
 import type {
   ThrelteContext,
   ThreltePointerEventMap,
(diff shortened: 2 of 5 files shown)


apps/docs/src/routes/concepts/interactivity.md
```diff
@@ -4,7 +4,7 @@ title: Interactivity
 
 # Interactivity
 
-[Open the interactivity example in a Svelte REPL](https://svelte.dev/repl/c2b74dcf88ec4b4681b664e202c54274?version=3.46.2)
+[Open the interactivity example in a Svelte REPL](https://svelte.dev/repl/612bfb61f8f84e299fe5083e4767b1e1?version=3.46.2)
```
PontusHorn, line 7: While working on this change, I added an example like the one in this REPL to this page in order to validate my changes. I'm curious if you have thoughts about having more live examples embedded in the docs, or if you prefer just the REPL links? I was also wondering if you have a better approach to testing your changes than that.

packages/core/src/lib/lib/interactivity.ts
```diff
@@ -68,27 +75,23 @@ export const useEventRaycast = (
...
-    if (intersects.length > 0 && rootCtx.interactiveObjects.has(intersects[0].object)) {
-      getThrelteUserData(intersects[0].object).eventDispatcher?.(
-        e.type as keyof ThreltePointerEventMap,
-        {
-          ...intersects[0],
-          event: e
-        }
-      )
+    const closestIntersection = getClosestIntersection(rootCtx, pointer, camera)
+    if (eventType === 'pointerdown' && e.button === MOUSE.LEFT) {
```
grischaerbe, line 85: Did you check if that check also working on mobile? Do you do that check because `pointerdown` is also called for the right click?

packages/core/src/lib/lib/interactivity.ts
```diff
@@ -68,27 +75,23 @@ export const useEventRaycast = (
...
+      rootCtx.lastPressIntersection = closestIntersection
```
grischaerbe, line 86: - I recently converted this system to hook-based, so this doesn't need to be in the context actually. A less poluted context is better I guess ;)
- I would rename it to something like `pointerDownOn` and just save the object on `pointerdown` event and compare that on `click` event
- `pointerDownOn` might need a reset on `pointerup`, `click` and other events, but I'm not sure of that. (Later me says it doesn't, am I missing something?)

packages/core/src/lib/lib/interactivity.ts
```diff
@@ -100,6 +103,32 @@ export const useEventRaycast = (
...
+  if (intersects.length === 0 || !rootCtx.interactiveObjects.has(intersects[0].object)) return null
+  return intersects[0]
+}
+
+/**
+ * Validates a click event to make sure the last pointerdown event
+ * hit the same instance as the click event did. This heuristic
+ * resembles how the DOM works and prevents accidental clicks while e.g.
+ * using OrbitControls.
+ */
```
grischaerbe, line 123: Even with comments and well formatted! Super nice!

apps/docs/src/routes/concepts/interactivity.md
grischaerbe: I think live examples are better, but a REPL is so nice because you can just dive in. It makes sense to me to use REPLs in the concepts documentation and use live examples in the rest? What do you think?

packages/core/src/lib/types/types.ts
```diff
@@ -54,6 +54,7 @@ export type ThrelteRootContext = {
   raycastableObjects: Set<Object3D>
   raycaster: Raycaster
   lastIntersection: Intersection<Object3D<Event>> | null
+  lastPressIntersection: Intersection<Object3D<Event>> | null
```
grischaerbe, line 57: Move to hook

grischaerbe (changes requested):
Thanks so much for your very welcomed PR, I totally missed that in the first interaction iteration 👍

apps/docs/src/routes/concepts/interactivity.md
PontusHorn: I guess the ideal thing would be a combination of both. 🙂 But I don't think the Svelte REPL lends itself that well to embedding. CodeSandbox embeds maybe?

grischaerbe: You can `iframe` it, did it before. Unfortunately the slider to move the view from Input <-> Output then breaks, which makes it pretty much unusable. CodeSandbox is a good idea but also hard to maintain since they belong to someone. Is there some space for a new Tool? 😃

packages/core/src/lib/lib/interactivity.ts
PontusHorn: It works with mobile/tap emulation in Chrome, at least. Yeah, that was the thought behind it - although I suppose it might not be strictly necessary and would only affect if you're doing something weird like holding several mouse buttons simultaneously. I think it's theoretically more correct to have the check, but don't mind removing it if you think it's cleaner without.

packages/core/src/lib/lib/interactivity.ts
PontusHorn: I have to admit that I'm not sure what you mean by hook-based in the Svelte context - is the implication that the function's lifecycle is tied to the lifecycle of the Canvas component?

Moving the state from the context sounds great - wasn't too happy about putting it there in the first place. Just to clarify, you're suggesting putting it in the top of the `useEventRaycast` scope instead, along `camera` and `pointer`?

It doesn't need a reset for the sake of this feature - but with a name like `pointerDownOn`, it probably should be reset as the name would be a bit misleading otherwise. I'll take care of it.

packages/core/src/lib/lib/interactivity.ts
grischaerbe: R3F handles this by checking the object on pointerdown only and I think this is the better because solution because it is more generic and has less unknown side effects: https://github.com/pmndrs/react-three-fiber/blob/master/packages/fiber/src/core/events.ts#L418-L422

packages/core/src/lib/lib/interactivity.ts
grischaerbe: Yep, exactly. just a regular `let lastPointerDownOn: Object3D | undefind = undefined` in `useEventRaycast` is perfectly fine 👍

PontusHorn: Probably best to store the instanceId also, so that we can compare that like you do in `targetChanged`?

grischaerbe: Good one, yes!

packages/core/src/lib/lib/interactivity.ts
PontusHorn: Fixed!

packages/core/src/lib/lib/interactivity.ts
PontusHorn: Fixed 👍

apps/docs/src/routes/concepts/interactivity.md
PontusHorn: Maybe 😄 Anyway, this is probably the wrong place for that discussion - I was just curious about the difference between pages and your distinction sounds reasonable enough to me!

PontusHorn:
Messed it up a bit with my last commit - will fix 😅

PontusHorn:
There we go, tested it a bit and seems to work decently again.

Normally I'd probably add some tests to validate `useEventRaycast`, but I can't find any other tests in the repo - what are your thoughts around that?

grischaerbe (approved):

grischaerbe:
Great stuff, thanks a lot ❤️

PR #87, opened 2022-07-16
Contact shadows

grischaerbe:
This is a port of https://github.com/pmndrs/drei#contactshadows

Unfortunately it has the same limitations in terms of moving it around. I'll see if I can fix that in a future release.


grischaerbe:
Hm, somehow the compiled output looks different, I have to investigate.

PR #95, opened 2022-07-24
Add links to related external docs

PontusHorn:
Adds a prop to the ModuleSummary component that allows linking out to related documentation, for example to the three.js concepts that most threlte components abstract over. Also adds some more or less obvious candidates for links to existing documentation.

### DIFF
--- a/apps/docs/src/routes/core/ambient-light.md
+++ b/apps/docs/src/routes/core/ambient-light.md
@@ -2,7 +2,7 @@
 title: AmbientLight
 ---
 
-!!!module_summary title=AmbientLight|name=AmbientLight|sourcePath=lights/AmbientLight.svelte|from=core|type=component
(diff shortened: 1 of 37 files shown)


apps/docs/src/routes/core/ambient-light.md
```diff
@@ -2,7 +2,7 @@
 title: AmbientLight
 ---
 
-!!!module_summary title=AmbientLight|name=AmbientLight|sourcePath=lights/AmbientLight.svelte|from=core|type=component
+!!!module_summary title=AmbientLight|name=AmbientLight|sourcePath=lights/AmbientLight.svelte|from=core|type=component|relatedDocs={[{name:"three.js AmbientLight reference",url:"https://threejs.org/docs/#api/en/lights/AmbientLight"}]}
```
PontusHorn, line 5: Not sure how I feel about the syntax - it's a bit clunky, but I feel like it's mostly due to it all being compressed on one line for this special shorthand component syntax. Maybe it'd be easier to read if we just use normal component syntax for the module summary now that there's so many props?

grischaerbe: I think both ways work fine-ish. There's unfortunately no type checking in .md files so, the markdown extension way is fine for me.

grischaerbe:
Thank you so much for this ❤️ Great change and a really good addition for people that are just starting off.

PR #101, opened 2022-07-25
Updated README

amr3k:
Added `@threlte/rapier` to packages


grischaerbe:
Ah, great catch! Thanks so much ❤️

PR #117, opened 2022-07-31
Updated docs

amr3k:
The main page of [threlte docs](https://threlte.xyz) has an invalid path for the code example as seen in this screenshot:
image


grischaerbe:
Good catch! Thanks for the effort! ❤️

PR #124, opened 2022-08-02
Added contact force events and streamlined event management

grischaerbe:


grischaerbe:
In the process also cleaned up the Event types on `<Collider>`, `<AutoColliders>` and `<RigidBody>`

PR #131, opened 2022-08-05
Examples Page + Playground

grischaerbe:
This PR is a combined effort to utilise a new Playground available at 
/playground
to also show and be able to edit examples.

We could stick to the following structure:
- all examples, whether it be the examples that are visible in the docs or examples that will be linked to from the`/examples` page reside in `apps/docs/src/examples` folder.
- ~~No subdirectories in examples are allowed (makes querying easier, we'd need to clean that up a little)~~ Actually subdirectories are allowed and were fairly easy to implement 👍 
- ~~Only `.svelte` files are allowed. When you would normally make a .ts/.js file, write that content in `<script context="module">`. By that, a playground from that example can be spun up by linking to /playground/example-dir-name~~ Also not true anymore, .ts files are also allowed and transpiled before spun up as a playground.
- The playground logic itself is in a layout file ~~`apps/docs/src/routes/__layout-playground.svelte`~~ `apps/docs/src/routes/playground/__layout@reset.svelte`
- Every example not only can be spun up into a playground but can also be annotated with a page:
  - Let's say there's an example in `/apps/docs/src/examples/cool-example`, the playground URL would be `/playground/cool-example`. Now if there's a page file located at `apps/docs/src/routes/playground/[slug]/cool-example` it's prepended/appended (TBD) to the playground page. This can be a good place to explain some core mechanics of the example.
  
At some point we can also think about screenshots for examples. Using Playwright we can easily automate that or put that in a package script.


grischaerbe:
The Playground is finished for now. It now uses the native Svelte preprocessor for typescript.
I added a menu icon in the navbar

DefinitelyMaybe:
Wow! dude that's super cool.

I've gotta get a move on to catch up with some examples.

Quick question, what's the zip file made up of? when saving a playground. I saved a simple scene and tried to unzip it but windows said there wasn't anything in the folder.

grischaerbe:
Its a simple JSON containing the components and some meta info, right now only the title of the playground. It's not in a folder though, I wasn't able to make folders in a `CompressionStream` which would've been helpful as then we could opt for individual `.svelte` files instead of a single JSON file for the components. This would make the local use in a project super simple. Do you know if that is possible?

grischaerbe:
Oh it seems [it is possible](https://github.com/Stuk/jszip)! I'm going to implement that then, what do you think?

grischaerbe:
The playground now exports a `.zip` file containing all components as individual `.svelte` files as well as a `meta.json` that can include meta information about a playground.
The playground also now imports a zip file containing individual `.svelte` files as well as a `meta.json` file.

DefinitelyMaybe:
I'm liking where it's at. feeling about ready?

DefinitelyMaybe:
failed because of file naming? @grischaerbe

grischaerbe:
@DefinitelyMaybe Is it building on your local machine? The deployment error hints towards file case issue or a route that is linked but not present.

DefinitelyMaybe:
Haiyaa... where is my brain at. Why no just local build check 🤔🤔
One moment.

DefinitelyMaybe:
Really not sure about this one actually. I've commented out all of my new examples except the terrain one which did have a successful build above ...but my local build is still failing with just that. Same Error about `404 /examples/....`

I even tried a build with terrain but commenting out all of this:

```md
<script lang="ts">
import Wrapper from '$examples/terrain/Wrapper.svelte'
</script>

# Terrain

<ExampleWrapper>
	<Wrapper />
</ExampleWrapper>

&&&code_wrapper
@[code svelte|title=Wrapper.svelte](../../examples/terrain/Wrapper.svelte)
@[code svelte|title=Scene.svelte](../../examples/terrain/Scene.svelte)
&&&
```

and still got the same error

apps/docs/src/routes/examples/camera-followobject.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Follow Object
+---
+
+<script lang="ts">
+import Wrapper from '$examples/camera/followobject/wrapper.svelte'
```
grischaerbe, line 6: I think this is the problem. Unfortunately I'm not on my dev machine and only on an iPad using Codespaces but it seems to be the culprit. TypeScript in .md files would be awesome.

grischaerbe: @DefinitelyMaybe

grischaerbe: Seems to work now :)

DefinitelyMaybe: My bad, and thank you. totally agree that typescript or a lsp picking that up would save both of us the trouble.

DefinitelyMaybe:
well that's weird. Vercel build worked but my local build did not.
I'm frustrated, thinking it's best I move onto something else

grischaerbe:
@DefinitelyMaybe what's the problem with your local build? GitHub Codespaces works fine with threlte if that's any help 👍

DefinitelyMaybe:
Wooh! how you doing @grischaerbe , are we ready for a merge?

grischaerbe:
Hey @DefinitelyMaybe, it's been a while :)
Unfortunately not quite yet, there are some broken things in the Playground (tailwind css is ignored) and I'm not sure about the new "Open In Playground" button:

Bildschirmfoto 2022-10-17 um 00 27 49

Do you think it still needs a "Show Code" Button?

DefinitelyMaybe:
Yeah, nah, I think the show code button is still a great idea, my preference would be to keep it.

It's feels like the intellisense snippets that come up in your IDE which present the simple example right there in your editor. 

Playground button bottom left, show code button bottom right?

tailwindcss I can live without for examples/playground, converting it to plain css will just be an extra step before finishing a new example.

grischaerbe:
Bildschirmfoto 2022-10-17 um 10 50 12

The Code is back in. I think you're right.

Considering the css part: We can inject some default CSS into the iframe of the REPL, I guess this would suffice for most examples. The styles for `button`, `input[type="range"]` and `input[type="checkbox"]` would be enough I guess?

DefinitelyMaybe:
Ah, cool. Both together, nice.

Yeah, there's options for css. I'd say getting svelte-knobby going would be a good next step for adding buttons and inputs to the examples.

grischaerbe:
True, a standard way would be the best bet and it also ships with UI styles.

grischaerbe:
So I tried adding `svelte-knobby` to a page with examples and unfortunately it currently has three relevant limitations:

- There's only one panel. So we cannot have multiple panels for multiple examples.
- The panel is always `fixed` and therefore also visible when no example is in the viewport.
- It's not ready for SvelteKit and can only be used inside a client-only component.

Are there any alternatives? It has great API though!

grischaerbe:
It's actually more concise than I thought it would look like. If you like Tweakpane, let's go for it, no?

grischaerbe:
I tweaked the hook a little more towards our use-case and implemented programmatic css directly in the hook so that the looks of the playground are aligned with the looks of the example in the regular docs.

Super basic theming is also implemented.

**Right now only one example/route is converted to using this hook: https://threlte-1xzacbpdx-threlte.vercel.app/rapier/collider**
If you like it, we can go ahead and convert all examples to the new layout, what do you think?

jerzakm:
I think this looks good, I could work on converting more examples later down the road.

DefinitelyMaybe:
Sweet! Can do

grischaerbe:
I think we're ready to ship, what do you all say? 🎉

jerzakm:
Let's do it! I'm all for incremental improvement over perfection and I think the playground is looks good now. We could look for a tweak pane alternative later down the road. 
I'd like to work on some examples/components and not having to re-write their docs later would be awesome.

PR #136, opened 2022-08-11
Updated installation instructions to be compatible with v4.x.x

amr3k:
Sorry, my contributions are only doc updates so far, real code is coming soon 😄


grischaerbe:
Thanks a lot @amr3k!

PR #137, opened 2022-08-11
Added default draco decoder path

### LINKED ISSUE #130: Default `dracoDecoderPath` ?
Hi, I believe it's a common use case is to set the draco decoder path to `https://www.gstatic.com/draco/v1/decoders/`, Should threlte add it as default value?
specifically in this line:

https://github.com/threlte/threlte/blob/7e2e46db8c36fd67265d360f570191ccc5a7c654/packages/extras/src/lib/components/GLTF/GLTF.svelte#L33

This is a slightly improvement to help devs not to include that URL in every GLTF component in every project.

amr3k:
This fixes #130


packages/extras/src/lib/components/GLTF/GLTF.svelte
```diff
@@ -30,7 +30,7 @@
   export let lookAt: GLTFProperties['lookAt'] = undefined
 
   export let url: GLTFProperties['url']
-  export let dracoDecoderPath: GLTFProperties['dracoDecoderPath'] = undefined
+  export let dracoDecoderPath: GLTFProperties['dracoDecoderPath'] = 'https://www.gstatic.com/draco/v1/decoders/';
```
grischaerbe, line 33: If you provide a default value for that prop, the loader which is instantiated in line 62 is always instantiated if the consumer does not provide `undefined` as a prop: `<GLTF dracoDecoderPath={undefined} />` which will lead to loading and instantiating of a DRACO decoder even if the user does not state that the model needs that.

grischaerbe (commented):
Hey @amr3k and thanks so much for your PR ❤️
I see a problem in just defining a default DRACO decoder path and I would suggest to align the API of `<GLTF>` to the one used in `useGltf`: a prop that is called `useDraco` which is of type `boolean | string` and by default `false`. If `useDraco` is true, the default Google-provided decoder is used. If a string is provided, it's used to instantiate a DRACO loader.

This is a breaking change and as such you need to include a changeset. Run `npx changeset` from the repository root to generate it via CLI.

amr3k:
I don't know what vscode did to change all of those lines!!
Should I revert that commit?

packages/extras/src/lib/types/components.ts
```diff
@@ -94,6 +94,6 @@ export type TextProperties = Omit<MeshInstanceProperties, 'mesh'> & {
 
 export type GLTFProperties = Omit<Object3DInstanceProperties, 'object'> & {
   url: string
-  dracoDecoderPath?: string
+  dracoDecoderPath?: string | boolean
```
grischaerbe, line 97: I think we should align the API of `useGltf` and `<GLTF>` so that this property should be called `useDraco`. The "use" also implies that a boolean value can be assigned.

grischaerbe (changes requested):
Change the property name, add a changeset (major bump for extras) and we're good to go 🚀 ❤️

packages/extras/src/lib/types/components.ts
amr3k: Should I just add a new `useDraco` and add a deprecation note for the older `dracoDecoderPath` so that it warns devs to use the newer property?
With that, the older `dracoDecoderPath` would be removed by v5.0.0 
What do you think?

amr3k: Maybe something like 
```js
if (dracoDecoderPath) { // it would be reverted back to the current type (string|undefined)
  console.warn('dracoDecoderPath will be deprecated soon, please use useDraco instead');
}
```

packages/extras/src/lib/types/components.ts
```diff
@@ -1,99 +1,101 @@
 import type {
```
amr3k, line 1: I really don't know why VScode does this!!!

amr3k: Is this encoding issue?

packages/extras/src/lib/types/components.ts
amr3k: I just pushed my proposed changes, but if you don't like it @grischaerbe , it's Ok, I'll remove them and do exactly what you said.

apps/docs/src/examples/extras/use-cursor/Scene.svelte
```diff
@@ -23,8 +22,6 @@
 		material.color = new Color(0xc62004)
 	}
 
-	writable
```
amr3k, line 26: > Absolutely nothing 😁 it probably slipped in. Good catch!

Now it's removed

grischaerbe: Ah perfect, thanks!

apps/docs/src/examples/use-gltf-animations/Scene.svelte
```diff
@@ -15,6 +15,7 @@
 
 	const { scene } = useThrelte()
 	scene.background = new Color(0xeae8e2)
+	// TODO: Use `useDraco` instead of `dracoDecoderPath` after #137 is merged
```
grischaerbe, line 18: You can do that immediately, as everything will be deployed at the same time as soon as #137 is merged.

apps/docs/src/routes/extras/gltf.md
```diff
@@ -73,7 +73,8 @@ frustumCulled: boolean | undefined = undefined
 renderOrder: number | undefined = undefined
 visible: boolean | undefined = undefined
 dispose: boolean | undefined = undefined
-dracoDecoderPath: string | undefined = undefined
+d̶r̶a̶c̶o̶D̶e̶c̶o̶d̶e̶r̶P̶a̶t̶h̶: string | undefined = undefined // Deprecated - use `useDraco` instead
+useDraco: string | boolean = false
```
grischaerbe, line 77: Well done to think about deprecation messages! Do you think it's necessary to add a little hint in terms of a 
```md
:::admonition type="tip"
… 
:::
```
that you can provide a boolean and a string value to useDraco?

packages/extras/src/lib/types/components.ts
grischaerbe: Yeah, I think it was indented with spaces before, so this is fine!

grischaerbe (changes requested):
We're getting really close :) Can you also add a changeset? Run `npx changeset` from the repo root.

apps/docs/src/routes/extras/gltf.md
amr3k: Great, I'll create that tip rn, thanks for the hint :)

amr3k:
> We're getting really close :) Can you also add a changeset? Run `npx changeset` from the repo root.

I chose minor for `@threlte/extras`, I believe it's the appropriate choice as these are not breaking changes :/

grischaerbe (approved):
Well done ❤️

amr3k:
Thanks @grischaerbe , appreciate your patience here.

PR #141, opened 2022-08-14
Removed duplicate slash `/` in source path

amr3k:
I noticed there's an extra `/` in some components/hooks source path URL, which is not good for the environment, so I removed them.

Screenshot_20220812_012352
 
You can check it out, by hovering over the `View Source Code` link in [this page](https://threlte.xyz/extras/edges#example) for example


grischaerbe:
Perfect 💯

PR #144, opened 2022-09-13
feature(rapier): add Attractor component

AlexWarnes:
This first iteration seems to mostly work but a few parts that could use a more thorough review:

1. Dynamically updating the range does not work and I'm not sure why. 
To reproduce: 
*  set world gravity to 0 in `Wrapper.svelte`
*  create a tweened store in `Scene.svelte` and onMount set it to 100 with duration 30000
* You can see with the helper (and via logs) that in the `Attractor.svelte` component the `range` is changing, but the `if (distance < range)` check is not

2. I would like Attractor to inherit parent position if not otherwise specified but I don't know how to do that. This would allow it to work inside a Mesh, Group, etc more easily.

3. I would like Attractor to work inside <RigidBody /> so that the attractors themselves could react to one another (this may be solved by addressing #2 above)

4. I added the `showHelper` prop because I think it's useful, but if #2 above is resolved then users could easily recreate this themselves. It's more of a convenience prop but happy to remove it.

Open to all feedback!

### DIFF
--- a/apps/docs/src/routes/rapier/attractor.md
+++ b/apps/docs/src/routes/rapier/attractor.md
@@ -1,0 +1,32 @@
+---
+title: Attractor
+---
+
+<script lang="ts">
+import Wrapper from '$examples/rapier/attractor/Wrapper.svelte'
+</script>
+
+!!!module_summary title=Attractor|sourcePath=components/Attractor/Attractor.svelte|name=Attractor|from=rapier|type=component
+
+An Attractor simulates a source of gravity. Any RigidBody within range will be "pulled" toward the Attractor at the specified strength.
+
+<ExampleWrapper>
+  <Wrapper />
+</ExampleWrapper>
+
+&&&code_wrapper
+@[code svelte|title=Wrapper.svelte](../../examples/rapier/attractor/Wrapper.svelte)
+@[code svelte|title=Scene.svelte](../../examples/rapier/attractor/Scene.svelte)
+@[code svelte|title=RandomMeshes.svelte](../../examples/rapier/attractor/RandomMeshes.svelte)
+&&&
+
+!!!
+
+### Properties
+
+```ts
+// optional
+position: Position | undefined = undefined
+strength: number | = 1
+range: number = 10
+```
--- a/packages/rapier/src/lib/components/Attractor/Attractor.svelte
+++ b/packages/rapier/src/lib/components/Attractor/Attractor.svelte
@@ -1,0 +1,48 @@
+<script lang="ts">
+  import type { RigidBody } from '@dimforge/rapier3d-compat'
+  import { Mesh, useFrame } from '@threlte/core'
+  import { SphereBufferGeometry } from 'three'
+  import { MeshBasicMaterial } from 'three'
+  import { Vector3 } from 'three'
+  import { useRapier } from '../../hooks/useRapier'
+  import type { AttractorProperties } from '../../types/components'
+
+  const { world } = useRapier()
+
+  export let position: AttractorProperties['position'] = { x: 0, y: 0, z: 0 }
+  export let strength: NonNullable<AttractorProperties['strength']> = 1
+  export let range: NonNullable<AttractorProperties['range']> = 10
+  export let showHelper: AttractorProperties['showHelper'] = false
+
+  const gravitySource = new Vector3(position?.x, position?.y, position?.z)
+
+  function applyImpulseToBodiesInRange() {
+    const impulseVector = new Vector3()
+    world.forEachActiveRigidBody((body: RigidBody) => {
+      const { x, y, z } = body.translation()
+      const bodyV3: Vector3 = new Vector3(x, y, z)
+      const distance: number = gravitySource.distanceTo(bodyV3)
+      if (distance < range) {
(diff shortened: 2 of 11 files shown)


AlexWarnes:
Updated with two additional props `gravityType` and `gravitationalConstant`

Using realistic gravity was annoying, and I don't think what most people would want so I didn't want that to be the only option. I added descriptions of each type and a controlled example ("Advanced") to the md file! Let me know if it's confusing or needs more work.

packages/rapier/src/lib/components/Attractor/Attractor.svelte
```diff
@@ -0,0 +1,69 @@
...
+      }
+    })
+  }
+
+  useFrame(() => {
+    applyImpulseToBodiesInRange()
+  })
+</script>
+
+<Object3DInstance bind:object={obj} {position} />
```
grischaerbe, line 56: You can use this Instance to let child components bind to their parent, which would make sense when you want to bind e.g. `<TransformControls>` to it like this (try it, it's fun 😄):

```html
<Attractor position={{ x: -25, y: 10 }} range={20} strength={strengthLeft} {showHelper}>
	<TransformControls />
</Attractor>
```

So I'm proposing this component layout instead (so that the child meshes inherit the position attribute rather than get their own "copy"):

```html
<Object3DInstance
  bind:object={obj}
  {position}
>
  <slot />
  {#if showHelper}
    <Mesh
      geometry={new SphereBufferGeometry(range)}
      material={new MeshBasicMaterial({ wireframe: true })}
    />
    <Mesh
      geometry={new SphereBufferGeometry()}
      material={new MeshBasicMaterial({ color: 'black', wireframe: false })}
    />
  {/if}
</Object3DInstance>
```

grischaerbe: When it comes to the property `showHelper` we should consider adding a context to the component `<World>` called `threlte-rapier-world-debugging` which is a `Writable<boolean>`. The component `<Debug>` would - besides drawing the helper stuff that is included in rapier - also set this store to `true` so that you would not need to add a separate property to the `<Attractor>` component, what do you think?

apps/docs/src/examples/rapier/attractor/Wrapper1.svelte
```diff
@@ -0,0 +1,101 @@
+<script lang="ts">
```
grischaerbe, line 1: If there's no reason to do otherwise please rename to Wrapper.svelte :)

apps/docs/src/routes/rapier/attractor.md
```diff
@@ -0,0 +1,59 @@
...
+The force applied to rigid-bodies within range is calculated differently, depending on the `gravityType`.
+
+## Basic Example
+
+<ExampleWrapper>
+  <Wrapper1 />
+</ExampleWrapper>
+
+&&&code_wrapper
+@[code svelte|title=Wrapper1.svelte](../../examples/rapier/attractor/Wrapper1.svelte)
```
grischaerbe, line 22: See component: Rename to `Wrapper.svelte`

grischaerbe (commented):
Absolutely love it! Well done 👍 
I have some smaller change requests, tell me what you think! ❤️

grischaerbe (changes requested):
Some smaller changes and then we ship it :) Thanks a ton!

grischaerbe:
Ah and please add a changeset: run `npx changeset` from the repo root 👍

packages/rapier/src/lib/types/components.ts
```diff
@@ -17,11 +17,13 @@ import type {
 import type { Position, Rotation, TransformableObjectProperties } from '@threlte/core'
 import type { RigidBodyTypeString } from '../lib/parseRigidBodyType'
 import type { AutoCollidersShapes, ColliderShapes, CollisionGroupsBitMask } from './types'
-
+import type { Writable } from 'svelte/store'
```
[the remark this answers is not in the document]
AlexWarnes: ✅

packages/rapier/src/lib/index.ts
```diff
@@ -39,5 +41,5 @@ export type {
   CollisionExitEvent,
   SensorEnterEvent,
   SensorExitEvent,
-  ContactEvent
+  ContactEvent,
```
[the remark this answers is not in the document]
AlexWarnes: ✅

packages/rapier/src/lib/components/World/World.svelte
```diff
@@ -1,7 +1,7 @@
 <script context="module" lang="ts">
   import RAPIER from '@dimforge/rapier3d-compat'
-  import { onMount } from 'svelte'
-  import { writable } from 'svelte/store'
+  import { onMount, } from 'svelte'
+  import { writable, type Writable } from 'svelte/store'
```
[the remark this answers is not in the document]
AlexWarnes: ✅

AlexWarnes:
- Cleaned up the imports and exports
- Took a shot at the changeset. Did a minor bump for `/docs` and `/rapier` and left a short summary but wasn't totally sure how to do all that! Let me know if it needs edits and I'll modify it

PR #145, opened 2022-09-30
Docs: Fixed broken link to comment regarding bundle size

lega0208:


grischaerbe:
Oh, thanks so much, good catch! ❤️

lega0208:
No problem 😊 Thanks for creating this project! Haven't had the chance to play around with it that much but I really like it so far. Will make more PRs if I come across anything else. I'm not the most well-versed with Svelte or threejs, but I might try contributing something with more substance eventually! Cheers.

PR #149, opened 2022-10-17
Feature (Environment)

jerzakm:
Environment component to conveniently add envmap to the scene.

Usage looks like this:
```svelte
<Environment
  path = '/envmap/bridge_cube/'
  files= {['posx.jpg', 'negx.jpg', 'posy.jpg', 'negy.jpg', 'posz.jpg', 'negz.jpg']}
  isBackground={true}
  format="ldr"
  groundProjection={{ radius: 200, height: 5, scalar: 100 }}
/>
```
### Component api
Generally pretty straightforward with a few caveats:
- `files` - string or string array with filenames. Based on this the component decides if it is cube or equirect map. Originally i had an `isCubeMap` bool property but it felt like unnecessary extra work for the user.
- `format` - this one I'm not sure about. It's a required prop and can be `ldr` - for .png, .jpg etc and `hdr` for .hdr . In my local playground I used to have support for .exr but there were a few issues and it's something I can implement after we figure out the basics here.

### Possible improvements
- `$:` I use two main reactive statements. One to apply prop changes to `Environment` (like isBackground, path, files ,etc) and another just to handle the Ground Projected Env. I split it to avoid very long code blocks and to allow for possibly removing the ground projected env in case we dont want to bump Three version yet.
- GroundProjectedEnv could possibly be `<Object3DInstance/>`  ?
- General logic could be cleaner (pickLoader, loadEnvironment, setEnvironment functions)
- exr, rgbm support (any others?)

### Other:
- I had to bump three version to at least v143 (went with v145) to support the GroundProjectedEnv.

chrome_SAI1kVXLWi


grischaerbe:
Thank you so much for your PR, from the looks I'd say it's an absolutely rock-solid implementation! ❤️ Unfortunately the build for "apps/docs" on Vercel fails, does it build on your local machine? The error message is a bit unclear, it reads like a broken link in the docs?
Later tonight I'll give it a run too, looking forward!

grischaerbe:
Ah, we had that before. It's a case-sensitivity issue. Did you rename some file or directory but only changed the case? In this case (ha! pun intended) rename the thing completely, make a commit and then name it whatever you want, again commit. This should solve the issue.

grischaerbe:
I'm 95% sure it's a case issue. Another option, stash all your changes (a stash does not contain git history, right?), make a new branch from main, apply stash and force push that new branch into this PR branch?

Please check if your filesystem is case-insensitive (like mine is): make a folder "abc" and after that make a folder "ABC". If your filesystem is case-insensitive, you will get a warning that the folder already exists. The problem is that on these filesystems git will not recognize case-only changes [if not pointed to it](https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git) and references get lost along the way but are still present on your local machine.

grischaerbe:
I built a fresh clone of your PR and it unfortunately also fails.

jerzakm:
It seems that maybe the issue is me bumping three to v145? I made a mess here, but I have a working preview branch:
https://github.com/jerzakm/threlte-environment/tree/environment-build-fix

Since no v145, It's missing GroundProjectedEnv.

I'll sleep on it and figure it all out tomorrow. I thought making the component will be the hard part but I got stuck on build/git magic, gee

grischaerbe:
Yes! Do you know what the problem was?

jerzakm:
I carelessly bumped Three to v145 along with some other packages. I think the issue was sveltekit and/or adapters getting updated. 
I'm going to look up some tutorials and best practices for updating dependencies in monorepo with turbo/pnpm now haha.

grischaerbe:
Alright! I think we are nearly there, do you want me to review it?

jerzakm:
Yes, please. It will be refreshing to actually work on my component for once!

packages/extras/src/lib/components/Environment/Environment.svelte
```diff
@@ -0,0 +1,133 @@
...
+	export let groundProjection: EnvironmentProperties['groundProjection'] = undefined
+	export let format: EnvironmentProperties['format'] = 'ldr'
+	export let encoding: EnvironmentProperties['encoding'] = undefined
+
+	const { scene, invalidate, renderer } = useThrelte()
+
+	$: isCubeMap = Array.isArray(files)
+	$: envPath = `${path || ''}${files}`
+
+	let previousEnvPath: string = ``
```
grischaerbe, line 31: Generally when you do these kind of checks, you would immediately assign the current value to a "previous" variable:

```ts
export let someProp: string
let previousProp = someProp
$: if (someProp !== previousProp) {
  // do something
  previousProp = someProp
}
```

This also ensures that the "do something"-part does not run until it actually changes.

packages/extras/src/lib/components/Environment/Environment.svelte
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	onDestroy(() => {
+		scene.environment = null
+		scene.background = null
+		if (currentProcessedEnv) currentProcessedEnv.dispose()
+		groundProjection = undefined
+		invalidate('Environment destroyed')
+	})
+</script>
```
grischaerbe, line 133: Do you think it's a good idea to outsource the GroundProjectedEnv into its own component to separate the logic and make this component more readable?

```html
{#if groundProjection}
	<GroundProjectedEnv {groundProjection} />
{/if}
```

packages/extras/src/lib/components/Environment/Environment.svelte
```diff
@@ -0,0 +1,133 @@
...
+			invalidate('Removing Environment as scene.background')
+		}
+		if (isBackground && !scene.background && currentProcessedEnv) {
+			scene.background = currentProcessedEnv
+			invalidate('Adding Environment as scene.background')
+		}
+	}
+
+	onDestroy(() => {
+		scene.environment = null
```
grischaerbe, line 127: Maybe we should save the previous (i.e. before this component was mounted) `scene.environment` and `scene.background` and `onDestroy` set it back to these values.

grischaerbe (changes requested):
Generally a really good start! ❤️  There are some things I added comments to. If you're done, add a changeset by running `npx changeset` from the repo root 👍

packages/extras/src/lib/components/Environment/Environment.svelte
jerzakm: I actually thought about it yesterday. GroundProjectedEnv as a child of <Environment/>.  It will make the component more readable but but since it's not something that will get used very often, is it worth adding yet another component to already growing docs?

grischaerbe: We don't have to expose it and frankly I wouldn't since it would then be a part of `@threlte/core`. I would just make it a separate component to clean it up and separate domains.

packages/extras/src/lib/components/Environment/Environment.svelte
```diff
@@ -0,0 +1,133 @@
...
+	import { HDRCubeTextureLoader } from 'three/examples/jsm/loaders/HDRCubeTextureLoader'
+	import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader'
+
+	import { GroundProjectedEnv } from 'three/examples/jsm/objects/GroundProjectedEnv'
+
+	export let path: EnvironmentProperties['path'] = undefined
+	export let files: EnvironmentProperties['files']
+	export let isBackground: EnvironmentProperties['isBackground'] = undefined
+	export let groundProjection: EnvironmentProperties['groundProjection'] = undefined
+	export let format: EnvironmentProperties['format'] = 'ldr'
```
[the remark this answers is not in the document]
jerzakm: Yes, I think so. Should I maybe make it an optional prop and infer by default?
There is a chance that user loads an image with non-standard extension in the filename, but is still for example a .jpg or .hdr .

Additionally .png might require using an RGBM loader (not implemented here yet), and infering format by extension wouldnt work.

RGBM example:
https://threejs.org/examples/?q=env#webgl_materials_envmaps_hdr
https://github.com/mrdoob/three.js/tree/master/examples/textures/cube/pisaRGBM16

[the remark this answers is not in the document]
grischaerbe: Check out [drei's implementation of the component `<Environment>`](https://github.com/pmndrs/drei/blob/master/src/core/Environment.tsx) for a hint on what loader to use when. They don't use any TextureLoader, there's only a `RGBELoader` and a `CubeLoader`.

packages/extras/src/lib/components/Environment/Environment.svelte
grischaerbe: Also check out [drei's implementation](https://github.com/pmndrs/drei/blob/master/src/core/Environment.tsx) for an inspiration on separation of concerns.

jerzakm:
Should I rebase onto upstream/main and make docs in the new format or should we ship this once we get the Environment component in a good shape?

I don't mind either way but I prefer multiple smaller PRs. In the new docs and examples I would like to replace all manually loaded environments (and there's quite a few of them) with the `<Environment/>`

packages/extras/src/lib/components/Environment/Environment.svelte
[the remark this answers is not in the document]
grischaerbe: Isn't the RGBELoader used to load `.hdr` files? What's the problem with it?

grischaerbe:
Concerning the rebase: I guess rebasing on main is probably a good idea. You just need to adapt the [new Examples format](https://github.com/threlte/threlte/tree/main/apps/docs/src/examples), it's probably mainly about how to include the Example in the associated documentation `.md` file.

grischaerbe:
Concerning adapting the other examples: Yes, we don't need to do that in this PR and I also prefer multiple smaller PRs when it comes to that 👍 

Looking forward!!

jerzakm:
I updated docs to a new format with a tweakpane. I also added a few environment files, I hope that's ok.

packages/extras/src/lib/components/Environment/Environment.svelte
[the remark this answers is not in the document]
jerzakm: I tried messing with loader logic at 2am and made some unintended changes. Disregard this for now, it works :)

grischaerbe (approved):
Thank you so much for your PR! Great work!

jkukatzki:
Very nice, it was something I wasn't looking forward to doing myself.
The isBackground boolean doesn't seem to work for me in the embedded example and changing textures and then toggling ground projection doesn't update the grounds texture.
I tried finding the caveat for isBackground but this looks correct so I didn't get any further:
https://github.com/threlte/threlte/blob/cc9f2bf799c05d51bd7dc39ee10172534892b3fa/packages/extras/src/lib/components/Environment/Environment.svelte#L79-L82

PR #157, opened 2022-10-24
useTrail prototype

grischaerbe:
Port of https://github.com/pmndrs/drei#usetrail


grischaerbe:
Preview: https://threlte-lnumcg2fs-threlte.vercel.app/extras/use-trail

jerzakm:
Fun component. I think that a different example would work better to show off what this does.  Right now it waits for user input before it's visible what it does. What if we changed the example to a cube travelling in a sinusoidal pattern.
- Cube's x += time
- Cube's y += Math.sin(time)
- Move camera x+= time like cube so that the cube appears to be in place

image

jerzakm:
Do we still plan to add it? I started prototyping something similar and then I remembered about this PR.

jerzakm:
Closing because it's really old. If anyone works on this they should make a fresh PR probably anyway.

Issue #123, opened 2022-08-02
Collapsible categories (packages) in docs


grischaerbe:
Currently, the sidebar is quite long, I think a tad too long. KitDocs does not provide the means to have a collapsible sidebar, so there's *some* effort involved.

ixxie:
I propose an alternative solution where packages are given tabs on the top-bar:

![new-docs-layout](https://user-images.githubusercontent.com/20320695/230765859-0679bf3e-ae0a-48ec-9efb-0104788c2b9e.jpg)

Here, *overview* corresponds to the current *learn* tab of the left sidebar.

Motivation:

1.  Makes it easy to see the overall structure of the docs (and threlte) and to jump between packages
2.  Declutters the sidebar without the need for accordion-style collapsible entries
3.  Makes it feasible to put the page navigation inside the left side-bar under the corresponding page, creating space for additional features, like an eventual AI chatbot

PR #158, opened 2022-10-28
Added collapsible categories in docs.

chocola-mint:
Addresses issue #123 (Collapsible categories in docs) by injecting code that modifies KitDocs' sidebar, using Svelte's onMount lifecycle callback. The [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) HTML element is used to minimize the code needed to implement the change.

As KitDocs doesn't include any classes or ids to identify sidebar categories, the resulting solution ended up being very hacky. I've left comments documenting my implementation, though.

Here is a demonstration:
demo

### DIFF
--- a/.changeset/chilled-ducks-know.md
+++ b/.changeset/chilled-ducks-know.md
@@ -1,0 +1,5 @@
+---
+'@threlte/docs': minor
+---
+
+Implemented package-level collapsibles
--- a/.changeset/curvy-apples-give.md
+++ b/.changeset/curvy-apples-give.md
@@ -1,0 +1,5 @@
+---
+'@threlte/docs': minor
+---
+
+Implementing collapsible categories.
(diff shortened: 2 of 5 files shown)


grischaerbe:
Hi @chocola-mint and thanks for your PR! It looks like a good approach. I somehow think it would make more sense to have the packages collapsed or have the collapsed subcategories occupy way less vertical space, or even both?

chocola-mint:
> Hi @chocola-mint and thanks for your PR! It looks like a good approach. I somehow think it would make more sense to have the packages collapsed or have the collapsed subcategories occupy way less vertical space, or even both?

Sounds like a good idea. Here's how it looks now:
demo

grischaerbe:
Hey @chocola-mint, can you try a version where only the top-level packages are collapsed and maybe "@threlte/core" is expanded by default?

chocola-mint:
> Hey @chocola-mint, can you try a version where only the top-level packages are collapsed and maybe "@threlte/core" is expanded by default?

Here it is:
demo

grischaerbe:
Closed due to deprecated docs

PR #164, opened 2022-11-07
Pass the DefaultLoadingManager to all hooks.

grischaerbe:
This will be subject to change in the future, for now this works well enough.


grischaerbe:
The component `<GLTF>` as well as the hook `useGltf` were for some reason not participating in the DefaultLoadingManager without explicitly stating it in the constructor.

grischaerbe:
In the future, it should be possible to pass a `LoadingManager` to a hook or a component, but for now this solves most use cases.

PR #169, opened 2022-11-13
fix(docs): fixes examples menu styling in navigation

SandipNirmal:
Addresses navigation bug where **Examples** is not highlighted also footer navigation is broken.


grischaerbe (approved):
Good catch! Thanks a lot :)

PR #179, opened 2022-11-18
Add glTF Meshopt decoding

Nicell:
Confirmed working with meshopt compressed glTFs!

chrome_RYnGvsFdga

gLTF metadata:
```js
{"buffers":[{"byteLength":536268},{"byteLength":1380444,"extensions":{"EXT_meshopt_compression":{"fallback":true}}}],"asset":{"version":"2.0","generator":"gltfpack 0.18"}
```

### DIFF
--- a/.changeset/slimy-bottles-do.md
+++ b/.changeset/slimy-bottles-do.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': minor
+---
+
+Add support for glTF Meshopt decoding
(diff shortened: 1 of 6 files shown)


Nicell:
@grischaerbe Thanks for the detailed feedback. I fixed the `useGltf` docs and added a compression section to the `GLTF` component docs. I did use "Meshopt" in titles, but it looks like the creator stylizes the name as "meshopt", so I didn't capitalize it in sentences. Just wanted to make that clear 😄

Nicell:
For fun, I measured the performance of my app before and after using meshopt with gltfpack instead of DRACO. I see a 60% reduction in frame times for my use case! My project now doesn't crash on my phone 😅

grischaerbe (approved):
Thanks so much <3

PR #184, opened 2022-11-27
Fix for canvas overflowing parent in flex containers

lauriegriffiths:
I took out the window resize listener and replaced it with a ResizeObserver to handle dynamic changes to the parent element size.

### DIFF
--- a/.changeset/honest-sheep-wait.md
+++ b/.changeset/honest-sheep-wait.md
@@ -1,0 +1,5 @@
+---
+'@threlte/core': patch
+---
+
+fixed 3D scene resizing inside flex layouts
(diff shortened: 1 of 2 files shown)


grischaerbe (changes requested):
Hi @lauriegriffiths and thanks for your contribution! I like the idea a lot!
The problem with your changes and why it's not deploying is that `ResizeObserver` is not available on the server. Also, we should have a fallback to the regular resize event on `window`. You can opt for just returning early if `typeof window === 'undefined'` and return the same shape, I guess that's what I'd do.

jirihon:
I think this is solved now by #470 :)

grischaerbe:
Yes, fixed by #470 :) Thanks nevertheless!

PR #186, opened 2022-11-27
Sparkles prototype

DaanKorver:
Port of https://github.com/pmndrs/drei#sparkles

### DIFF
--- a/packages/extras/src/lib/components/Sparkles/Sparkles.svelte
+++ b/packages/extras/src/lib/components/Sparkles/Sparkles.svelte
@@ -1,0 +1,102 @@
+<script lang="ts">
+	import { useFrame, Three  } from '@threlte/core';
+	import {
+		Vector2,
+		Vector3,
+		Vector4,
+		Color,
+		type ColorRepresentation,
+		ShaderMaterial
+	} from 'three';
+	import { randFloatSpread } from 'three/src/math/MathUtils';
+	import { SparkleShader } from './Shader';
+  import { BufferGeometry } from 'three'
+  import { BufferAttribute } from 'three'
+  import { Points } from 'three'
+	import type { SparklesProperties } from '../../types/components';
+
+	export let count: SparklesProperties['count'] = 100;
+	export let speed: SparklesProperties['speed'] = 1;
+	export let opacity: SparklesProperties['opacity'] = 1;
+	export let color: SparklesProperties['color'] = 100;
+	export let size: SparklesProperties['size']= Math.random();
+	export let scale: SparklesProperties['scale'] = 1;
+	export let noise: SparklesProperties['noise'] = 1;
+
+	const isVector = (v: any): v is Vector2 | Vector3 | Vector4 =>
+		v instanceof Vector2 || v instanceof Vector3 || v instanceof Vector4;
+
+	const isFloat32Array = (def: any): def is Float32Array =>
+		def && (def as Float32Array).constructor === Float32Array;
+
+	const normalizeVector = (v: any): number[] => {
+		if (Array.isArray(v)) return v;
+		else if (isVector(v)) return v.toArray();
+		return [v, v, v] as number[];
+	};
+
+	const expandColor = (v: Color) => [v.r, v.g, v.b];
+
+	const positions = Float32Array.from(
+		Array.from({ length: count }, () => normalizeVector(scale).map(randFloatSpread)).flat()
+	);
+
+	const usePropAsIsOrAsAttribute = <T extends any>(
+		count: number,
+		prop?: T | Float32Array,
+		setDefault?: (v: T) => number
+	) => {
+		if (prop !== undefined) {
+			if (isFloat32Array(prop)) {
+				return prop as Float32Array;
+			} else {
+				if (prop instanceof Color) {
+					const a = Array.from({ length: count * 3 }, () => expandColor(prop)).flat();
+					return Float32Array.from(a);
+				} else if (isVector(prop) || Array.isArray(prop)) {
+					const a = Array.from({ length: count * 3 }, () => normalizeVector(prop)).flat();
+					return Float32Array.from(a);
+				}
+
+				return Float32Array.from({ length: count }, () => prop as unknown as number);
+			}
+		}
+		return Float32Array.from({ length: count }, setDefault!);
+	};
+
+	let opacities = usePropAsIsOrAsAttribute<number>(count, opacity);
+	let sizes = usePropAsIsOrAsAttribute<number>(count, size, Math.random);
+	let speeds = usePropAsIsOrAsAttribute<number>(count, speed);
+	let noises = usePropAsIsOrAsAttribute<typeof noise>(count * 3, noise);
+	let colors = usePropAsIsOrAsAttribute<ColorRepresentation>(
+				color === undefined ? count * 3 : count,
+				!isFloat32Array(color) ? new Color(color) : color,
+				() => 1
+			)
+
+	const material = new ShaderMaterial({
+		uniforms: SparkleShader.uniforms,
+		vertexShader: SparkleShader.vertextShader,
+		fragmentShader: SparkleShader.fragmentShader,
+		transparent: true,
+		depthWrite: false,
+	})
+
+	const geometry = new BufferGeometry()
+
+	geometry.setAttribute('position', new BufferAttribute(positions, 3))
+	geometry.setAttribute('size', new BufferAttribute(sizes, 1))
+	geometry.setAttribute('opacity', new BufferAttribute(opacities, 1))
+	geometry.setAttribute('speed', new BufferAttribute(speeds, 1))
+	geometry.setAttribute('color', new BufferAttribute(colors, 3))
+	geometry.setAttribute('noise', new BufferAttribute(noises, 3))
+
+	useFrame((state) => (material.uniforms.time.value = state.clock.elapsedTime));
+
+</script>
+
+<Three
+type={Points}
+{material}
+{geometry}
+/>
--- a/packages/extras/src/lib/index.ts
+++ b/packages/extras/src/lib/index.ts
@@ -12,6 +12,7 @@ export { default as GLTF } from './components/GLTF/GLTF.svelte'
 export { default as Disposables } from './components/Disposables/Disposables.svelte'
 export { default as ContactShadows } from './components/ContactShadows/ContactShadows.svelte'
 export { default as Environment } from './components/Environment/Environment.svelte'
+export { default as Sparkles } from './components/Sparkles/Sparkles.svelte'
 
 // text component
 export { default as Text } from './components/Text/Text.svelte'
--- a/packages/extras/src/lib/types/components.ts
+++ b/packages/extras/src/lib/types/components.ts
@@ -6,7 +6,14 @@ import type {
   Object3DInstanceProperties,
   Scale
 } from '@threlte/core'
-import type { Camera, ColorRepresentation, Material, Object3D, TextureEncoding } from 'three'
+import type {
+  Camera,
+  ColorRepresentation,
+  Material,
+  Object3D,
+  TextureEncoding,
+  Vector3
+} from 'three'
 import type { Text as TextType } from './types'
 
 export type EdgesProperties = Omit<LineSegmentsProperties, 'geometry' | 'material'> &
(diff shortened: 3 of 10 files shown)


packages/extras/src/lib/types/components.ts
```diff
@@ -146,3 +153,20 @@ export type EnvironmentProperties = {
    */
   encoding?: TextureEncoding
 }
+
+export type SparklesProperties = {
```
grischaerbe, line 157: Export these types from packages/extras/src/lib/index.ts as well 👍

grischaerbe (changes requested):
Hey @DaanKorver,

thanks for your PR! There are some changes to make but all in all it's a pretty solid PR, thanks!!

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
```diff
@@ -0,0 +1,102 @@
...
+	geometry.setAttribute('opacity', new BufferAttribute(opacities, 1))
+	geometry.setAttribute('speed', new BufferAttribute(speeds, 1))
+	geometry.setAttribute('color', new BufferAttribute(colors, 3))
+	geometry.setAttribute('noise', new BufferAttribute(noises, 3))
+
+	useFrame((state) => (material.uniforms.time.value = state.clock.elapsedTime));
+
+</script>
+
+<Three
```
grischaerbe, line 98: In order to be able to position the Sparkles with `<Sparkles position={[10, 10, 10]} >` you can pass them $$restProps here.

grischaerbe: And adjust the types of the component to extend the types of `<Three type={Points}>`

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
```diff
@@ -0,0 +1,102 @@
...
+  import { Points } from 'three'
+	import type { SparklesProperties } from '../../types/components';
+
+	export let count: SparklesProperties['count'] = 100;
+	export let speed: SparklesProperties['speed'] = 1;
+	export let opacity: SparklesProperties['opacity'] = 1;
+	export let color: SparklesProperties['color'] = 100;
+	export let size: SparklesProperties['size']= Math.random();
+	export let scale: SparklesProperties['scale'] = 1;
+	export let noise: SparklesProperties['noise'] = 1;
```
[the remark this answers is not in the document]
DaanKorver: Currently only works with remount. Should I make them reactive.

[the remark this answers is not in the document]
DaanKorver: Should I use the implementation of useMemo we have?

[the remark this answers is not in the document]
grischaerbe: The implementation details are up to you as long as it works :) I guess memoization is not needed in this case, regular reactive statements will do just fine.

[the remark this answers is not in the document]
DaanKorver: Using `$: reactiveVar = someValue` doesn't work on initial render since it is undefined. When checking with if statements to see if the value is set then it will.

I've tried doing the following and that works:
```js
// All these variables are 'undefined' at initial render for some reason.
$: opacities = usePropAsIsOrAsAttribute<number>(count, opacity);
$: sizes = usePropAsIsOrAsAttribute<number>(count, size, Math.random);
$: speeds = usePropAsIsOrAsAttribute<number>(count, speed);
$: noises = usePropAsIsOrAsAttribute<typeof noise>(count * 3, noise);
$: colors = usePropAsIsOrAsAttribute<ColorRepresentation>(
			color === undefined ? count * 3 : count,
			!isFloat32Array(color) ? new Color(color) : color,
			() => 1
		)

$: {
   // Waiting to see if all the variables are set and then applying them to the geometry
   if(opacities && sizes && speeds && noises && colors) {
        geometry.setAttribute('position', new BufferAttribute(positions, 3))
        geometry.setAttribute('size', new BufferAttribute(sizes, 1))
	geometry.setAttribute('opacity', new BufferAttribute(opacities, 1))
	geometry.setAttribute('speed', new BufferAttribute(speeds, 1))
	geometry.setAttribute('color', new BufferAttribute(colors, 3))
	geometry.setAttribute('noise', new BufferAttribute(noises, 3))
   }
}
```

Even though it works, I feel like this isn't the right way of doing it. Do you happen to know a beter solution for this?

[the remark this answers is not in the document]
grischaerbe: If it's necessary to set all values, then yes, this approach is fine and valid. From my understanding you provide default values for all props in question, so the check if everything is set is redundant, right? I'd probably make the call to `geometry.setAttribute('name', new BufferAttribute(variable, 3))` a function:

```ts
const setAttribute = (name: string, value: any, size: number) => {
  geometry.setAttribute(name, new BufferAttribute(value, size))
}
```
and use it like this:

```ts
$: setAttribute('noises', usePropAsIsOrAsAttribute<typeof noise>(count * 3, noise), 3)
```

Would that suite the use case? Your call :)

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
grischaerbe: So the idea how component types are structured changes a bit with v5. Instead of placing `SparklesProperties` in types/components you need to place the file "Sparkles.svelte.d.ts" right next to "Sparkles.svelte". It's content looks like this:

```ts
	import type { Events, Props, Slots } from '@threlte/core'
	import { SvelteComponentTyped } from 'svelte'
	import type { Points, Vector3, ColorRepresentation } from 'three'

	export default class Grouped extends SvelteComponentTyped<
		Props<Points> & {
			/** Number of particles (default: 100) */
			count: number
			/** Speed of particles (default: 1) */
			speed: number | Float32Array
			/** Opacity of particles (default: 1) */
			opacity: number | Float32Array
			/** Color of particles (default: 100) */
			color: ColorRepresentation | Float32Array
			/** Color of particles (default: 100) */
			size: number | Float32Array
			/** The space the particles occupy (default: 1) */
			scale: number | [number, number, number] | Vector3
			/** The space the particles occupy (default: 1) */
			noise: number | [number, number, number] | Vector3 | Float32Array

			/** Experimental: try to exclude material and geometry */
			material: never
			geometry: never
		},
		Events<Points>,
		Slots<Points>
	> {}
```

This will provide proper type intellisense in Editors like VS Code. Sorry for the confusion, this is a new mechanic also to me.

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
```diff
@@ -0,0 +1,95 @@
...
+	useFrame((state) => (material.uniforms.time.value = state.clock.elapsedTime));
+
+</script>
+
+<Three
+type={Points}
+{material}
+{geometry}
+{...$$restProps}
+/>
```
grischaerbe, line 95: In order to support the mentioned types and the slot prop, this needs to be

```html
<Three
	type={Points}
	{material}
	{geometry}
	{...$$restProps}
	let:ref
>
	<slot {ref} />
</Three>
```

So that you can also nest objects in `<Sparkles>`:

```html
<Sparkles>
	<T.Mesh />
</Sparkles>
```

DaanKorver: Done!

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
grischaerbe: Edit: I changed `Grouped` to `Sparkles` as the exported class name, sorry!

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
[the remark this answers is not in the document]
grischaerbe: Nice!

packages/extras/src/lib/index.ts
```diff
@@ -1,29 +1,27 @@
...
   EdgesProperties,
+  EnvironmentProperties,
   FloatProperties,
-  HTMLProperties,
-  TextProperties,
   GLTFProperties,
-  ContactShadowProperties,
-  EnvironmentProperties
+  HTMLProperties,
+  SparklesProperties,
```
grischaerbe, line 24: Remove this export

packages/extras/src/lib/types/components.ts
```diff
@@ -146,3 +154,20 @@ export type EnvironmentProperties = {
...
+  opacity: number | Float32Array
+  /** Color of particles (default: 100) */
+  color: ColorRepresentation | Float32Array
+  /** Color of particles (default: 100) */
+  size: number | Float32Array
+  /** The space the particles occupy (default: 1) */
+  scale: number | [number, number, number] | Vector3
+  /** The space the particles occupy (default: 1) */
+  noise: number | [number, number, number] | Vector3 | Float32Array
+}
```
grischaerbe, line 173: The types are now located in a separate file next to the component itself, remove the types here.

grischaerbe: (Other components will follow that principle too in the future)

packages/extras/src/lib/components/Sparkles/Sparkles.svelte
```diff
@@ -0,0 +1,98 @@
...
+	import type { SparklesProperties } from '../../types/components'
+	import { SparkleShader } from './Shader'
+
+	export let count: SparklesProperties['count'] = 100;
+	export let speed: SparklesProperties['speed'] = 1;
+	export let opacity: SparklesProperties['opacity'] = 1;
+	export let color: SparklesProperties['color'] = 100;
+	export let size: SparklesProperties['size']= Math.random();
+	export let scale: SparklesProperties['scale'] = 1;
+	export let noise: SparklesProperties['noise'] = 1;
```
grischaerbe, line 18: Removing the types from "types/components.ts" will render these prop definitions invalid. Can you think of a good solution? Still, this mechanic is also new to me and this PR is kind of an important testbed to think if the most ergonomic way of preparing the types.

What do you think of exporting the props from Sparkles.svelte.d.ts:

```ts
import type { Events, Props, Slots } from '@threlte/core/src/lib/three/types'
import { SvelteComponentTyped } from 'svelte'
import type { ColorRepresentation, Points, Vector3 } from 'three'

export type SparklesProps = Props<Points> & { … }

export default class Sparkles extends SvelteComponentTyped<
  SparklesProps,
  Events<Points>,
  Slots<Points>
> {}
```

and importing them in the component:

```html
<script>
	import type { SparklesProps } from './Sparkles.svelte'

	export let count: SparklesProps['count'] = 1
</script>
```

It *looks* odd and cyclical but essentially it's only types but needs a test drive.

DaanKorver: Did that. Should I remove the SparklesProperties from types/components.ts since it is not being used now?

grischaerbe: 👍 yep, needs to be removed!

DaanKorver: Should be up to date now

grischaerbe:
Thanks for resolving all these things, it looks good now, type-wise 👍 Again sorry for all the confusion, it shows we need to have docs for fleshing out components.
Unfortunately I think it also broke some things. Running the docs locally I get a lot of these errors and the sparkles don't show up:

image

DaanKorver:
Mmmm, not getting this warning myself, I am using brave myself

DaanKorver:
https://discourse.threejs.org/t/vertex-buffer-is-not-big-enough-for-the-draw-call-only-on-mac-computers/43628/2

DaanKorver:
Kinda stuck on this one. And since I am not using MacOS, it is hard debugging

grischaerbe:
Sorry for letting you wait so long, I'm a bit busy at the moment. I'm on a mac and the first deployment worked for me, so I guess there's an error in the bufferAttribute sizes. Can you double check?

DaanKorver:
Hey @grischaerbe 
After a long time I took the time to fix the issue. Had a lot of other work to do in the meanwhile. Hopefully it's not too late to merge this PR for the upcoming version of threlte, let me know if there's anything I can do to help.

DefinitelyMaybe:
@DaanKorver Hey looks like we took a heck of a detour here, sorry. Current dev work has moved into the `next` branch. What would you like to do here?

Two options I see are:
- merging next branch then updating it to svelte 5 or
- recreating in another PR

PR #213, opened 2022-12-17
Grid: remove axes, include color mapping and encoding

jerzakm:
- Removed axes from types, docs and the example -


jerzakm:
Component:
- removed the changes to axes and we're keeping them for now

Example:
- tweakpane starts closed because it was big af
- thicker lines, more visible colours

grischaerbe:
Thanks ❤️

PR #215, opened 2022-12-20
Environment - V5 Types

jerzakm:


packages/extras/src/lib/components/Environment/Environment.svelte
```diff
@@ -15,15 +14,18 @@
   } from 'three'
   import { HDRCubeTextureLoader } from 'three/examples/jsm/loaders/HDRCubeTextureLoader'
   import { RGBELoader } from 'three/examples/jsm/loaders/RGBELoader'
+  import type { GridProps } from '../Grid/Grid.svelte'
```
grischaerbe, line 17: Wrong import path

packages/extras/src/lib/components/Environment/Environment.svelte.d.ts
```diff
@@ -0,0 +1,36 @@
+import type { Scale } from '@threlte/core'
```
grischaerbe, line 1: This type will change in the transition to v5 and will probably just vanish. Can you change the type of this prop to something more v5-ish?

packages/extras/src/lib/components/Environment/GroundProjectedEnv.svelte
```diff
@@ -1,11 +1,13 @@
 <script lang="ts">
-  import type { EnvironmentProperties } from '$lib/types/components'
   import { useThrelte } from '@threlte/core'
   import { onDestroy } from 'svelte'
   import type { Texture } from 'three'
   import { GroundProjectedEnv } from 'three/examples/jsm/objects/GroundProjectedEnv'
+  import type { GridProps } from '../Grid/Grid.svelte'
```
grischaerbe, line 6: GroundProjectedEnv is not part of the public API, it doesn't need external component types.

grischaerbe (changes requested):
Thanks for your PR ❤️ 
I left some remarks. Please also delete the prop type export from `index.ts` as well :)

packages/extras/src/lib/components/Environment/Environment.svelte.d.ts
jerzakm: Something like this? 

`[scale: number | { x?: number; y?: number; z?: number } | number[]](scale: number | Partial<Vector3> | number[])`

The Scale type is actually pretty handy for these things but I get why we're tossing it.

packages/extras/src/lib/components/Environment/GroundProjectedEnv.svelte
jerzakm: Does this mean you don't want me to import props in GroundProjection sub-component? I did it to easier reason with types when developing it.

```ts
  import type { EnvironmentProps } from './Environment.svelte'

  type Props = EnvironmentProps

  export let groundProjection: Props['groundProjection']
  export let currentEnvMap: Texture
  ```

PR #216, opened 2022-12-20
@threlte/extras: ContactShadows Conversion

grischaerbe:


jerzakm:
*Re-post from discord to remember and keep discussion in one place*

I was looking into porting <Edges/> to V5. It's basically an abstraction on top of a 3 liner like this. Do we even need it anymore? I think deprecating and replacing it with a docs example entry might be better. 
```jsx
<T.LineSegments>
    <T.EdgesGeometry args={[geometry, threshold]} />
</T.LineSegments>
```

PR #222, opened 2022-12-20
WIP: Add clipping example

jpaquim:
Working on translating https://github.com/mrdoob/three.js/blob/dev/examples/webgl_clipping.html?ts=2


jerzakm:
Closing because it's really old. We can reopen later or make a fresh PR.

PR #227, opened 2022-12-22
V5 Transition @extras/Float

jerzakm:
Just letting you know I'll be starting work on it later today


packages/extras/src/lib/components/Float/Float.svelte.d.ts
```diff
@@ -0,0 +1,13 @@
+import type { Events, Props, Slots } from '@threlte/core'
+import { SvelteComponentTyped } from 'svelte'
+import type { Group } from 'three'
+
+export type FloatProps = Props<Group> & {
+  speed: number | [x: number, y: number, z: number]
+  floatIntensity: number | [x: number, y: number, z: number]
+  floatingRange: [number, number] | [x: [number, number], y: [number, number], z: [number, number]]
+  rotationIntensity: number | [x: number, y: number, z: number]
+  rotationSpeed: number | [x: number, y: number, z: number]
```
grischaerbe, line 10: Mark them as optional.

packages/extras/src/lib/components/Float/Float.svelte
```diff
@@ -1,89 +1,83 @@
...
-  export let viewportAware: FloatProperties['viewportAware'] = false
-  export let inViewport: FloatProperties['inViewport'] = false
-  export let castShadow: FloatProperties['castShadow'] = undefined
-  export let receiveShadow: FloatProperties['receiveShadow'] = undefined
-  export let frustumCulled: FloatProperties['frustumCulled'] = undefined
-  export let renderOrder: FloatProperties['renderOrder'] = undefined
-  export let visible: FloatProperties['visible'] = undefined
-  export let userData: FloatProperties['userData'] = undefined
-  export let dispose: FloatProperties['dispose'] = undefined
+  type Props = Required<FloatProps>
```
grischaerbe, line 6: Change type name to `$$Props`

packages/extras/src/lib/components/Float/Float.svelte.d.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export default class Float extends SvelteComponentTyped<FloatProps, Events<Group>, Slots<Group>> {}
```
grischaerbe, line 13: export events and slots the same way as the props. Usage in component ↑

packages/extras/src/lib/components/Float/Float.svelte
grischaerbe, line 6: Add the types for events and slots:

```ts
type $$Events = FloatEvents
type $$Slots = FloatSlots
```

This will ensure that the Svelte LSP can pick up on mistakes in the implementation. Currently this is especially useful on slot types.

packages/extras/src/lib/components/Float/Float.svelte
```diff
@@ -1,89 +1,83 @@
...
-  {userData}
-  {dispose}
-  {viewportAware}
-  bind:inViewport
-  on:viewportenter
-  on:viewportleave
+<T.Group
+  position={floatPosition}
+  rotation={floatRotation}
+  {...$$restProps}
```
grischaerbe, line 80: Add `let:ref` here and pass it in the slot: `<slot {ref} />`.

grischaerbe (changes requested):
Thanks ❤️ 
I found some things :)

PR #248, opened 2022-12-26
v6: Move <Audio*> components to extras

jpaquim:


packages/extras/src/lib/audio/Audio/Audio.svelte
```diff
@@ -0,0 +1,79 @@
...
+  if (!listener) {
+    throw new Error(`No Audiolistener with id ${id} found.`)
+  }
+
+  export const audio = new ThreeAudio(listener)
+</script>
+
+<Three
+  type={audio}
+  {position}
```
grischaerbe, line 49: use `{...$$restProps} instead of {position}, etc.`

grischaerbe: This includes removing these props from the component script block

packages/extras/src/lib/audio/AudioListener/AudioListener.svelte
```diff
@@ -0,0 +1,61 @@
...
+  addAudioListener(listener, id)
+
+  onDestroy(() => {
+    removeAudioListener(id)
+  })
+</script>
+
+<Three
+  type={listener}
+  {position}
```
grischaerbe, line 44: Same here, please remove `{position}` et al and insert these via `{...$$restProps}` and remove these props from the script block

packages/extras/src/lib/audio/AudioListener/AudioListener.svelte.d.ts
```diff
@@ -0,0 +1,13 @@
+import type { Props } from '@threlte/core'
+import type { SvelteComponentTyped } from 'svelte'
+import type { AudioListener as ThreeAudioListener } from 'three'
+
+export type AudioListenerProps = Props<ThreeAudioListener> & {
+  inViewport?: boolean
```
grischaerbe, line 6: viewport awareness will be removed in v6 and inserted back via a plugin.

packages/extras/src/lib/audio/PositionalAudio/PositionalAudio.svelte
```diff
@@ -0,0 +1,96 @@
...
+        directionalCone.coneOuterAngle,
+        directionalCone.coneOuterGain
+      )
+    }
+  }
+</script>
+
+<Three
+  type={audio}
+  {position}
```
grischaerbe, line 67: Same here, please remove `{position}` et al and insert these via `{...$$restProps}` and remove these props from the script block

packages/extras/src/lib/hooks/useAudioListener.ts
```diff
@@ -1,5 +1,5 @@
 import type { AudioListener } from 'three'
-import { useThrelteAudio } from './useThrelteAudio'
+import { useThrelteAudio } from '@threlte/core'
```
grischaerbe, line 2: The audio context and its hook should be moved to extras as well.

grischaerbe: The problem then is: what component is providing that context? Will there be a component `<Extras>` that you wrap your entire scene in that provides all necessary contexts used in `@threlte/extras`? Is this a use case for a root plugin?

grischaerbe (changes requested):
Thanks @jpaquim for your PR!
I added some comments, for some I'm not sure how to handle it concerning the contexts, others are clear.

packages/extras/src/lib/audio/AudioListener/AudioListener.svelte.d.ts
jpaquim: Removed viewport-related props and events

grischaerbe:
The deployment fails because of this error: 

```
@threlte/docs-next:build: "useAudioListener" is not exported by "../../packages/core/src/lib/index.ts", imported by "src/examples/core/positional-audio/Turntable.svelte".
```

Other than that, is this ready to be merged? 🚀

grischaerbe (approved):
LGTM 🚀 
Thanks so much!

PR #249, opened 2022-12-27
Rapier heightfield fix?

jerzakm:
I tried to make a heightfield today and I had a lot of issues initializing it (error below).  I thought our <Debug/> is incorrect but it seems there is a problem with a function that makes Collider constructor args.

I commented out one line that was setting scale.x in  `packages/rapier/src/lib/lib/scaleColliderArgs.ts`. I also made an example to debug. 

Without this PR this code didn't run.
```
<RigidBody type={'fixed'}>
	<Collider shape={'heightfield'} args={[nsubdivs, nsubdivs, heights, scale]} />
</RigidBody>
```

image

packages/rapier/src/lib/lib/scaleColliderArgs.ts

### DIFF
--- a/apps/docs/src/examples/terrain-heightfield/App.svelte
+++ b/apps/docs/src/examples/terrain-heightfield/App.svelte
@@ -1,0 +1,11 @@
+<script lang="ts">
+	import { Canvas } from '@threlte/core'
+	import { World } from '@threlte/rapier'
+	import Scene from './Scene.svelte'
+</script>
+
+<Canvas>
+	<World>
+		<Scene />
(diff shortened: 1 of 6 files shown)


grischaerbe:
I guess this PR is ready for review, right? The deployment seems to show it's working. So ready to merge?

jerzakm:
Yes, it's ready for review. I just wasn't sure at a time if what I encountered was really a bug since I didn't work with rapier much.

grischaerbe:
Cool, I think this works! Thanks for your PR!

PR #250, opened 2023-01-08
[refactor] Sets instead of arrays

sawyerclick:
Small performance improvement here. Replace arrays here that use `.includes()` with Set and `.has()`, which has a better time complexity.


grischaerbe:
Hey @sawyerclick, thanks for your PR! You're right, this should be faster 👍 
Can you change the PR to merge against the branch `threlte-v5-transition`? It's the branch that eventually will be threlte v6 (sorry for the bad naming of that branch).

sawyerclick:
@grischaerbe base is changed to `threlte-v5-transition`!

grischaerbe:
Ah I guess you need to rebase also. Sometimes it's easier and cleaner to start off with a fresh branch, cherry pick the old commit and do a force push.

sawyerclick:
that rebase got... complicated. To avoid the other file changes going in with this PR I will close this one up and open another

PR #251, opened 2023-01-08
Refactor: Sets instead of arrays

sawyerclick:
#250 but for the right base branch. Refactor `useProps.ts` to utilize Set and its `.has()` method over array with its `.includes()` method. Set has a better time complexity, make this a small performance gain.


grischaerbe:
Awesome, thanks!

PR #253, opened 2023-01-11
v6: Move <OrbitControls> to extras

jpaquim:


grischaerbe:
Hey @jpaquim, I ran into this while porting other stuff, went ahead and completely forgot about this PR. So it's in there now 😄

PR #259, opened 2023-01-20
Remove unnecessary store get calls

jpaquim:
Follow-up to the discussion: https://discord.com/channels/985983540804091964/986233334747254854/1064575611332473019

Given that `store.subscribe()` synchronously runs its callback, it's not necessary to call `get(store)` immediately beforehand.

### DIFF
--- a/.changeset/five-buckets-search.md
+++ b/.changeset/five-buckets-search.md
@@ -1,0 +1,5 @@
+---
+'@threlte/core': patch
+---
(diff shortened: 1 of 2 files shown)


grischaerbe:
Absolutely right, thanks for the contribution! Can you add a quick changeset?

jpaquim:
@grischaerbe sure, done!

PR #264, opened 2023-01-28
New Docs Prototype

grischaerbe:
After a Discord discussion, we kind of settled with using [Astro](https://astro.build) for building the new docs. This is a first prototype.

### Structure

The page structure is leaning on the one from Astro's own docs:

- Home
- Showcase
- Docs
  - Learn
  - Reference
  
### Notes:

- I'd like to keep the differentiation between "Learn" and "Reference" because the new `<T>`/`<Three>` requires education that goes beyond "This is component X and these are its props". This is what this section is for.
"Reference" is where component and hook references are kept.
- The "Docs" pages themselves (so everything in /docs/*) have a special layout called DocsLayout. Everything else is using the "MainLayout"
- I'm not entirely sure how to persist state between page loads/reloads/navigation. This is important for components like the LeftSidebar and is currently implemented via [persistent svelte stores](https://github.com/MacFJA/svelte-persistent-store) that save to the session storage.


grischaerbe:
closing in favor of #274

PR #266, opened 2023-02-02
Preprocessor fix

grischaerbe:
The types were a bit off.


grischaerbe:
Hm, something seems off here in general. Needs more investigation.

PR #274, opened 2023-02-06
Next docs prototype

grischaerbe:


grischaerbe:
Prototype of next documentation

PR #279, opened 2023-02-10
Docs Layout Mobile

jerzakm:
wip, don't look - dragons inside


jerzakm:
Let's merge to keep in sync, I'll be short on time for 2-3 days now. I'll rework "On this page" menu when I get a chance, I got hung up on making it css only, but then had to use ugly hacks to make it work.
Quickly tested on 2 phones and a few browsers, seems fine.

PR #281, opened 2023-02-15
Prettier enhanced for docs: Svelte, Astro, Tailwind (with sorting)

jerzakm:
Like discussed, added tailwind support for prettier. It seems prettier is a bit quirky with pnpm and I had to add format overrides

as per:
https://github.com/pnpm/pnpm/issues/4700
https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/113

(and a few others I stumbled on)

I temporarily added two extra files in the commit to test the functionality and compare to tailwind example. Seems to work fine for me.
(TempPrettierTest.svelte, TempPrettierTest.astro)

I'm not at all well versed in prettier config, let me know what you think. If this is fine then I'll remove the temp files to get it ready to merge.

As a side effect turbo and prettier got bumped but I think that should be fine.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -4,5 +4,21 @@
   "trailingComma": "none",
   "printWidth": 100,
   "semi": false,
-  "singleAttributePerLine": true
+  "singleAttributePerLine": true,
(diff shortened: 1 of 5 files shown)


grischaerbe:
LGTM 🚀

PR #290, opened 2023-02-21
useLoader

grischaerbe:
`useLoader` has a few flaws in the current implementation:

- assets are not cached: Loading an asset multiple times (e.g. textures or models) results in parsing the raw asset multiple times, which is a performance hit both initially and on the rendering side of an app.
- the signature was a bit funky: Provide the prototype as well as a function to instantiate the loader.

It also has upsides: textures could be used directly, as this loader returns a `THREE.Texture` immediately (not on load).

This new implementation caches results and on subsequent requests returns the cached result rather than parsing it again. The signature is also significantly reduced as only the prototype of the loader needs to be provided.

```ts
const texture = useLoader(TextureLoader, 'path/to/texture.png')
```

A loader can be *extended*:
```ts
const model = useLoader(GLTFLoader, 'path/to/model.glb', {
  extend: (loader) => {
    // do something with the loader
  }
})
```

Listening to events of a loader:
```ts
const texture = useLoader(TextureLoader, 'path/to/texture.png', {
  onLoad: (result) => {},
  onProgress: (event) => {},
  onError: (event) => {
    // an error is silent by default. To run into SvelteKit's error boundaries, throw here
    throw new Error('Texture load error')
  }
})
```

The result of `useLoader` is a store that gets populated as soon as the resource is loaded.


grischaerbe:
The most recent implementation can be found as `useLoader3` and will eventually be exported as `useLoader` after some cleanup and adapting existing examples and the docs.

edil-it-them:
Niiice! From where do we import TextureLoader?

grischaerbe:
> Niiice! From where do we import TextureLoader?

Just like it did before, `@threlte/core` or `@threlte/extras` (undecided about that) will provide convenience exports for textures (`useTexture`) or .gltf models `useGltf`. Apart from that I grew fond of `three-stdlib` for everything that normally sits in `three/examples`.

PR #293, opened 2023-03-01
Docs: Fix core-transition docs to match preprocessThrelte configuration docs

jasoneliu:
The [core-transition #extending-the-processor](https://threlte.xyz/core-transition#extending-the-preprocessor) docs are incorrect, omitting quotes for import names. I've updated the core-transition docs to match the [preprocessThrelte #configuration](https://threlte.xyz/preprocess/preprocessThrelte#configuration) docs, which are accurate.

Before:

After:

### DIFF
--- a/.changeset/short-houses-protect.md
+++ b/.changeset/short-houses-protect.md
@@ -1,0 +1,5 @@
+---
+'@threlte/docs': patch
+---
+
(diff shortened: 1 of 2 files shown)


grischaerbe:
Hey @jasoneliu, good catch! Thanks for your PR! ❤️

PR #311, opened 2023-03-15
Removing and merging contexts, implementing color management

grischaerbe:
- removed "root" and "disposal" and "render" contexts and merged them into the existing main (and exposed) "threlte" context as well as in a new "internal" context. This new internal context is _not_ exposed and can be used internally via the hook "useThrelteInternal".
- Added general color management (three.js > r147) and more concise state management in Canvas.svelte
- Made some properties of the state reactive stores (`CurrentWritable`) so that they can be reacted on and also be used in a loop


grischaerbe:
Giving it another try in #312

PR #314, opened 2023-03-17
New type for prop `colorSpace` on `<Canvas>`

grischaerbe:
The type of the `colorSpace` property on `<Canvas>` should be `THREE.ColorSpace`, as that's what Three.js is transitioning to. Right now, for older Three.js versions, we need to map the provided `THREE.ColorSpace` to a `THREE.TextureEncoding` to be set on `renderer.encoding`


packages/core/src/lib/Canvas.svelte
```diff
@@ -38,7 +37,7 @@
   /**
    * @default sRGBEncoding
```
donmccurdy, line 38: nit:

```suggestion
   * @default 'srgb'
```

packages/core/src/lib/lib/useRenderer.ts
```diff
@@ -1,16 +1,34 @@
...
+ * the new `colorSpace` to the old `encoding`.
+ */
+const colorSpaceToEncoding: Record<ColorSpace, TextureEncoding> = {
+  srgb: sRGBEncoding,
+  'srgb-linear': LinearEncoding,
+  '': LinearEncoding
+}
+
+const rendererHasColorSpaceProperty = (renderer: any): renderer is { colorSpace: string } => {
+  return renderer.colorSpace !== undefined
```
donmccurdy, line 29: The plan is for this property to be named `renderer.outputColorSpace` rather than `renderer.colorSpace`, to differentiate the "working" color space used in the render pipeline from the "output" color space used when drawing to the canvas. 

Because threlte syntax would have this property on the `<Canvas />` component, including the "output" term felt redundant so I suggested omitting it there.

donmccurdy (commented):
Two suggestions, otherwise looks good to me!

PR #328, opened 2023-03-24
removed transform props from rapier package components

grischaerbe:
Separation of concerns: These components housed too many features


grischaerbe:
Closed because #353 includes these changes.

Issue #332, opened 2023-03-30
three v151 updates GroundProjectedEnv to GroundProjectedSkybox


ngsmcphrsn:
Update import in [GroundProjectedEnv component](https://github.com/threlte/threlte/blob/d6b27bde4309637eaa56bc020acb47a289382c3f/packages/extras/src/lib/components/Environment/GroundProjectedEnv.svelte) to match v151, per https://github.com/mrdoob/three.js/pull/25645.

grischaerbe:
Fixed in `'@threlte/extras@next'`.

PR #343, opened 2023-04-03
environment name changes. update three.js dep

DefinitelyMaybe:
This'll work for #332 but also in it's current version will not work for three.js versions < 151.

I'm thinking about adjusting this to use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import). yay or nay?


grischaerbe:
I'm totally fine with making it a regular import and requiring at least r151 👍 
Can you add a changeset (`npx changeset`) `major` to `extras` indicating that min. r151 is required? I think then we're good to go :) Thanks so much!

DefinitelyMaybe:
cool cool! done :)

PR #353, opened 2023-04-14
@threlte/rapier and @threlte/studio refactoring

grischaerbe:
This PR is a refactoring of the packages `@threlte/rapier` and `@threlte/studio`.

## @threlte/rapier

- Removed transform props from Collider, AutoColliders and RigidBody because they are and won't be in-line with the syntax used on `<T>`.
- `<Collider>` components are now **not** reactive to parent transforms by default. Add `type="dynamic"` to make them reactive.
- A binding `refresh` was added on `<Collider>` and `<AutoColliders>` to refresh the transform and – in regards to `<AutoColliders>` recreate the colliders from the underlying geometry.
- Updated dependencies

## @threlte/theatre

- `<Editable>`
  - Added events `on:create` and `on:change` to listen to prop values
  - Added prop `detach` to detach a SheetObject `onDestroy`
  - Added keyboard shortcut  "g" to change transform space from world to local

- `<Studio>` now has an inner slot that only mounts when the Studio is mounted:
```html
<Studio>
  <!-- Threatre.js Studio is loaded -→
</Studio>
```
- Adde hook `useStudio`
- Updated dependencies


grischaerbe:
Documentation is missing for some of these things … Part of another PR.

PR #357, opened 2023-04-17
Docs: Clippy user assistant (search enhancement)

jerzakm:
(wip, probably not worth looking into the code yet)

The goal here is to add an AI powered Clippy-like assistant that will interact with user and answer their questions.

- not meant as a total search replacement - I think we should still add Algolia or similar for a more traditional searching experience
- currently powered by https://markprompt.com/ but I will also be looking into other solutions. From my limited testing it seems to be handling our docs pretty well and is very quick to bootstrap
- I am currently using an orange dragon model from https://quaternius.com/ . At the moment, I will not put too much work into animations and interactivity of the Clippy model but eventually I think it would be cool to have it fly around the screen, interact with docs and user (waving, pointing to something etc).


jerzakm:
https://user-images.githubusercontent.com/16734228/232451769-de421412-5aa5-4a9e-a81d-e2b3a32bd76c.mp4

grischaerbe:
Works already pretty good! It's not only fun but a highly valuable addition to the docs 👍

grischaerbe:
Thanks for that, works like a charm!  ❤️

PR #362, opened 2023-04-23
Remove old docs, unused custom types

jpaquim:
Remove old docs, remove unused custom types `Position`, `Rotation`, `Scale`, `LookAt`


packages/core/src/lib/index.ts
```diff
@@ -34,16 +34,7 @@ export {
...
-  Position,
-  Scale,
-  Rotation,
-  LookAt,
-  ThrelteUseFrameOptions,
-  ThrelteContext,
-  ThrelteInstance,
-  Size
-} from './types/types'
+export type { ThrelteUseFrameOptions, ThrelteContext, ThrelteInstance, Size } from './types/types'
```
grischaerbe, line 37: `ThrelteInstance` is also unused,

grischaerbe (changes requested):
Hey @jpaquim, thanks for your contribution! Can you remove that one unused type? Then I think we're good to go!

PR #372, opened 2023-04-24
Colocate context types in respective hook files

jpaquim:


grischaerbe:
LGTM! Can you add a `changeset` for core?

jpaquim:
@grischaerbe done, I added it as a patch change

PR #380, opened 2023-04-30
Theatre 6

ixxie:
All items marked as complete include completed MVP docs. Many include new examples for exposition and validation purposes.
  
- [x] `Sequence` store provides reactive and portable wrapping around the Theatre sequence, including unified options handling.
- [x] `<Sequence>` component implemented, exposing the store API through props. The API should be forward compatible with forthcoming multisequence sheets.
- [x] `<Sheet>` simplified since most of the logic has been moved elsewhere. The `Sequence` store is instantiated here and retrieved via `getContext` in children.
- [x] A `useSequence` hook allows retrieval of the store API by the user.
- [x] The `<Project>` component is streamlined and cleaned up a bit.
- [x] A `useObject` hook is provided, which provides a Svelte Action that lets the user animate the DOM.
- [x]  `<Editable>` component is revamped based on @grischaerbe's new design, now called `<SheetObject>`
- [ ]  `<Studio>` is upgraded and complementary extension components are implemented


packages/theatre/src/lib/actions/object.ts
```diff
@@ -0,0 +1,33 @@
...
+    props: Props,
+    sheet: ISheet,
+    sync: (
+      obj: Parameters<Parameters<ISheetObject<Props>['onValuesChange']>[0]>[0]
+    ) => ({ [key: string]: string }),
+    options?: { reconfigure?: boolean }
+  }
+) {
+
+  const obj = params.sheet.object<Props>(params.name, params.props, params.options)
```
grischaerbe, line 18:
```ts
const obj = params.sheet.object<Props>(params.name, params.props, {
  reconfigure: true,
  ...params.options
})
```

packages/theatre/src/lib/components/Sequence/sequence.ts
```diff
@@ -0,0 +1,191 @@
...
+		this.playing = new SequencePlaying(sequence, () => this.play())
+		this.length = new SequenceLength(sequence)
+	}
+	config(options: SequenceOptions, { replay } = { replay: false }) {
+		// update options
+		const updatedOptions: SequenceOptions = { ...this.options, ...options }
+
+		// ensure no unnecessary updates happen
+		const noChange = Object.keys(updatedOptions)
+			.every((key: keyof SequenceOptions) => this.options[key] == updatedOptions[key]) // TODO: fix type of key
```
ixxie, line 40: The key type is wrong here and I don't know how to fix it.

grischaerbe:
## Current Status of `<AutoProps>`

I implemented a pattern I'm currently calling `Transformer`. 

```ts
export const color = createTransformer<Color>().build({
  // by default, Theatre wouldn't be able to handle a THREE.Color directly,
  // so we have to transform it.
  initialize(value) {
    return types.rgba({ r: value.r, g: value.g, b: value.b, a: 1 })
  },
  apply(target, path, value) {
    target[path].r = value.r
    target[path].g = value.g
    target[path].b = value.b
  }
})

```

Every prop is being initialized and applied by a transformer. There's a set of default transformers:

-  colors (shows a color picker)
- degrees (rad2deg conversion)
- euler (rad2deg as compound)
- normalized (for metalness, roughness, opacity, … with range UI)
- side (shows buttons for "front", "back" and "double")
- generic (catch-all)

A user may opt to provide a custom transformer which will be used to transform a prop.

## Example

Bildschirmfoto 2023-06-27 um 16 14 16

*The output of some `<AutoProps>` components collecting some props across a scene graph*

PR #386, opened 2023-05-07
fix diff lines in `your-first-scene.mdx`

ArnaudLier:


grischaerbe:
Thanks @ArnaudLier, good catch!

PR #389, opened 2023-05-11
Update loading-assets.mdx

ajaypillay:
Hi, I've added an example on how to apply multiple textures to different faces of a BoxGeometry instance. It differed from how it was done in React Three Fiber (by passing `"material-0"`, ..., `"material-5"` in the `attach` prop for each material), so I thought this would be worth highlighting in the documentation.


grischaerbe:
We may at some point get to a feature parity when it comes to the attach API but I'm not a fan of the array index access because it's somewhat limited: the array index can only be placed at the end of the attach string, what happens if there's *some* value but not an array, etc. So I find this approach safer from a consumer point of view and I agree that this example has been missing, thanks a lot! ❤️

PR #404, opened 2023-06-07
Rename GroundProjectedEnv to GroundProjectedSkybox

caerus706:
three.js version 0.151.0 renamed GroundProjectedEnv to GroundProjectedSkybox   [#25645](https://github.com/mrdoob/three.js/pull/25645)  

This breaks @threlte/extras when it attempts to import GroundProjectedEnv.  
I renamed the file and the import to work with newer versions of three.js

### DIFF
--- a/packages/extras/package.json
+++ b/packages/extras/package.json
@@ -33,7 +33,7 @@
     "svelte-preprocess": "^4.10.5",
     "svelte-sequential-preprocessor": "^0.0.7",
     "svelte2tsx": "^0.5.9",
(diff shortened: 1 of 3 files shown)


grischaerbe:
Hey @caerus706, the changes you suggest have already been applied to Threlte on the branch `next` which will be Threlte 6 being out of pre-release access in the very near future. I suggest to use Three.js in a version prior to the renaming to use alongside Threlte 5 or update Threlte to `@next` to use with the latest Three.js versions.

PR #405, opened 2023-06-09
fix (docs): maintain scroll position on  page change

midanosi:
The new docs are great! thank you

But I was getting frustrated with my scroll position in the left sidebar being lost after each page change.
So I copied [the code from the Astro docs](https://github.com/withastro/docs/blob/main/src/components/LeftSidebar/LeftSidebar.astro#L87) to fix it

P.S. it seems like all the files in the repo have indentation set to 2 spaces, but the editorconfig file has "indent_style=tab", rather than "indent_style=space", so vscode was automatically converting all the 2-spaces into tabs for me.
Seems a mismatch between editorconfig setting and what is actually in code?

### DIFF
--- a/.changeset/silver-books-kneel.md
+++ b/.changeset/silver-books-kneel.md
@@ -1,0 +1,5 @@
+---
+'@threlte/docs-next': patch
+---
+
+maintain left sidebar scroll position on page change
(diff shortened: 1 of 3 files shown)


grischaerbe:
Nevermind, I'll merge it back in and we'll see if that issue comes up for anybody in the first place :) Thanks for your effort!

midanosi:
> Nevermind, I'll merge it and we'll see if that issue comes up for anybody in the first place :) Thanks for your effort!

Thanks for merging!
I was intending to give what you suggested a go, but didn't as am abroad this week. And I do think that leaving it until it the problem is raised is wise

PR #409, opened 2023-06-16
Add outline demo to docs

midanosi:
Wanted to contribute an example to the docs.
Decided to do a "thing moving behind other thing" demo, a la teammate outlines in Left 4 Dead.
Animated with svelte/motion instead of theatrejs, to keep it as simple as possible.

Compared to R3F, this example is less elegant, because R3F can use the `react/post-processing` and the <Selection> helper, so when those equivalent things are in the threlte ecosystem, then this example could be updated.

### DIFF
--- a/.changeset/afraid-parrots-smoke.md
+++ b/.changeset/afraid-parrots-smoke.md
@@ -1,0 +1,5 @@
+---
+'@threlte/docs-next': patch
+---
+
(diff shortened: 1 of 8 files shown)


grischaerbe:
Awesome! Thanks for the example, looks neat ❤️

PR #410, opened 2023-06-16
Better examples titles

midanosi:
## Problem
I noticed that the example title text is not presented in the page content in the new docs. 
Example of problem in new docs: https://next.threlte.xyz/docs/learn/examples/geometry/terrain
Here, there is no text saying "Terrain with 3D noise", in order to differentiate it with the other terrain example with rapier physics.

## What I have changed
So this PR adds a `subtitle` content field, and adds to it page as a `<h2>`.

### What else could be done
Adding the examples to the sidebar, i.e. setting `showInSidebar:true`. But don't know best way to do that, via maybe nesting? or another example-specific sidebar to the right.  
Probably worth doing only when there are more examples though.


grischaerbe:
Concerning the examples section in the sidebar: If there are at some point too many examples, we should think about adding an examples section as we currently have for the showcase section (which isn't shown at the moment, but prepared).

midanosi:
> the root problem is that the titles of the existing examples are just wrong. For instance the title of the Animation Transitions example should be "Animation Transitions", not "Animation" ... What do you think?
Sounds good, I'll revert the subtitle prop, and just put the example titles in the `title` field instead.

> I would also love to add some context (i.e. page content) to the individual examples, explain a bit why things are made the way they are etc.
yeah agreed, I'll add some explanation I can to the Outline one I just made, and some text to the others if I can think of any

> Concerning the examples section in the sidebar: If there are at some point too many examples, we should think about adding an examples section as we currently have for the showcase section (which isn't shown at the moment, but prepared).
sounds good :+1:

grischaerbe:
Sounds good to me, then I'll leave this PR open. Do you want to add example content in this PR or a separate one?

grischaerbe:
Awesome, thanks so much! ❤️

PR #428, opened 2023-06-21
implement static plugin

### LINKED ISSUE #426: Make option to inject plugin statically
## Background
Dynamically injecting using `injectPlugin` in `@next` can be useful for many usecases, but in most cases, we only uses fixed module like `interactive()`. In this case, using injectPlugin can be annoying because we need to split 'World' and 'Scene' even they are very simple so it can written into one file.

## Proposal
Make `plugins` option into `Canvas`. This option accepts array of plugins. Make some internal component inside Canvas. Canvas just passes plugins properties into that component, and that component calls injectPlugin. It won't change any API for plugin, just making another wrapper method to inject fixed plugins easily.

<hr>

If you think this proposal is reasonable, I'll make PR for it!

seo-rii:
Fixes #426.
## Changes
1. Accepts plugins array and applies them in Canvas component.
2. Changed injectPlugin logic.
(Accepts unnamed plugin from injectPlugin without naming, Allow to rename named plugin, ...)

### DIFF
--- a/apps/docs-next/src/content/learn/advanced/plugins.mdx
+++ b/apps/docs-next/src/content/learn/advanced/plugins.mdx
@@ -6,9 +6,13 @@ order: 0
 
 ## Injecting a Plugin
 
+Plugin are a way to extend the functionality of the `<T>` component. They are a way to inject code into the component without having to fork the codebase.
+
 ### What it looks like
 
-Plugins open up the component `<T>` to external code that will be injected via context into every child instance of a `<T>` component.
+There are two ways to inject a plugin into the `<T>` component. The first one is to use the `injectPlugin` function. This function is available in the `@threlte/core` package.
+
+Like this:
 
 ```ts
 import { injectPlugin } from '@threlte/core'
@@ -17,6 +21,18 @@ injectPlugin('plugin-name', ({ ref, props }) => {
   console.log(ref, props)
 })
 ```
+
+The second way is to provide `plugins` prop to the `<Canvas>` component. This is useful if you want to make a simple application that doesn't need to split up the code into multiple files.
+
+```svelte
+<Canvas plugins={[({ ref, props }) => {
+	console.log(ref, props)
+}]}>
+	<T.Mesh />
+</Canvas>
+```
+
+Plugins open up the component `<T>` to external code that will be injected via context into every child instance of a `<T>` component.
 
 If a plugin decides via `ref` or `props` analysis that it doesn't need to act in the context of a certain `<T>` component, it can return early.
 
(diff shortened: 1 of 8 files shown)


grischaerbe:
Hey @seo-rii and thanks for your PR! ❤️ 

First, we have to establish what the public facing interface of a [plugin](https://next.threlte.xyz/docs/learn/advanced/plugins) may and should look like.
Let's take the example [`interactivity` from `@threlte/extras`](https://next.threlte.xyz/docs/reference/extras/interactivity).
It's a function that takes a [set of predefined arguments](https://github.com/threlte/threlte/blob/72591456c02c5d25d90730b722b65d2b9d72d23f/packages/extras/src/lib/interactivity/index.ts#L9C5-L9C5). These arguments are used to invoke `injectPlugin` and set up a special state. This pattern enables the consumer of the plugin to **customize the plugin instance** and is the primary way to enable plugins in a Threlte app. All currently available plugins are set up like this.

```ts
const somePlugin = (options) => {
  // set up plugin based on passed options
  injectPlugin('name', plugin)
  // we *may* return something here, but it's not necessary
}

// consume via
somePlugin({ /* … */ })
```

That being said, I'm suggesting a different route here. You already paved the way with a component called `<Plugins>` which I think is clever and the right thing to do. We would release a component `<Plugins>` in `@threlte/extras` and it would accept a single prop called `plugins` which is of type `(() => void)[]`. This will enable you to use plugins in the top-level of your application:

```svelte
<script lang="ts">
  import { Canvas, T } from '@threlte/core';
  import { transitions, interactivity, Plugins } from '@threlte/extras';
  import { somePlugin } from './somePlugin';
</script>

<Canvas>
  <Plugins
    plugins={[
      somePlugin,
      transitions,
      () => interactivity({ target: someElement }) // example of usage with customization
    ]}
  >
    <T.Mesh
      on:click={() => {
        console.log('click');
      }}
    >
      <T.BoxGeometry />
      <T.MeshBasicMaterial />
    </T.Mesh>
  </Plugins>
</Canvas>
```

The component `<Plugin>` itself is very lean and straight forward:

```svelte
<script lang="ts">
  export let plugins: (() => void)[] = [];

  plugins.forEach((plugin) => plugin());
</script>

<slot />
```

What do you think about that? I'm also leaning towards this solution because it doesn't break our [recommended app structure](https://next.threlte.xyz/docs/learn/basics/context#recommended-app-structure) and the option to override plugins mid-graph.

seo-rii:
I think it's more reasonable. I'll change code.

grischaerbe:
@seo-rii Are you still on this? I think it's a reasonable thing to do but your implementation is still out of line with what I had in mind. Let me know if you have any questions!

PR #433, opened 2023-06-24
Use renderer.setAnimationLoop for main frame loop

michealparks:
### Overview

This PR replaces the driver of the current frame loop, `window.requestAnimationFrame()`, with the THREE.js `renderer.setAnimationLoop()` method. This is achieved by replacing the `useRaf()` call and deferring the initialization of the frame loop until after the default renderer has been created. `useRaf()` was removed since it is no longer used.

I've tested these changes locally with `npm link` under the following scenarios:
* With my recent [svelte hack project](https://github.com/michealparks/odyssey), which makes use of `useFrame()` in multiple components, as well as `useRender()` in order to have the pmndrs/postprocessing library render the scene.
* With the demos within the `routes` section of [threlte-xr](https://github.com/michealparks/threlte-xr), specifically the in-progress [`<Teleport>`](https://github.com/michealparks/threlte-xr/blob/main/src/lib/teleport/index.svelte) component requires `useFrame()` to work during a VR session.

If this approach doesn't make sense or should be tweaked I'll be happy to update this! 

### Advantages
* Allows the use of `useFrame()` and `useRender()` while within a VR or AR session 🎉  

### Caveats
* Any calls within userland to `setAnimationLoop()` will stop the main frame loop. This seems like an uncommon / not recommended usage of threlte, but thought it should be mentioned.
* When within an XR session $frameloop must be set to "always", or the headset will not render new frames as the user moves their head. I do this automatically when starting an XR session within threlte-xr but not sure if it should be called out more.


grischaerbe:
Thanks @michealparks for your contribution! ❤️ 

I have one concern/note:
With the addition of `useRender` for Threlte 6 we introduced custom render pipelines. A custom pipeline might use a custom renderer: 

```ts
const webGpuRenderer = new WebGPURenderer()

useRender(({ scene, camera }) => {
  webGpuRenderer.render(scene, camera.current)
})
```

I'm wondering if there are any side effects in this scenario and whether we need to implement a `setRenderer` method.
I assume the "backend" for `setAnimationLoop` is `requestAnimationFrame` or the XR equivalent, so we're good but I didn't check in detail.

michealparks:
Ah! Yep, I had forgotten that case. Does it make sense to allow users to optionally pass a renderer into the `<Canvas>` component? I'm thinking that might be an easy entry point resulting in few changes since `createRenderer()` is called there. That would also probably allow users to set WebGPU renderers without having to call `useRender()`?

michealparks:
> I assume the "backend" for setAnimationLoop is requestAnimationFrame or the XR equivalent, so we're good but I didn't check in detail.

Yep, it seems as though the backend is sending the callback arg to both `requestAnimationFrame()` and `xr.requestAnimationFrame()`.

I additionally just thought of checking `WebGPURenderer` to ensure that it also has `setAnimationLoop`. [It does](https://github.com/mrdoob/three.js/blob/master/examples/jsm/renderers/common/Renderer.js#L314), and it seems (from my quick reading of THREE's issues) that the team is moving toward adopting this class as a common renderer, with multiple backends, like WebGL, WebGPU, etc.

grischaerbe:
> [...] and it seems (from my quick reading of THREE's issues) that the team is moving toward adopting this class as a common renderer, with multiple backends, like WebGL, WebGPU, etc.

It only makes sense, I'm glad they're making the move. For now, I think we should not provide the option to pass a renderer in `<Canvas>` because the resulting types would affect the overall ecosystem. `useThrelte().renderer` would then be `WebGLRenderer | Renderer | …` or we would have to provide a generic `useThrelte` which I would like to prevent. We can think about that path as soon as they make the move to a unified backend so that `useThrelte().renderer` is just `THREE.Renderer`.

because rAf is essentially the "backend" for `setAnimationLoop` I don't see why we cannot just go for it as it is currently in your PR, what do you think?

michealparks:
> because rAf is essentially the "backend" for setAnimationLoop I don't see why we cannot just go for it as it is currently in your PR, what do you think?

Yep, I think so! This should be good to go as is.

grischaerbe:
@michealparks I took the liberty to reset the package.json file to its state before your PR. Don't worry about the versioning, we're running [changesets](https://github.com/changesets/changesets) which is doing that for us based on the changeset that you provided with your PR already, so all good! 👍 
Thanks again for your contributions and really looking forward to the XR package!

michealparks:
Yay! Glad to assist, and expect it soon!

Issue #427, opened 2023-06-21
Setting position of T.PerspectiveCamera causes unexpected behavior


seo-rii:
Setting T.PerspectiveCamera's position using an array causes strange errors such as camera movement during rendering. To maximize this, try inserting OrbitControl into the camera. This does not happen when each coordinate is specified separately, such as position.x.

You can identify this bug by creating a new local project by copying the code from https://threlte.xyz/getting-started#first-scene.

grischaerbe:
Hey @seo-rii,

I'm not sure I understand your issue. What do you mean with "_camera movement during rendering_"? Can you make a screen recording? You're also linking to the documentation of Threlte 5, but passing an array to the prop `position` is only supported from Threlte 6 on.

seo-rii:
Oh, never mind. This was a matter of example in the document. To be exact, `T.PerspectiveCamera` was not a problem, but `<OrbitControls maxPolarAngle={degToRad(80)} target={{y:0.5}}/>` was the problem. Target had to be specify x and z values too. I don't think there need any further action because @next removed this example.

seo-rii:
I'll reverse my words. This was a kind of issue. I would appreciate it if you could refer to #438.

PR #438, opened 2023-06-28
fix: use copy on props

seo-rii:
Summary: This PR improves compatibility by improving the logic of `useProps`.

In the @next version, threlte dynamically binds all the props. However, this creates significant compatibility issues. The problem is that the value of the three.js object cannot be applied to the element as it is. This was addressed in #427, and to sum up the problem, if you put a value such as `new Vector3(0, 0, 0)` in `position`, this `Vector3` is substituted in `position.x`, resulting in strange behavior.
This PR modifies the logic to use a `copy` function, not a `set`, where appropriate. This allows objects such as `Vector3`, `Color` to be substituted directly.

### DIFF
--- a/packages/core/src/lib/components/T/utils/useProps.ts
+++ b/packages/core/src/lib/components/T/utils/useProps.ts
@@ -74,11 +74,13 @@ export const useProps = () => {
     } else {
       if (typeof target[key]?.set === 'function') {
         // if the property has a "set" function, we can use it
-        if (Array.isArray(value)) {
+				if (target[key].constructor === value.constructor || (value.constructor === Object && !Object.keys(value).some(k => !(k in target[key])))) {
+          target[key].copy(value)
+        } else if (Array.isArray(value)) {
           target[key].set(...value)
         } else {
-          target[key].set(value)
-        }
+					target[key].set(value)
+				}
       } else {
         // otherwise, we just set the value
         target[key] = value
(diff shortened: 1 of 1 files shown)


grischaerbe:
Hey @seo-rii and thanks for this PR ❤️ 

I'm a little bit hesitant with your proposal because of several reasons:

1. It's tempting people to use a difficult pattern. Consider this example:
```svelte
<script>
  let rotationX = 0
  useFrame(() => {
    rotationX += 0.01
  }
</script>

<T.Mesh rotation={new Euler(rotationX, 0, 0)} />
```
In this example, a new `THREE.Euler` object is created every frame. This is an expensive operation compared to instantiating a new array that the function `set` can use: `<T.Mesh rotation={[rotationX, 0, 0]} />`. Of course the solution is to instantiate the object in the script block and reference it in the markup. But this needs education and would mean we would need to implement a section called "Performance Pitfalls" which I'd like to prevent. In special occasions this pattern necessitates forcing Svelte to update a prop with `rotation = rotation`, which again would need explaining.

2. It's a complex check that adds a significant runtime overhead. Especially `Object.keys(value).some…` has the potential to have a hit the performance.

3. It can be implemented via [plugins](https://next.threlte.xyz/docs/learn/advanced/plugins) easily. If this is something you personally need, feel free to hit me up on [Discord](http://chat.threlte.xyz), so we can discuss how to lay out such a plugin.

4. This addition also poses an educational problem as it's becoming increasingly harder to convey what's possible to pass as a prop to `<T>`. Also, the types are not reflected and would need to be adapted.

PS: My guess is that you wanted to check for the function `copy`, not `clone` on line 76, right?

I'm leaving this one open to discuss the matter 👍

grischaerbe:
We decided not to work on it but feel free to discuss the matter in any case if you'd like to.

PR #440, opened 2023-06-28
Terrain rapier example fixup

midanosi:
As mentioned in last PR, have updated this example, and written an explanation for it too.


grischaerbe:
Oh very nice! That example looks good! ❤️ 
LGTM 🚀

Issue #435, opened 2023-06-26
New Docs Navigation Layout


ixxie:
- [ ] Top navigation bar instead of accordion 
- [ ] In-page navigation moves to left sidebar
- [ ] Revise logo to use full Threlte branded logo

![image](https://github.com/threlte/threlte/assets/20320695/c47f9a27-e3a9-45dd-b9d0-8ab32d23f62d)

ixxie:
See also: https://github.com/threlte/threlte/issues/394

seo-rii:
I think it's better to split examples into menu, right to "learn".

grischaerbe:
> I think it's better to split examples into menu, right to "learn".

Good idea, implemented that in the PR #441 👍

ixxie:
Noting we are still keeping this issue open because the mobile menu has not been updated to reflect the new nav.

When addressing this, see also https://github.com/threlte/threlte/issues/545

PR #441, opened 2023-06-28
Documentation Layout

grischaerbe:
Refactoring the Documentation Layout, related to #435


grischaerbe:
It's a hot mess, but I'll merge it as it is. I'd like to make the individual packages separate collections as they also might have different requirements data-wise which would also result in a nicer left sidebar, but that's for another day. 👍

grischaerbe:
Thanks @ixxie for the input! ❤️

PR #442, opened 2023-06-30
Upgrade troika-three-text to 0.47.2

michealparks:
The latest THREE r154 release removes all references to `Buffer[name]Geometry`, which are imported in `troika-three-text@0.46.4`, causing `@threlte/extras` to throw uncaught reference errors when used with this release. This PR upgrades to `troika-three-text@0.47.2` since these imports were [were updated to `[name]Geometry`](https://github.com/protectwise/troika/compare/v0.46.4...v0.47.0) in `0.47.0`. The `<Text>` component was tested locally with `npm link` and appears to be working normally 👍

### DIFF
--- a/.changeset/fluffy-kids-shout.md
+++ b/.changeset/fluffy-kids-shout.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': patch
+---
+
+upgrade troika-three-text to 0.47.2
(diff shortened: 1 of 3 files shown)


grischaerbe:
Thanks, good catch! ❤️

PR #445, opened 2023-07-04
Performance: sort render and frame handlers when added instead of during each frame

michealparks:
### Overview

This PR contains a tiny optimization: it moves sorting of frame and render handlers out of the frame loop to only when these handlers are added to their respective `Set`. This change brings a few benefits:

* New arrays are not created on every frame with array methods that copy: `Array.from`, which may ease pressure a little on the garbage collector.
* Sorting is now "cached" and removed from a hot path. This probably doesn't matter much until many `useFrame` calls exist, but probably still worth doing.

As far as I can tell this doesn't change the behavior of the `order` param since it's only relevant to execution order of handlers, but I very well may have missed something important so calling that out!

### DIFF
--- a/.changeset/rude-guests-bathe.md
+++ b/.changeset/rude-guests-bathe.md
@@ -1,0 +1,5 @@
+---
+'@threlte/core': patch
+---
+
+Sort render and frame handlers on add instead of during each frame
--- a/.changeset/rude-guests-bathe.md
+++ b/.changeset/rude-guests-bathe.md
@@ -2,4 +2,4 @@
 '@threlte/core': patch
 ---
 
-Sort render and frame handlers on add instead of during each frame
+Sort render and frame handlers only when new ordered callbacks are introduced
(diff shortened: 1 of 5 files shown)


grischaerbe:
Hey @michealparks and thanks for your contribution, it's highly appreciated! ❤️ 

I have a proposal to further enhance performance in that regard: Instead of sorting the handlers on adding a handler, we implement sorting as the first step of the frameloop *if the respective set/array of handlers is flagged as "dirty"*.

Let's say we're instantiating 500 components, each with a `useFrame` handler. Here's how often we would sort:

- Current solution: Sorting 500 frame handlers each frame, obviously not good.
  - In the time frame of one frame: `O(1)`.
- Sorting on adding a handler: Sorting happens 500 times with increasing array length.
  - In the time frame of one frame `O(n)`.
- Sorting when the handlers array/set is marked as "dirty": Sorting happens *once*.
  - In the time frame of one frame: `O(1)`.
  
  This is not respecting the array length.

What do you think about that?

michealparks:
Yep! Agreed, I think that's a solid improvement, will make the appropriate tweaks.

michealparks:
Updated to option 3!

grischaerbe (approved):
That's excellent, thanks so much! ❤️ 

Also a great changeset name 😬

PR #452, opened 2023-07-11
added a tip to only use npx with @threlte/gltf CLI

voiys:


grischaerbe:
Thanks so much 🙌
LGTM 🚀

PR #454, opened 2023-07-21
Use three.js REVISION export for breaking api changes

michealparks:
This PR is a small quality-of-life tweak: it updates Threlte to use the three.js `REVISION` export to decide which properties to use if a breaking change has happened in a past revision. Properties are currently accessed if they aren't undefined, which can still cause deprecation warnings in the console:

Screenshot 2023-07-21 at 11 28 36 AM

This can be confusing or distracting to developers. By using `REVISION` these warnings are no more! 🎉

### DIFF
--- a/.changeset/friendly-tigers-wait.md
+++ b/.changeset/friendly-tigers-wait.md
@@ -1,0 +1,5 @@
+---
+'@threlte/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


grischaerbe:
Oh this is great! So much cleaner, thanks so much.
❤️

PR #455, opened 2023-07-22
Update Package READMEs

ixxie:
The NPM package READMEs are a little sad. This uses the new root README as a template and creates prettier package specific READMEs.

There isn't much package specific information added to most, but it serves as a starting point for future improvements.


grischaerbe:
Hey @ixxie, nice READMEs :)

Some minor details I noticed:
- I saw that the README for the package `extras` is talking about `core`.
- The section "Features" is only really valid for the package `core`, I'd remove it for every other package.
- The section "Quickstart" should be renamed to "Usage" for `create-threlte` and it should focus on the `npm create` command and not list the manual install method.
- The section "Quickstart" should link to the "getting started" documentation page (if available) of the respective package.

grischaerbe:
Thanks so much! 🚀

PR #456, opened 2023-07-23
Add LOD documentation for Threlte 6

nemzyx:
I was able to easily understand the three.js documentation on LODs, but i found it quite hard to implement LODs in Threlte without shooting myself in the foot. After diving deeper into how Threlte really works though, i was able to figure it out, but it was far from immediately obvious for someone new to the declarative way of thinking.

I looked in the documentation examples, but they all seemed very ambitious - which is great- but i was lacking some simpler examples to really make me understand how refs work in Threlte.

So - I created a really nice example for LODs, and the best practice to do them in Threlte in a declarative way.

This is basically the snippet, but the documentation i added goes into greater detail, and explaining what Threlte is really doing on the hood, and how it relates to the original three.js documentation on LODs.

```svelte
<T.LOD let:ref={lod}>
  {#each ["red","green","blue"] as color, i}
    <T.Mesh
      on:create={ ({ref}) => {
        lod.addLevel(ref, i * 75) // i * 75 = distance
      }}
    >
      <T.IcosahedronGeometry args={[10, 3-i]} />
      <T.MeshStandardMaterial {color} wireframe />
    </T.Mesh>
  {/each}
</T.LOD>
```

image

### DIFF
--- a/apps/docs-next/src/content/examples/geometry/LOD.mdx
+++ b/apps/docs-next/src/content/examples/geometry/LOD.mdx
@@ -1,0 +1,38 @@
+---
+category: Geometry
+title: LOD in Threlte
+showInSidebar: false
+---
+
+This is a practical example showing a best-practice method of implementing LOD in Threlte. The example is a great demonstration of the power of [ref](../../reference/core/t) bindings.
+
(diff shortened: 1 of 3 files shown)


grischaerbe:
Hey @nemzyx,

Beautiful example and great use of `ref` and the `on:create` event 👍 
Thank you so much, LGTM 🚀

PR #470, opened 2023-08-02
Canvas resize observer

michealparks:
This PR moves from using an `onresize` listener to a `ResizeObserver` in the `useParentSize` hook. A `MutationObserver` was added to the parent to account for reparenting. Let me know if there are any issues with this solution and I can update it 😎 

Note: there is still one case that I don't think this handles: if a user removes the canvas, waits some period of time, and then re-adds it to the document. I can't think of a clean browser API that handles this. Polling the canvas for a parent or adding a mutation observer to the entire document are solutions I can think of but seem potentially expensive.

### DIFF
--- a/.changeset/quiet-chairs-grow.md
+++ b/.changeset/quiet-chairs-grow.md
@@ -1,0 +1,5 @@
+---
+'@threlte/core': patch
+---
+
+Use a resize observer for handling parent size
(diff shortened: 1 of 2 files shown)


grischaerbe:
Thank you for your contribution! ❤️
Very clean implementation and listening to reparenting is a nice bonus 👍

grischaerbe (approved):

PR #480, opened 2023-08-04
Fix #478 Switch to lodash-es to solve vite build issues

### LINKED ISSUE #478: @threlte/theatre causes sveltekit build to fail
In a typescript sveltekit project `@threlte/theatre` causes errors on build using `@sveltejs/adapter-static`.

```
import { pick, omit } from "lodash";
               ^^^^
SyntaxError: Named export 'omit' not found. The requested module 'lodash' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'lodash';
const { pick, omit } = pkg;
```

Edit: also happening on @threlte/theatre@1.0.2

roelvanhintum:


grischaerbe (approved):
Thanks for your contribution! ❤️

grischaerbe (approved):

PR #482, opened 2023-08-05
Remove optional type for renderer

michealparks:
This PR changes the type of `renderer` returned by the `useThrelte` hook to be non-optional.

I noticed that a few components in extras were throwing errors with messages if `renderer` was undefined, so I instead moved this error into the `useThrelte` hook itself:

```ts
export const useThrelte = (): ThrelteContext => {
  const context = getContext<ThrelteContext>('threlte')

  if (context.renderer === undefined) {
    throw new Error('WebGLRenderer is undefined, is this component a child of <Canvas>?')
  }

  return context
}
```

This should make it so that users still get clear feedback if they accidentally use this hook outside of the `<Canvas>` component.

### DIFF
--- a/packages/core/src/lib/Canvas.svelte
+++ b/packages/core/src/lib/Canvas.svelte
@@ -110,7 +110,9 @@
 
   onDestroy(() => {
     contexts.internalCtx.dispose(true)
-    contexts.ctx.renderer?.setAnimationLoop(null)
+
+    // Renderer is marked as optional because it is never defined in SSR
+    contexts.ctx.renderer?.dispose()
   })
 </script>
 
--- a/packages/core/src/lib/Canvas.svelte
+++ b/packages/core/src/lib/Canvas.svelte
@@ -110,7 +110,7 @@
 
   onDestroy(() => {
     contexts.internalCtx.dispose(true)
-    contexts.ctx.renderer.setAnimationLoop(null)
+    contexts.ctx.renderer?.setAnimationLoop(null)
   })
 </script>
 
--- a/packages/core/src/lib/Canvas.svelte
+++ b/packages/core/src/lib/Canvas.svelte
@@ -61,7 +61,7 @@
    */
   export let useLegacyLights: boolean = true
 
-  let canvas: HTMLCanvasElement | undefined
+  let canvas: HTMLCanvasElement
   let initialized = false
 
   // user size as a store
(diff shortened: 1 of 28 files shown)


packages/core/src/routes/+page.svelte
```diff
@@ -1,3 +1,25 @@
-<!-- Chirping -->
+<script lang='ts'>
```
michealparks, line 1: I added a simple Threlte "Hello World!" here since it made quick debugging easier. It shouldn't affect published library output but if it isn't desired I can remove!

packages/core/src/lib/hooks/useThrelte.ts
```diff
@@ -2,5 +2,11 @@ import { getContext } from 'svelte'
 import type { ThrelteContext } from '../lib/contexts'
 
 export const useThrelte = (): ThrelteContext => {
-  return getContext<ThrelteContext>('threlte')
+  const context = getContext<ThrelteContext>('threlte')
+
+  if (context.renderer === undefined) {
```
grischaerbe, line 7: Actually we can test for the `context` instead of the `context.renderer`, looks cleaner.

grischaerbe (changes requested):
This is so good! Finally no `if (!renderer) …` :)
I made one tiny remark, apart from that perfect!

packages/core/src/lib/hooks/useThrelte.ts
grischaerbe: With an error message like this: "No Threlte context found, are you using this hook inside of <Canvas>?"

michealparks: Ah yes! Totally forgot that Svelte contexts will become undefined outside of their boundaries, will update

packages/core/src/lib/Canvas.svelte
```diff
@@ -103,15 +103,16 @@
...
     startFrameloop(contexts.ctx, contexts.internalCtx)
     initialized = true
   })
 
   onDestroy(() => {
     contexts.internalCtx.dispose(true)
-    contexts.ctx.renderer?.setAnimationLoop(null)
+
+    // Renderer is marked as optional because it is never defined in SSR
+    contexts.ctx.renderer?.dispose()
```
michealparks, line 115: This is still one case where an optional check needs to occur because `onMount` is never called during SSR but `onDestroy` is. Also, `dispose` stops the animation loop internally.

grischaerbe: Makes sense!

grischaerbe (approved):
LGTM 🚀

PR #485, opened 2023-08-05
Proposal: use peer dependencies for Svelte and Three

michealparks:
This PR contains a working proposal for moving `svelte` and `three` from `dependencies` to `peerDependencies` in all packages, with range limits set, which is how some other libraries like postprocessing ([package.json](https://github.com/pmndrs/postprocessing/blob/main/package.json#L91)) and r3f ([package.json](https://github.com/pmndrs/react-three-fiber/blob/master/packages/fiber/package.json#L58)) approach dependency management. There are a few big advantages for users with this change:

* If a user installs a newer version of three or svelte than @threlte's version, resolution to only one dependency will still occur. This proposal was inspired by getting past this problem, as I've noticed that I and some other threlte discord users have `npm install`ed three@0.155.0, resulting in multiple three versions loading in our projects.
* If a user installs a version of three that is lower (or higher if the approach that [postprocessing uses](https://github.com/pmndrs/postprocessing/blob/main/package.json#L91) is taken) then npm / pnpm will warn the user.

### DIFF
--- a/.changeset/cool-cameras-repeat.md
+++ b/.changeset/cool-cameras-repeat.md
@@ -1,0 +1,8 @@
+---
+'@threlte/theatre': patch
+'@threlte/extras': patch
+'@threlte/rapier': patch
+'@threlte/core': patch
+---
+
+Use peer dependencies for svelte and three
--- a/apps/docs/src/components/ManualInstallGuide/ManualInstallGuide.svelte
+++ b/apps/docs/src/components/ManualInstallGuide/ManualInstallGuide.svelte
@@ -24,7 +24,7 @@
         installTypes || installTheatre ? divider : ''
       }`,
     installTheatre &&
-      `${space}@threlte/theatre @theatre/core @theatre/studio${installTypes ? divider : ''}`,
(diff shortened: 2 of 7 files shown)


grischaerbe:
I'm all for this proposal. There's no particular reason for the current setup and there are good reasons to follow existing packages in that regard. The deployment error is because of the lockfile being out of sync. Can you `pnpm i` and push the updated lockfile?

grischaerbe:
Would you consider this a breaking change? I think *technically* it is but I'd file this as bugfixing?

michealparks:
Yeah I agree, it potentially breaks or fixes things based on how people have set up their projects. Hopefully most will already have three and svelte installed as dependencies (the [install guide](https://threlte.xyz/docs/learn/getting-started/installation) is pretty good about adding three and other required deps, but I do think this will possibly break some setups.

grischaerbe:
Hm, if I understand it correctly, the proposal rather opens up the version ranges that can be used alongside and also fixes the bug of three.js being loaded twice (because of different module versions resolved separately) so I wouldn't consider it a breaking change. We'll go for it!

PR #490, opened 2023-08-10
Dynamically import svelte component instead of theatre studio

michealparks:
The latest `@threlte/theatre` can't seem to resolve the dynamic import for `@theatre/studio`. I think that this may be a consequence of removing direct dependencies in [this PR](https://github.com/threlte/threlte/pull/485). When packaging for npm, studio is no longer bundled into the published files, and the dependency installed in the user project will instead be used. I'm not exactly sure of the root cause here - I've found that dynamic imports with Vite can sometimes be tricky, but I was able to fix this issue (tested with `npm link` on another threlte project) by instead dynamically importing the `InnerStudio` component via a relative path.

Not sure if this is the best solution, but it seems to fix the issue, and allowed a little simplification of InnerStudio in the process.


apps/docs/src/content/reference/theatre/getting-started.mdx
```diff
@@ -68,9 +68,9 @@ In your Scene, add the component `<SheetObject>` as a parent of any component yo
 
 ```svelte title=Scene.svelte
 <script lang="ts">
-  import { OrbitControls, T } from '@threlte/core'
```
michealparks, line 71: Fixed an import and removed an unused one in these docs.

packages/theatre/package.json
```diff
@@ -4,6 +4,7 @@
   "author": "Grischa Erbe <hello@legrisch.com> (https://legrisch.com)",
   "license": "MIT",
   "scripts": {
+    "dev": "vite dev",
```
michealparks, line 7: Made another little hello world for quick tests.

packages/theatre/src/lib/theatre/Theatre.svelte
```diff
@@ -11,10 +11,7 @@
   export let config: IProjectConfig | undefined = undefined
 </script>
 
-<Studio
-  enabled={studio.enabled}
-  hide={studio.hide}
->
+<Studio {...studio}>
```
michealparks, line 14: This fixes another bug if the component looks like this:
```svelte
<Theatre studio={{ hide }}>
```

Enabled was passed explicitly, and therefore svelte was setting it as `undefined` rather than the default `true`:

```svelte
<script>
export let enabled = true // is set as undefined
...
</script>

<!-- will not mount because enabled is undefined -->
{#if browser && enabled}
  <InnerStudio {hide}>
    <slot />
  </InnerStudio>
{:else}

grischaerbe (approved):

grischaerbe:
Thanks @michealparks, good idea to dynamically load the component instead of the dependency 👍

PR #495, opened 2023-08-13
Fix @threlte/core useRender postprocessing example

Joshimello:
Not sure if the example in apps/docs/src/examples/core/use-render is being used. 😅


grischaerbe:
Thanks a lot! ❤️

PR #511, opened 2023-08-17
Set useLegacyLights to false if greater than or equal to r155

michealparks:
This PR changes the `useLegacyLights` prop default to false if the user is on r155 onward.

Also ran `pnpm run format` on `@threlte/core` 🧹🧽🧼✨


grischaerbe (approved):
👍 🚧 🧹

PR #524, opened 2023-08-18
Resolving the dependencies whenever we create a new Threlte app

### LINKED ISSUE #523: Add correct semver ranges to dependencies when "npm create threlte"
When creating a project with `npm create threlte`, the script should use the latest versions of Threlte, Theatre and Rapier, but not with the tag `"latest"`, but with something like `^6`

grischaerbe:
Fixes #523 

@michealparks 

With `npm create threlte` we're dealing with a lot of different package versions:
- installed version of `create-svelte`
- version under our control: `@threlte/core,` `@threlte/extras`, `@threlte/theatre`, `@threlte/rapier`
- versions under external control: `@theatre/core`, `@theatre/studio`, `@dimforge/rapier3d-compat`, `three`, `@types/three`

The idea is that we at least automate updating some of them in the sense that we just fetch the current version from the latest published package via one of the package CDNs. For our packages this is safe, for external packages I limited the range to `^{current_version}` whereas `create-svelte` currently must be updated manually. I'd like to automate that as well as a lot of dependencies (svelte, vite, typescripte, …) come from that package, do you have an idea how?

Does that make sense to you in the first place?

### DIFF
--- a/packages/create-threlte/package.json
+++ b/packages/create-threlte/package.json
@@ -25,11 +25,12 @@
   "dependencies": {
     "@clack/prompts": "^0.6.3",
     "@expo/spawn-async": "^1.7.2",
-    "create-svelte": "^5.0.2",
+    "create-svelte": "^5.0.5",
     "execa": "^7.1.1",
     "fs-extra": "^9.0.1",
     "json-merger": "^1.1.10",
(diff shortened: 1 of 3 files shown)


grischaerbe:
Duplicate of #652

Issue #526, opened 2023-08-19
404ing link in docs example "Terrain with Rapier physics"


mknadler:
Under `How does it work`, the link 404's: 
> 1. Similar to the [3D noise example](https://threlte.xyz/terrain), loop over the vertices of a PlaneGeometry, and use a noise map to create a heightfield array heights."

From poking at this locally it looks like the url needs a `.` prefixing the `/terrain`.

grischaerbe:
Fixed in #527

PR #527, opened 2023-08-19
Fix docs link in terrain physics example

mknadler:
Described in https://github.com/threlte/threlte/issues/526. Went with the full path instead of `.` since it seems that's what you're doing for other links in the docs


ixxie (approved):
Looks good to me!

PR #534, opened 2023-08-21
Docs: Landing page odd behaviour fix 

jerzakm:
https://github.com/threlte/threlte/assets/16734228/e1019aa3-1062-4f1c-abd0-c13e338f3813

I ran into this issue some time ago and I decided to take a stab at it. I can only reproduce it on 1 of my 3 mobile devices - 2019ish Honor 10X.

I'm not sure about specifics but it has to do with the implementation of implementation of the Static noise pass and/or Color Dodge blend mode in postprocessing.

Disabling the static noise pass or changing `blendFunction: BlendFunction.COLOR_DODGE` to another one removes this issue as well.

A workaround I found, was to set `premultilpiedAlpha` in the renderer options. As far as I can tell it fixes the issue and doesn't change the visuals.

### DIFF
--- a/apps/docs/src/components/Hero/HeroWrapper.svelte
+++ b/apps/docs/src/components/Hero/HeroWrapper.svelte
@@ -68,7 +68,8 @@
           powerPreference: 'high-performance',
           antialias: false,
           stencil: false,
-          depth: false
+          depth: false,
+          premultipliedAlpha: false
         }}
       >
         <App />


grischaerbe:
Interesting, how is that the fix then? I'm failing to understand. Thanks for taking a successful stab! 🚀

jerzakm:
> Interesting, how is that the fix then? I'm failing to understand. Thanks for taking a successful stab! 🚀

Specifics are a little beyond my knowledge. I found it by commenting out random parts of Postprocessing and then found NoisePass to be the issue. I played with glsls there and then stumbled onto an unaswered stackoverflow question from years ago that mentioned something similar. Not worth going down the rabbithole for it.

The noisePass has premultiply parameter that sets a #DEFINES in the sahder. My guess is that something doesn't get applied correctly there on some devices. I checked and my old phone was using WebGL2 but maybe the hardware/firmware doesn't have some specific capabilities.

grischaerbe:
Good find 👍

Issue #536, opened 2023-08-22
Clean up doc links


DefinitelyMaybe:
Easy contribution for someone to who's wanting to get involved.

Occasionally we'll bump into links like the one in this section:
https://threlte.xyz/docs/reference/extras/transitions#usage
where there's a reference to localhost 😅 

If someone could do a decent comb over the docs and sort those out that'd be super 🎉

eerkmen:
On it!

eerkmen:
On it! I am going to create a fork and then check and fix the documentation in apps/docs/src/content/reference/extras/ directory. Let me know if there is another directory that needs a similar fix!

DefinitelyMaybe:
Well that was pretty good! thank you @eerkmen

PR #538, opened 2023-08-22
remove  localhost reference in transition.mdx

eerkmen:
Clean up doc links #536 issue fixed in transition.mdx file

### DIFF
--- a/apps/docs/src/content/reference/extras/transitions.mdx
+++ b/apps/docs/src/content/reference/extras/transitions.mdx
@@ -18,7 +18,7 @@ The plugin `transitions` enables [Svelte-like transitions](https://svelte.dev/do
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


DefinitelyMaybe:
nice! thank you @eerkmen ! Have a look around the other pages too, if you're using vscode [find and replace](https://code.visualstudio.com/docs/editor/codebasics#_advanced-search-options) can help track them down easily

eerkmen:
Oh sorry for not being clearer @DefinitelyMaybe , I already checked other instances of localhost with advanced search option in VSC. There is only one reference of localhost in the project folder(which was the one you pointed out on the issue page)

grischaerbe:
Good catch! 👍 
Thanks a lot, high appreciated ❤️

DefinitelyMaybe:
Oh man! I'm happily surprised 😅 thank you for clearing that up! I guess I was simply the unlucky person for finding it haha

PR #541, opened 2023-08-24
Docs Example: Interactive Shader

jerzakm:
(wip) Shader material example - Noise terrain with lines based on terrain height.
Interaction on click.

I've put shaders in separate files and ran into issues so:
- i bumped shiki to support .glsl
- added an icon for .glsl in file explorer - I couldn't find a good one though, any suggestions?

https://github.com/threlte/threlte/assets/16734228/40036a4a-ab73-43b6-9465-387a88f19c9e


grischaerbe:
Hey @jerzakm 👋 
The shader looks awesome, great work! Can we add a bit of text above the example that calls to click on the mesh? I guess it might be confusing otherwise. The other option (more work) would be to add a slight glow when hovering added to the click pulse. Do you want to add a line or two about "How it works"? I guess it would be mostly the Threlte implementation, since explaining glsl is out of scope for Threlte.

jerzakm:
> Hey @jerzakm 👋 The shader looks awesome, great work! Can we add a bit of text above the example that calls to click on the mesh? I guess it might be confusing otherwise. The other option (more work) would be to add a slight glow when hovering added to the click pulse. Do you want to add a line or two about "How it works"? I guess it would be mostly the Threlte implementation, since explaining glsl is out of scope for Threlte.

Thanks, I've had it ready for a while now but I've been putting off cleaning up text. Should be ready to merge now

grischaerbe:
Well written and on point, thanks ❤️

PR #543, opened 2023-08-25
fix broken doc links and 404s: core/utilities, core/use-render, core/…

JeffRice:
…use-frame, extras/instance, extras/instanced-mesh

Hi,

First time using changesets and contributing, hopefully it's all good!


grischaerbe (approved):
Absolutely all good! Thanks for checking the links and making a PR ❤️

Issue #554, opened 2023-08-26
Fix mobile layout to display Search on top.


jerzakm:
On mobile,  the Search dialog is displayed below the menu, probably need to have a look at z-index values.

https://github.com/threlte/threlte/assets/16734228/51c05424-a35f-407d-adac-a7d0df196672

Issue #545, opened 2023-08-25
Buggy behavior in the mobile menu (FF)


ixxie:
Two issue pop up when using the mobile menu of the docs site in Firefox on Android:
  1. The hamburger button seems intermittently unresponsive to taps
  2. Occasionally closing the menu results in a null page

See the following screen recording:

https://github.com/threlte/threlte/assets/20320695/ae45c8ef-7606-48a1-bb49-6e13bc93d042

grischaerbe:
The menu burger button seems to be broken after navigation, which is probably a side effect of the experimental Astro page transitions. @jerzakm if you have time on your hand, can you check that? I can also look into it but probably not before end of next week.

grischaerbe:
Since #556 is merged, can this be closed now?

PR #556, opened 2023-08-27
Mobile nav bug fixes

jerzakm:
**Buggy behavior in the mobile menu (FF) https://github.com/threlte/threlte/issues/545**
I changed default behaviour of ViewTransitions to fallback to "none" when they are not supported. There are two other options but I'm more confident in only having any ViewTransitions at all if they are fully supported (chrome) or none at all.

I tested for a little bit and it seemed to have worked on my Firefox. Can you let me know if the issue is also gone for you @ixxie ?

**Fix mobile layout to display Search on top. https://github.com/threlte/threlte/issues/554**
Search and menu were both using `z-50` as we initially assumed that menu will be always on top.

### DIFF
--- a/apps/docs/src/components/Menu/MobileNav.svelte
+++ b/apps/docs/src/components/Menu/MobileNav.svelte
@@ -13,7 +13,7 @@
   const { action, mounted } = useElementMounted()
 </script>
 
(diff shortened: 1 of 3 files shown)


ixxie:
Yes, that seems to have fixed the issues in FF mobile! Well done!

PR #562, opened 2023-08-29
Update Volume Description in <Audio/> Component to Specify Range

### LINKED ISSUE #553: <Audio/> explain volume range (0 - 1)
Discord: "please add to <Audio />  component into "volume" description that value is from 0 to 1. I put 50 and almost broke my ears 😄"

eerkmen:
This pull request resolves issue #553 by updating the description field for the `volume` attribute in the `<Audio/>` component. 

The new description clearly mentions that the volume range should be between 0 and 1. This clarification aims to prevent any confusion for users setting the volume and to avoid issues with unintended loudness.

Closes #553


grischaerbe:
Thanks @eerkmen ❤️

grischaerbe (approved):

PR #573, opened 2023-08-30
Create filter for running dev so 'create-threlte' package doesn't run

DefinitelyMaybe:
I've seen this pop up a couple of times for folks i.e. from root running `pnpm i` then `pnpm run dev` only for it to pop up with errors from create-threlte.

[relevant turbo docs](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#excluding-workspaces)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build": "turbo run build",
     "build:docs": "turbo run build --filter='@threlte/docs'",
     "preview": "turbo run preview",
-    "dev": "turbo run dev --parallel",
+    "dev": "turbo run dev --parallel --filter=!create-threlte",
     "lint": "turbo run lint",
     "install:all": "pnpm install --filter='!@threlte/gltf'",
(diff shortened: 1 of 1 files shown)


grischaerbe:
Hey @DefinitelyMaybe, thanks for your contribution! I'm sorry I have to close it though. Running dev in the root would still run mutliple repos which is probably never what you want:

- in core, extras and theatre (and rapier + xr in the future) `dev` would start the dev-related playground which you would only start when developing on a specific package
- in create-threlte `dev` is to test out the CLI creating a package (as this PR acknowledges)

We have a setup which links the actual source files to workspace repos (such as the docs) so we don't need to rely on intermediate tooling (transpiling, bundling, …) and running that is then not needed. Therefore, I'd just remove the `dev` script entirely from the root package.json, it's been there for a long time and it's an oversight. If you want to create a PR for that, I would greatly appreciate!

Sorry! 🙏

PR #574, opened 2023-08-30
Update Stackblitz template and reorganize useTweakPane imports

### LINKED ISSUE #568: Improve or replace Tweakpane for readability and to fix StackBlitz examples
Some of our examples use tweakpane to showcase what changing parameters does but it has two main drawbacks:
- StackBlitz examples are broken because we import useTweakpane hook with internal relative path
- it produces very verbose code (over 100 lines in Grid example)

Self notes, I'm going to look into:
- replacing tweakpane? Are there any similar libs in svelte?
- refactoring the hook for cleaner code
- if we don't repalace it, how to include it in stackblitz (make use of alias like $lib/utils instead of relative ../ paths)

I assigned myself, but any input is appreciated!

jerzakm:
**Reorganized Stackblitz template to make it easier to work with:**
Everything was in one file with raw strings and it was difficult to read. Instead I made a special /stackblitz-files/ directory and import the contents as strings with `?raw` vite import

**useTweakPane moved**
Tweakapane is now imported from $lib/useTweakPane instead of relative path. I added the file to stackblitz template and updated TS/aliases support so it should all work now.

Caveat: All of the examples technically seem to work but I noticed that we often use tailwind classes to size canvas so since it's not included in Stackblitz canvas is often 0x0. I'll have to go over examples and fix them up anyway, there's a lot of old stuff that's broken in various ways, imho out of scope for this PR

closes #568


jerzakm:
I'm aware of the whitespace. I'll fix and squash tomorrow probably. I need to tinker with my editor config, it was not picking up our  .editorconfig and auto formatted with spaces :(

grischaerbe (approved):
Looks good! As we're going tabs anyway, this is fine 👍

PR #579, opened 2023-08-31
Disable mouse acceleration

FlokiTV:
I'm on my laptop and mouse acceleration causes the camera to flick. To get a more accurate position and make the experience more pleasant on laptops, I tweaked the example.


FlokiTV:
> Thanks @FlokiTV for your contribution! Seems reasonable to do, so LGTM 🚀

it's a small fix, I'm getting familiar with the repo to contribute more!

Issue #582, opened 2023-08-31
Clean up the Documentation Layout


grischaerbe:
The styling and user expoerience in the docs is kind of a patchwork concerning the two sidebars and the footer on desktop. We should look into a better UX.

grischaerbe:
Current situation:

https://github.com/threlte/threlte/assets/46897060/f7c30a96-0e68-453d-8f29-5e879552aeb0

Issue #552, opened 2023-08-26
Improve consistency of text wrapping in sidebar nav


jerzakm:
Changing styling of currently active element causes some links with longer text to shift layout. 

https://github.com/threlte/threlte/assets/16734228/3a7553ae-2070-4e3e-8623-bdd0bb6f1fba

PR #583, opened 2023-08-31
Docs layout UX & cleanup pass

jerzakm:
# Search location
Search remains on the sidebar in the layout without sidebar. Notice that for smaller screens I'm also hiding icons for "Learn" and "Examples". Doing this allows for a few more packages up on top, and even more if we were to re-style it a bit and remove package icons.

image

# Sidebars + footer
Adressed UX issue of footbar + sticky sidebars - I swapped to fixed sidebars of varaible length, based on the visible size of footer.
I check footer size on scroll and then use a css variable to `calc()` sidebar height. Adds custom styled scrollbars
Closes https://github.com/threlte/threlte/issues/582

https://github.com/threlte/threlte/assets/16734228/431661af-05bd-4f7d-bfd7-9f3aed5ca32e

# Sidebar text wrapping
Not a direct fix, but upgrades to sidebars cause less annoying text wrapping. Here are two 'fixed' examples

Closes https://github.com/threlte/threlte/issues/552

https://github.com/threlte/threlte/assets/16734228/9bb394b1-ffa4-454a-ae0c-5bdf1d8a65eb

# Social icons

Github + Discord icons on larger screens. Each option is optional and in docs they are smaller than on main page (by 6px)
Added props to socials component:
```ts
interface Props {
  small?: boolean
  twitter?: boolean
  github?: boolean
  discord?: boolean
}
```

image


ixxie:
Generally awesome work @jerzakm!

Just a couple of nitpicks:

**Why did Twitter fly away?** It's there in the home page but not in the docs. Personally, I don't think we necessarily _need_ the social icons in the top bar, but _if_ we have them, we should be consistent about it.

**The gap on the right of the search bar is larger than the left** - see the image below

image

jerzakm:
> Generally awesome work @jerzakm!
> 
> Just a couple of nitpicks:
> 
> **Why did Twitter fly away?** It's there in the home page but not in the docs. Personally, I don't think we necessarily _need_ the social icons in the top bar, but _if_ we have them, we should be consistent about it.
> 
> **The gap on the right of the search bar is larger than the left** - see the image below
> 
> image

My thinking about social icons was that Discord and Github would be relevant to someone browsing documentation, twitter less so. 
I know we have the sidebar with these now but I would leave a github icon in top right corner. That's where I personally look for github link when I'm browsing other sites.

grischaerbe:
@jerzakm While you're at it, can you homogenize the bg color of the mobile menu, the topbar and the theme color (which iOS safari uses to colorize the status bar)?

image

grischaerbe:
Overall i do like the changes a lot, great attention to detail 👍. I must say I'm not particularly fond of the always visible scrollbar. On macOS it used to hide when you're not scrolling which is different on Windows I guess. Can we hide them completely or do you think that would be a great a11y concern?

jerzakm:
> Overall i do like the changes a lot, great attention to detail 👍. I must say I'm not particularly fond of the always visible scrollbar. On macOS it used to hide when you're not scrolling which is different on Windows I guess. Can we hide them completely or do you think that would be a great a11y concern?

Do you know how you'd like to solve it? Scrollbar visible only on hover and when the scroll event happens? I'm not sure. Feel free to jump in with any macOS specific tips

jerzakm:
> While you're at it, can you homogenize the bg color of the mobile menu, the topbar and the theme color (which iOS safari uses to colorize the status bar)?

Status bar color should be changed now

grischaerbe (approved):
LGTM 🚀

PR #584, opened 2023-09-01
Rapier events bugfix

grischaerbe:


grischaerbe:
Added proper event types to `<RigidBody>` components, updated from `SvelteComponentTyped` to `SvelteComponent`

PR #600, opened 2023-09-06
Add a title to every docs page

tombl:
Currently every page has a `<title>` of "Threlte", which makes referencing multiple pages at once difficult.

I've made it mandatory to specify a title for each page, and provided a title for each.


jerzakm:
Looks good to me, I don't know how we missed it 😄

grischaerbe (approved):
Thanks, good catch! ❤️

PR #610, opened 2023-09-06
Added slot `"fallback"` to `<XR>`

### LINKED ISSUE #597: [xr] Provide fallback slot in `<XR>`
XR apps are different to regular apps in that they do not need a manually set up camera. This means that if you launch an XR app, you're thrown in the potentially already visible scene but with the default camera, which looks dull:

Bildschirmfoto 2023-09-04 um 23 55 50

We should think about providing a fallback slot in the component `<XR>` that renders its contents when the XR session has *not* been entered yet as opposed to the contents of the default slot:

```svelte
<XR>
  <Hand left />
  <Controller right />

  <svelte:fragment slot="fallback">
    <T.PerspectiveCamera makeDefault position={[10, 10, 10]}>
      <OrbitControls />
    </T.PerspectiveCamera>
  </svelte:fragment>
</XR>
```

grischaerbe:
Fixes #597

### DIFF
--- a/.changeset/sixty-insects-listen.md
+++ b/.changeset/sixty-insects-listen.md
@@ -1,0 +1,5 @@
+---
+'@threlte/xr': patch
+---
+
+Added slot `"fallback"` to `<XR>`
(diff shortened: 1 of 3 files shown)


michealparks (approved):
🚀 🚀

PR #614, opened 2023-09-08
XR - Initialize event dispatchers for input devices only once

michealparks:
Fixes https://github.com/threlte/threlte/issues/602.


michealparks:
Closing because this solution works with the simulator but not in headset :| will revisit this weekend

PR #618, opened 2023-09-10
Move vite-built packages over to `moduleResolution: "bundler"`

### LINKED ISSUE #617: THREE example types fail for projects using `moduleResolution: "bundler"`
Hi, firstly thank every one of you for this wonderful project.

## Explanation

When using TypeScript with the config `moduleResolution: "bundler"`, some of the types will resolve to `any`.

image

This is due to a combination of the usage of `three/examples/jsm/*` with no extension at the end (i.e. `three/examples/jsm/loaders/GLTFLoader` instead of `three/examples/jsm/loaders/GLTFLoader.js`), & the difference of resolution between `"bundler"` (the current recommended standard) & `"node"` (the config used by this project).

Adding '.js' to the end of the `d.ts` manually,

image

 ...resolves the issue:

image

## Repro

1. Set `moduleResolution: "bundler"` in `tsconfig.json`
2. Create the following file:
```svelte
<script lang="ts">
  import { useGltf } from '@threlte/extras';

  const gltf = useGltf<{ nodes: {}, materials: {} }>('/amogus.glb');
     // ^ AsyncWritable<any>
</script>
```

## Solution

If it's okay with the maintainers, I propose the library to ship with `.js` at the end of these imports to support every one using `"bundler"` whilst maintaining `"node"` compatibility.

Sxxov:
Fixes #617, by:

1. Adding '.js' to all usages of `three/*` sub-dir imports
2. Making all vite-built packages (`core`, `extras`, `rapier`, `theatre`, & `xr`) use `moduleResolution: "bundler"` to enforce this behaviour moving forward

> [!NOTE]
> This also includes a small additional fix for building on Windows by changing some of the quotes to be consistent, but I thought it was benign enough to not need a separate PR. Any guidance on this change would be good.


jerzakm:
I had a look at the type resolution issue and from what I can tell, adding the following line to tsconfig.json resolves it.

```
"typeRoots": ["node_modules/@types"]"
```

It seems that Astro needed extra configuration to find Definitely Typed types. I'm not very knowledgeable in this area, but if this works then I think we should move forward with `moduleResolution: "bundler"` and this.

grischaerbe:
Hi @Sxxov and thanks for your PR. Fortunately, @jerzakm found out that the type problem (exclusively?) exists in our documentation and also found the right fix for it. We are moving to moduleResolution: "bundler" at some point, but adding extensions to all files will mess with our current setup in a way of which I can't comprehend the consequences and will also make it harder for future contributions. Nevertheless I appreciate this PR as without it we wouldn't have dealt with this issue and possibly would've not found a fix for it, thanks! ❤️

PR #619, opened 2023-09-12
More <Grid/>  options and updates

jerzakm:
1. **Added 3 new grid types:**
	- topology (lines based on specified axis)  
	- circular (concentric circles) 
	- polar (circular with line dividers for circle angles)

	Grid type is set by a "type" prop. Default is {type: "grid").	
	
2. **Background color** for the grid through `backgroundColor` and `backgroundOpacity`.
	Setting opacity to 0 disables background (default) - current behaviour.

3. Support for custom geometries by using the slot. Default is the current Plane. 
		
I moved the shader to a separate .ts file with /*glsl*/ comment for readability and code highlighting. We used to have to re-generate a ShaderMaterial every time axes changed, but it's handled via uniforms now.


packages/extras/src/lib/components/Grid/Grid.svelte.d.ts
```diff
@@ -10,11 +10,35 @@ export type GridProps = Props<Mesh> & {
...
   sectionSize?: number
   sectionThickness?: number
+  backgroundColor?: ColorRepresentation
+  backgroundOpacity?: number
   gridSize?: number | [number, number]
   followCamera?: boolean
   infiniteGrid?: boolean
   fadeDistance?: number
   fadeStrength?: number
+  type?: GridParams | TopologyParams | CircularParams | PolarParams
```
grischaerbe, line 20: Hm, this might look a bit odd:

```
<Grid type={{ type: 'lines' }} />
```

and I would prefer 

```
<Grid type="lines" />
```

and based on the prop `type` other props are necessary. Possible type implementation:


```ts
export type GridProps = Props<Mesh> & {
    axes?: 'xzy' | 'xyz' | 'zyx'
    // …
  } & (
      | {}
      | {
          type: 'grid'
        }
      | {
          type: 'lines'
          axis?: 'x' | 'y' | 'z'
        }
    )
```

grischaerbe: That being said, maybe I would rename `axes` to `plane`:

```
<Grid plane="xz" type="lines" axis="x" />
```

jerzakm: Updated the types. We decided to do the `axes` to `plane` rename at next major bump.

grischaerbe:
Thanks @jerzakm! Looks awesome!

PR #620, opened 2023-09-13
Fix MeshLineMaterial shader to calculate normals and offsets correctly

laszlokorte:
Currently the line thickness varries because the normals are not calculated correctly in screen space.

For example here you can see that the line thickness reduces if the line curves away from the camera:
grafik

This occurs even if `scaleDown` is not set. This PR fixes this by adjusting the the vertex shader.

The result looks like this:
grafik

`scaleDown` can still be set (orange line):
grafik

Custom `shape`, `dashArray` and `attenuation` options are still working:
grafik

### DIFF
--- a/packages/extras/src/lib/components/MeshLine/vertex.ts
+++ b/packages/extras/src/lib/components/MeshLine/vertex.ts
@@ -67,6 +67,7 @@ export const vertexShader = `
 
         if(sizeAttenuation != 0.0) {
             normal /= currentClip.w;
+						normal *= min(resolution.x, resolution.y) / 2.0;
         }
 
         if (scaleDown > 0.0) {
--- a/packages/extras/src/lib/components/MeshLine/vertex.ts
+++ b/packages/extras/src/lib/components/MeshLine/vertex.ts
@@ -75,7 +75,7 @@ export const vertexShader = `
             normal *= smoothstep(0.0, scaleDown, dist);
         }
 
-        vec2 offsetInScreen = actualWidth * normal * side;
+        vec2 offsetInScreen = actualWidth * normal * side * 0.5;
 
         vec2 withOffsetScreen = currentScreen + offsetInScreen;
         vec3 withOffsetNormed = vec3((2.0 * withOffsetScreen/resolution - 1.0), currentNormed.z);
--- a/packages/extras/src/lib/components/MeshLine/vertex.ts
+++ b/packages/extras/src/lib/components/MeshLine/vertex.ts
@@ -67,7 +67,6 @@ export const vertexShader = `
 
         if(sizeAttenuation != 0.0) {
             normal /= currentClip.w;
-            normal *= min(resolution.x, resolution.y)*0.05;
         }
 
         if (scaleDown > 0.0) {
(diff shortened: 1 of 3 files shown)


grischaerbe:
Hey @laszlokorte, thanks for your contribution and sorry for my late response 🙏 
The examples though look different now, see for instance the example "cursor line":

https://github.com/threlte/threlte/assets/46897060/3e56b64b-8801-498e-986a-007e26adb338

Other examples (line thickness, geometry when camera is pointing towards stroke):
Bildschirmfoto 2023-09-14 um 22 44 39

https://github.com/threlte/threlte/assets/46897060/f376a5fb-a7f1-4612-bea2-271932c68dad

This is not a blocker (although it's *kind of* a breaking change), I'm just curious, is that by design? I'm also going to ping @edsunman as they're the original author and wait for their opinion.

edsunman:
Perhaps the widths of the lines in the examples need adjusting? I’ll double check when I’m back at my desk tomorrow.

laszlokorte:
I agree that it would be preferable if the behavior would not change that much. 
In my opinion the previous implementation was just wrong in some fundamental ways (as indicated by the issues in the repo the code was originally based on: https://github.com/spite/THREE.MeshLine/pull/150, https://github.com/spite/THREE.MeshLine/issues/52, https://github.com/spite/THREE.MeshLine/issues/49, https://github.com/spite/THREE.MeshLine/issues/132, https://github.com/spite/THREE.MeshLine/issues/136).

It looks like most of the differences could be corrected by adjusting the line width. The main issues seems to be the scaling when attenuation is turned on.

packages/extras/src/lib/components/MeshLine/vertex.ts
```diff
@@ -20,62 +21,72 @@ export const vertexShader = `
...
-        normal.xy *= .5 * w;
-        normal *= projectionMatrix;
+        vec2 normal = vec2(-dir.y, dir.x);
 
-        if( scaleDown > 0. ) {
-            float dist = length(currentP - prevP);
-            normal *= smoothstep(0.0, scaleDown, dist);
+        if(sizeAttenuation != 0.0) {
+            normal /= currentClip.w;
+						normal *= min(resolution.x, resolution.y) / 2.0;
```
laszlokorte, line 70: This factor could be adjusted to make the result similar to the previous implementation

grischaerbe: If this has no implications on for instance the possible range of attenuation, I'm fine with adjusting it like this 👍

edsunman: Yes, I have just had a play around with the examples and it looks like removing the `/ 2.0` at the end of line 70 will match the width to the original implementation. Currently the lines are half the width they should be.

Apart from that it all looks good to me!

The lines in the CursorLines example look very large because I'm using an orthographic camera so I should have set the attenuation to false. I'm happy to submit a PR to fix that.

laszlokorte: I have remove the `/ 2.0` scaling now.

edsunman: Thanks for your work on this @laszlokorte ! Happy for you to merge @grischaerbe 👍

grischaerbe:
Thanks @laszlokorte and @edsunman for both your work, merging now 🚀

Issue #605, opened 2023-09-06
[xr] Move `useGamepad` to extras, add `useMappedGamepad` hook


grischaerbe:
`useMappedGamepad('standard')` or `useMappedGamepad('xr-standard')` as event based hooks.

Something like this:
```svelte
<script>
  const gamepad = useMappedGamepad('standard')
  gamepad.on('A', 'down', () => {
    console.log('Button A pressed!')
  })
</script>
```

PR #622, opened 2023-09-14
Add useGamepad hook

michealparks:
This PR sets up a `useGamepad` hook that provides both a raw and mapped gamepad. This was a first pass at an idea for the API, so feedback and ideas would be appreciated!

Docs page: https://threlte-git-fork-michealparks-usegamepad-threlte.vercel.app/docs/reference/extras/use-gamepad

Right now this only supports mapping for classic gamepads, if we end up liking this design then as a follow up I'll add the `standard-xr` mapping to this hook and remove the `@threlte/xr` useGamepad hook since this'll cover it much better :)

Resolves https://github.com/threlte/threlte/issues/605


michealparks:
> Can we make { mapping: 'none' } the default and rename mapped? If you see mapped` out of context, it's maybe a bit too generic.

Good point, and agreed. Any thoughts on a good name? Maybe: `mappedGamepad` or `standardGamepad`?

> What do the touch events react to?

These are really cool. Some controllers will emit data if a button is pressed slightly, or "touched". If I recall correctly the quest 2 controllers support this and you can observe just by touching and not pressing the `a` or `b` buttons. Others like the PS5 controller will report this for the triggers if they're only slightly pulled.

> Do you think it's a good idea to implement "change" events for the axes?

Yes! Great idea. I'll add this 👍 

> Do you think it makes sense to implement catch-all events?

I also think this makes sense. My intuition would be to keep both and to normalize the event object:

```ts
gamepad.on('press', (e) => {
  e.type // 'press'
  e.button // 'clusterBottom'
})

gamepad.leftTrigger.on('down', (e) => {
  e.type // 'press'
  e.button // 'leftTrigger'
})
```

Think this makes sense?

michealparks:
@grischaerbe I just finished changes based on feedback; I definitely like this version more 🎉 let me know what you think!

packages/extras/src/lib/hooks/useGamepad.ts
```diff
@@ -0,0 +1,270 @@
+import { onDestroy } from 'svelte'
+import { currentWritable, useFrame } from '@threlte/core'
+
+type UseGamepadOptions = {
+  /** An optional gamepad index, if multiple gamepads are used. */
+  index?: number
+  /** The mapping strategy for the gamepad. Default is 'none'. Choose 'standard' to enable mapping. */
+  mapping?: 'none' | 'standard' // 'standard-xr'
+  /** The threshold value of an axis before change events are fired. Default is 0.05. */
+  axisChangeThreshold?: number
```
michealparks, line 10: rename deadzone?

michealparks:
@grischaerbe updates completed! Unless you notice anything, I think this should be ready!

grischaerbe (approved):
Hey @michealparks,
I made some adjustments to the hook: 
- I removed the browser check. Since we use `useFrame` in here, we can only use the hook in `<Canvas>` anyway and this is already browser-only be default.
- I added the handling of gamepads that are already connected at the time `useGamepad` is called. This is useful for usage inside components that get mounted later in time.
- I added `off` events to all buttons and the main event handling to stop listening to events. `on` now also returns an unsubsciber immediately. Strictly speaking calling `off` is not necessary when unmounting the component since `useFrame` is automatically cleaning up itself, but it's a good practice anyway. I added docs about that as well.

PR #624, opened 2023-09-16
Cursor Lines width fix

edsunman:
Changes the widths of the lines in the Cursor Lines example and sets 'attenuate' correctly.

Also tiny change to the AlphaMap example on the <MeshLineMaterial> page to make the brush texture visible from both sides.


grischaerbe:
Thanks a lot!

PR #629, opened 2023-09-17
removed dev script

DefinitelyMaybe:
related #573

won't print out anything just yet. simple solution first.


grischaerbe:
Thanks @DefinitelyMaybe !

PR #633, opened 2023-09-19
Fix color management of Theatre.js color picker

donmccurdy:
Unless otherwise specified, a color picker usually has sRGB input and output. When assigning the color to a THREE.Color instance, this should be specified so that three.js can make any necessary conversions to the working color space.

To test this change, create a canvas with `toneMapping={LinearToneMapping}`, and a MeshBasicMaterial. Assign a color to the material that isn't fully saturated, like `#FF69B4`. Then use a color meter (like [macOS Color Meter](https://docs.google.com/document/d/12W-n3nTl6C8-SyeP_trvKBXQirMLmvsGGl3yUlQsHaQ/edit#heading=h.vks9e22yqulu)) to check that the value shown on the material matches exactly. If the color meter can display colors in other color spaces, ensure that it's set to sRGB.

If non-linear tone mapping is enabled, some differences in display are expected.

Screenshot 2023-09-19 at 11 03 20 AM

### DIFF
--- a/.changeset/pink-seals-serve.md
+++ b/.changeset/pink-seals-serve.md
@@ -1,0 +1,5 @@
+---
+"@threlte/theatre": patch
+---
+
+Fix color management of Theatre.js color picker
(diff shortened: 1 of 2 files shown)


grischaerbe:
Thanks @donmccurdy for again and always getting into the details when it comes to colors!

PR #636, opened 2023-09-20
ContactShadows: use WebGLRenderTarget#colorSpace when available

donmccurdy:
VSCode will warn that the second branch is "unreachable", but I kept that in. Wasn't sure how long you'd like to keep support for three.js <= r151.

Screenshot 2023-09-19 at 9 54 00 PM


grischaerbe:
Thanks @donmccurdy !

PR #638, opened 2023-09-21
[xr] - Refactor and fix events for controllers and hands

michealparks:
### Overview

This PR has these goals:
* Slightly refactor the `<Controller>` and `<Hands>` components.
* Fix an issue where multiple events would fire for a single component.
* Improve types for hand and controller events.
* Remove the `useHandEvent` and `useControllerEvent` hooks. 

The removed hooks were complicating code and in my opinion they provided little value since these event listeners can be just as easily added to these components. With the upcoming `useGamepad({ mapping: 'xr-standard' })` I think they'll be less valuable and more redundant, and I'd like to trim any excess or redundant APIs since that seems to be our philosophy :) I'm curious if anyone thinks otherwise!

### Examples

This PR also makes a second pass at the `<Controller>` and `<Hand>` examples to make them a bit more fun 😎 

Resolves https://github.com/threlte/threlte/issues/602


packages/xr/src/lib/components/Controller.svelte
```diff
@@ -66,33 +61,44 @@
...
   }
 
-  $: if (left && right) {
-    throw new Error('A <Controller> component can only specify one hand.')
-  }
-  $: if (!left && !right) {
-    throw new Error('A <Controller> component must specify a hand.')
-  }
-  $: handedness = left ? 'left' : 'right'
+  $: handedness = ($$props.left ? 'left' : $$props.right ? 'right' : $$props.hand) as 'left' | 'right'
```
michealparks, line 64: Use `$$restProps`

packages/xr/src/lib/components/Hand.svelte
```diff
@@ -18,62 +18,67 @@
...
-  $: handedness = left ? 'left' : 'right'
-
-  const handleConnected = (event: XRHandEvent<'connected', null>) => {
-    const inputSource = event.data.hand as globalThis.XRHand
-    const eventHandedness = event.data.handedness as 'left' | 'right'
-    if (eventHandedness !== handedness) return
-    stores[handedness].set({ ...eventMap.get(event.target), inputSource })
-    fire('connected', event, { input: 'hand' })
-  }
+  $: handedness = ($$props.left ? 'left' : $$props.right ? 'right' : $$props.hand) as 'left' | 'right'
```
michealparks, line 47:
```suggestion
  $: handedness = ($$restProps.left ? 'left' : $$restProps.right ? 'right' : $$restProps.hand) as 'left' | 'right'

File: apps/docs/src/examples/xr/controller/Controller.svelte [hunk 3]
```diff
@@ -0,0 +1,110 @@
...
+  let text = ''
+
+  const count = 100
+  const positions = new Float32Array(count * 3)
+  const velocities = new Float32Array(count * 3)
+  const vec3 = new THREE.Vector3()
+  const matrix = new THREE.Matrix4()
+  const color = new THREE.Color()
+
+  const controller = useController($$props.left ? 'left' : 'right')
```
michealparks, line 20:
```suggestion
  const controller = useController($$restProps.left ? 'left' : 'right')

PR #639, opened 2023-09-25
[xr] - Fix headset position when teleporting

michealparks:
This PR fixes a few small issues I noticed with `<Headset>` and `useTeleport`, especially when used together:

* I previously set `<Headset>` to sync with the base reference space that the user is when the immersive session starts. The problem with this is that if a user teleports the reference space changes, and the component will be in an accurate space once a teleport occurs. `<Headset>` now uses the most up-to-date reference space.
* There are use cases for syncing `<Headset>` to the current camera before an immersive session starts, so I added an opt-in `<Headset syncCamera>` prop. Curious if you would change anything about this!
* I changed teleport so that it eagerly gets the original reference space instead of waiting for a store value. This fixed an issue where teleporting wasn't working if done immediately after an immersive session starts.


grischaerbe:
Thanks @michealparks, I would suggest to make `syncCamera` just the default behaviour and remove the prop. It seems logical to me. A note in the docs is all it needs 👍 
The rest looks good to me 🚀

michealparks:
@grischaerbe That makes sense! I was initially on the fence but I think you're right, will update!

michealparks:
@grischaerbe Updated!

PR #640, opened 2023-09-25
Fix color prop type for materials used with T component

michealparks:
This PR fixes an issue where Typescript would complain about valid color values passed as a prop to materials declared with the `<T>` component.

Screenshot 2023-09-25 at 4 21 48 PM

### DIFF
--- a/packages/core/src/lib/components/T/types.ts
+++ b/packages/core/src/lib/components/T/types.ts
@@ -87,6 +87,16 @@ export type CameraProps<Type extends any> = MaybeInstance<Type> extends { isCame
   : Record<string, unknown>
 
 /**
+ * ### Material Props
+ */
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/components/T/types.ts
```diff
@@ -129,7 +139,8 @@ export type Props<Type extends any> = AnyProps &
   BaseProps<Type> &
   ClassProps<Type> &
   CameraProps<Type> &
-  InstanceProps<Type>
+  InstanceProps<Type> |
```
michealparks, line 142: I'm not entirely sure if this is the correct solution... `|`ing seems logical thought since it will overwrite `color` rather than `&` merging it with the broken type definition that doesn't include `string`.

michealparks:
Opened https://github.com/three-types/three-ts-types/pull/614 to attempt to fix there.

PR #641, opened 2023-09-28
@threlte/flex

grischaerbe:
# @threlte/flex

This package will introduce a wrapper for [`yoga-layout`](https://yogalayout.com/).

## Features

- `<Flex>` and `<Box>` components
- Automatically centered anchor if a child is not a `<Box>` component (DOM anchors are upper-left aligned whereas Three.js object anchors are centered)
- Full type support on props
- prop setter architecture makes it easy to type props
- A class parser can be introduced to use a class-based syntax as opposed to individual props, a parser for tailwind-like classes is included
- Slot props and bindings to force a layout reflow
- The reflow is batched and will never happen more than once per frame

## Examples

The package contains a SvelteKit app with examples located at the route `'/'`, run `pnpm dev` at `packages/flex` to get started.

https://github.com/threlte/threlte/assets/46897060/5f381f09-b5aa-4c30-8c9b-f430647a5a57

## Credits

Some calculations are taken from [react-three-flex](https://github.com/pmndrs/react-three-flex), thanks.


grischaerbe:
Because changesets is not allowing us to do an alpha release only for this package, we'll release this into the wild without a documentation. As long as 0.1.0 is not reached this package is considered alpha and likely to break anytime. Use at your own risk.

PR #652, opened 2023-09-29
create-threlte fixes

grischaerbe:
- Added `@threlte/flex` to `create-threlte`
- Added dependency resolution based on package CDN
- Added a custom tsconfig.json that enables `"moduleResolution": "Bundler"`
- Updated `create-svelte`


jerzakm (approved):
New package is added correctly and our type resolution bug for Flex seems to be gone. I think we're good to merge this

PR #654, opened 2023-09-29
useFrame: stages and fixed step

michealparks:
~~This is a working proof-of-concept to gather feedback and experiment with adding a first version of `useFrame` stages and fixed timesteps in a backwards-compatible way~~ 🚀 

Here I've introduced two new properties to `useFrame`: `stage` and `fixedStep`.

With this solution current `useFrame` calls should not be affected and run the same as before.

A lot of inspiration was taken from [Unity's order of execution](https://docs.unity3d.com/Manual/ExecutionOrder.html) and [R3Fs recent update lifecycle changes](https://github.com/pmndrs/react-three-fiber/pull/2279).

### Stages

`useFrame` callbacks can now execute at various stages:

* `'fixed'` - Fixed handlers will executed first during a frameloop run, and will only execute if the amount of time required for them between executions has passed. This should be really valuable for steps that we want to operate more infrequently, like game AI calculations.
* `'before'` - These handlers will fire before render. This is the default for `useFrame` and how it has previously behaved.
* `'after'` - These handlers will fire after `renderer.render()`.

```svelte
useFrame(() => {
  // Runs before render, after render, or at a fixed rate
}, {
  stage?: 'fixed' | 'before' | 'after', // Default is 'before'
  fixedStep?: 1 / 60 // Default is a 60FPS target, will only have an effect if stage is 'fixed'
})
```

Order within these stages can still be specified.

```svelte
useFrame(() => {}, {
  stage: 'fixed',
  fixedStep: 1 / 30,
  order: Infinity,
})
```

### Other changes 
* ~~I replaced usage of clock internally with a [Timer](https://github.com/mrdoob/three.js/pull/17912) implementation due to it being more reliable. Timer is also sensitive to page visibility changes and won't produce large deltas when users navigate back to a tab.~~ removed from this PR.
* I altered `internalCtx.dispose()` to avoid calling `tick()` when not necessary since this is called on every frame and `await tick()` seems fairly expensive [due to triggering a `flush`.](https://github.com/sveltejs/svelte/blob/master/packages/svelte/src/runtime/internal/scheduler.js#L40) 

### Open questions
* Should the stages be called `beforeRender` and `afterRender`? `useFrame(() => {}, { stage: 'afterRender' })`
* Should we eventually migrate `useFrame` to `useUpdate` in order to move closer to R3F and Unity?
* Should we eventually replace `Clock` with `Timer` in user facing APIs?


packages/core/src/lib/lib/startFrameloop.ts
```diff
@@ -95,31 +60,79 @@ const shouldRender = (ctx: ThrelteContext, internalCtx: ThrelteInternalContext)
...
-    if (!shouldRender(ctx, internalCtx)) return
+    // run all fixed useFrame callbacks
+    if (fixed.size > 0) {
+      if (internalCtx.handlersNeedSort.fixed) {
+        sortHandlers(fixed)
+        internalCtx.handlersNeedSort.fixed = false
+      }
+
+      // From: https://gafferongames.com/post/fix_your_timestep/
+      fixed.forEach((h) => {
```
michealparks, line 81: Fixed frame implementation

PR #656, opened 2023-09-30
[xr] - Fix hands, hand joints, and useHitTest docs examples

michealparks:
This PR primarily focuses on fixing docs examples, and adds a few AR enhancements quietly:

* The Hands example had a bug caused by on variable overriding another in the `<Hands>` component.
* The `useHandJoint` example was not working because RigidBodies were being instantiated before each joint radius was defined.
* The `useHitTest` example had an array that was missing an assignment to make it reactive.
* Removed `useHandEvent` and `useControllerEvent` docs.

I also added the following improvements:

* I added 'plane-detection' and 'light-estimation' as required and optional features when starting an AR session. This'll become more important as headsets role out more mixed reality features.


grischaerbe:
Nice, thanks @michealparks!

PR #657, opened 2023-09-30
Make canvas properties reactive and document non-reactive properties

michealparks:
This PR:
* Makes all properties of `<Canvas>` reactive that can be reactive
* Adds JSDoc comments for non-reactive properties
* Makes watchers in `useRenderer` more granular so that only updated properties will update

I also alphabetized these props to make them easier to trace when viewing side by side 😄


grischaerbe:
Awesome, thanks @michealparks !

PR #660, opened 2023-10-01
[xr] - Add useHeadset hook

michealparks:
This PR adds a `useHeadset` hook that exposes the pose of the user headset. It also ensure that syncing the viewer pose the the headset object only occurs once per frame no matter how many `<Headset>` components or `useHeadset` hooks exist.

### DIFF
--- a/.changeset/lucky-ears-applaud.md
+++ b/.changeset/lucky-ears-applaud.md
@@ -1,0 +1,5 @@
+---
+'@threlte/xr': patch
+---
+
+Add useHeadset hook
--- a/apps/docs/src/content/reference/xr/headset.mdx
+++ b/apps/docs/src/content/reference/xr/headset.mdx
@@ -22,7 +22,9 @@
 </Headset>
 ```
 
-Like a Portal, it can exist anywhere in your Threlte application.
+If you need to only read from the current headset pose, the [useHeadset](/docs/reference/xr/use-headset) hook is available.
+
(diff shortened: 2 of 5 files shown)


packages/xr/src/lib/internal/headset.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const pose = xr.getFrame().getViewerPose(space)
+
+    if (pose === undefined) return
+
+    const { position, orientation } = pose.transform
+
+    headset.position.set(position.x, position.y, position.z)
+    headset.quaternion.set(orientation.x, orientation.y, orientation.z, orientation.w)
+  }, { autostart: false })
```
grischaerbe, line 25:
```suggestion
  }, { autostart: false, invalidate: false })
```

packages/xr/src/lib/internal/headset.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const nonImmersiveFrame = useFrame(() => {
+    headset.position.copy(camera.current.position)
+    headset.quaternion.copy(camera.current.quaternion)
+  }, { autostart: false })
```
grischaerbe, line 30:
```suggestion
  }, { autostart: false, invalidate: false })
```

grischaerbe (changes requested):
Hey @michealparks,

absolutely clean execution, I like it a lot. Great use of `watch` and `useFrame` ❤️ . I added two comments because I *think* the framehandlers don't need to invalidate, which is practically only important for non-XR content.

PR #671, opened 2023-10-03
useTexture enhancement proposal

### LINKED ISSUE #655: useTexture - default colorSpace should match Canvas
useTexture default is `THREE.NoColorSpace` and since our renderer is sRGB by default, it leads to confusing situations where material colors are correct, but Textures are loaded in the wrong colorSpace. 

Default color space used by useTexture hook should match <Canvas/> colorSpace, unless otherwise specified

jerzakm:
work in progress:
1. texture's color space matches renderer color space by default
2. added a third prop to assign options to a texture when it's loaded

closes #655

### DIFF
--- a/.changeset/nervous-hounds-drum.md
+++ b/.changeset/nervous-hounds-drum.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': major
+---
+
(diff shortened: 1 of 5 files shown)


DefinitelyMaybe:
Found this! I noticed the difference earlier when I upgraded my dependencies. Thank you 🙏 looks much better now

PR #676, opened 2023-10-04
[extras] Add `ref` binding to `<InstancedMesh>`

michealparks:


packages/extras/src/lib/components/Instancing/InstancedMesh.svelte
```diff
@@ -17,6 +18,9 @@
   export let range = 1000
   export let update = true
 
+  // forward ref binding
+  export let ref: InstancedMesh | undefined = undefined
```
grischaerbe, line 22: Because the actual instance is not going to change, it's more comfortable to just instantiate one here, so we don't have to deal with `| undefined` upstream when binding.

```suggestion
  export const ref = new InstancedMesh(null as any, null as any, 0)
```

michealparks: Good point!

michealparks:
>  create the instance in the component code

@grischaerbe I'm not sure I follow, could you explain a bit more?

grischaerbe:
I hope you're fine with me adding the finishing touches, sorry it took so long and thanks as always ❤️

michealparks:
You can finish my PRs anytime! 😄

PR #679, opened 2023-10-07
Allow sessionInit overrides and fix useHeadset android chrome bug

michealparks:
This PR:
* Fixes an issue where headset pose can be `null` in Chrome Android - causing the app to crash. I observed this on a Pixel 4A.
* Adds the ability to override `sessionInit` on VR and AR buttons.
* Adds a nativeEvent to an XRButton click.


grischaerbe (approved):
Awesome, thanks 🙏

PR #691, opened 2023-10-12
Force svelte/motion to use session.requestAnimationFrame when presenting

michealparks:
This PR uses the `set_raf` svelte internal to force motion to use session.requestAnimationFrame. Thank you @grischaerbe for helping investigate this!

I tested that this works with the new pointer controls example on that branch.

### DIFF
--- a/.changeset/clean-students-shout.md
+++ b/.changeset/clean-students-shout.md
@@ -1,0 +1,5 @@
+---
+'@threlte/xr': patch
+---
+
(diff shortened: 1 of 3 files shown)


grischaerbe (approved):
Good solution! When I tested it, it for some reason wasn't working when I wouldn't set `set_now`, [see this REPL](https://svelte.dev/repl/b584ad8a0bea4dd6954a5b4a869ec832?version=4.2.1), but maybe there's another reason for that 🤔

PR #699, opened 2023-10-17
Improve linting, format, typecheck

michealparks:
### Overview 

This is a small maintenance PR mainly meant to begin to start to improve linting. I started with the XR package, and if we like the direction I'm going in I'd be happy to keep up the janitorial work on other packages! 

Things done:

* Updated xr from the eslint svelte3 plugin (now deprecated) to the officially supported plugin.
* Fixed easily problems in xr and extras after running `pnpm lint` and `pnpm check`.
* Ran `pnpm format` on all of xr.
* I was able to remove lodash-es in extras since it was used only in one place

I didn't add a changeset here since this was all tidying 💅  - but let me know if I should!


packages/xr/tsconfig.json
```diff
@@ -2,31 +2,33 @@
   "extends": "./.svelte-kit/tsconfig.json",
   "include": ["./src/**/*"],
   "compilerOptions": {
+    /**
```
michealparks, line 5: @grischaerbe Fixed this deploy! This was the issue.

PR #711, opened 2023-11-01
Instanced Animated Sprites

jerzakm:
Most of it doesn't quite work yet

preview is at: examples/extras/instanced-sprite


apps/docs/src/content/reference/extras/instanced-spritesheet.mdx
```diff
@@ -0,0 +1,230 @@
...
+            required: false,
+            description: 'Url of a spritesheet json with metadata.'
+          },
+          {
+            name: 'textureUrl',
+            type: 'string',
+            required: false,
+            description: 'Url of a spritesheet image. This will get overwritten if you declare spritesheet meta with help of <File/> and <Animation/> child components.'
+          },
+					{
```
michealparks, line 24: Why is this boolean needed to load json metadata? Is it not enough to have dataUrl?

apps/docs/src/content/reference/extras/instanced-spritesheet.mdx
```diff
@@ -0,0 +1,230 @@
...
+```svelte
+<Spritesheet
+  let:File
+  let:Animation
+>
+  <File
+    textureUrl="/textures/sprites/cacodaemon.png"
+    options={{
+      type: 'rowColumn',
+      w: 8,
```
michealparks, line 74: Would recommend avoiding abbreviations for code readability (`w` -> `width and `h` -> `height`)

apps/docs/src/content/reference/extras/instanced-spritesheet.mdx
```diff
@@ -0,0 +1,230 @@
...
+**One spritesheet file with multiple animations**
+<details>
+	<summary>Preview sprite</summary>
+	*cacodaemon.png*
+	![Multi image spritesheet](/textures/sprites/cacodaemon.png)
+</details>
+
+```svelte
+<Spritesheet
+  let:File
```
michealparks, line 95: If possible I'd avoid using component `let` syntax, since this will eventually be deprecated and removed from Svelte

apps/docs/src/content/reference/extras/instanced-spritesheet.mdx
```diff
@@ -0,0 +1,230 @@
...
+  let:Animation
+>
+  <File
```
michealparks, line 98: Is it possible to role `<File>` and `<Animation>` up into `<Spritesheet>`?

```html
<Spritesheet textureUrl={...} options={...} animations={...} />

File: apps/docs/src/examples/extras/instanced-sprite/Trees.svelte [hunk 5]
```diff
@@ -0,0 +1,106 @@
...
+    if (sprite) {
+      for (let i = 0; i < points.length; i++) {
+        sprite.flipX.setAt(i, Math.random() > 0.6 ? true : false)
+      }
+      sprite.update()
+    }
+  }
+</script>
+
+<Spritesheet
```
michealparks, line 57: It feels a little strange having configuration as components. Even though it's complex, I would maybe roll all of this up into two components (which would result in one extras export):

```html
<InstancedSprite
  let:Instance
  files={[
    textureUrl: '/textures/sprites/trees-pixelart.png',
    options: { ... },
    animations: [ ... ],
  ]}
>
  <Instance />
  <Instance />
  ...
</InstancedSprite>

File: apps/docs/src/content/reference/extras/instanced-spritesheet.mdx (hunk 1 above)
Comment (jerzakm) [reply apps/docs/src/content/reference/extras/instanced-spritesheet.mdx:24]: This is strictly for aseprite jsons. It then gets (roughly-wip) translated to the native spritesheet format for my library looks like this.

```ts
type SpritesheetFormat = {
  frames: [x: number, y: number, w: number, h: number][];
  animations: Record<string, [frameId: number, duration: number][]>;
  sheetSize: [w: number, h: number][];
  animationLengths: number[];
};
```

You could load a json with this format too. Maybe in the future if there's demand it could be like:
`dataFormat: 'aseprite' | 'three-kit' | 'anotherProgram' | 'anotherProgram'`

apps/docs/src/examples/extras/instanced-sprite/Trees.svelte
jerzakm: I see your point and I've reached a similar conclusion today, which is why I tried out a context-builder approach in the `VariantD.svelte` file

I don't like spritesheet configuration being provided as props on `<InstancedSprite/>` because it makes it harder to reuse it. 

Doing something like this would be awkward:

```html
<!-- same behaviour if <InstancedSprite> are wrapped in <Spritesheet> -->
{#await cacodaemonSpritesheet.sheet then { spritesheet, texture }} 
	<!-- hue -->
  <InstancedSprite
    {count}
    {billboarding}
    {spritesheet}
    {texture}
    hueShift={{
      h: 0.3,
      s: 1.5,
      v: 1.5
    }}
  >
    <UpdaterFlyingHook />
  </InstancedSprite>

	<!-- no hue -->
	<InstancedSprite
    {count}
    {billboarding}
    {spritesheet}
    {texture}    
  >
    <UpdaterFlyingHook />
  </InstancedSprite>
	<!-- no hue -->
	<InstancedSprite
    {count}
    {billboarding}
    {spritesheet}
    {texture}    
  >
		<!-- Different behaviour? Team? Culling? -->
    <DifferentUpdater />
  </InstancedSprite>
{/await}
```

https://github.com/threlte/threlte/pull/711/files/8a98a79236c325ca6ad2ca51b847e38991c87cfb#diff-92eff44b60f56ab50d5fa8a0aac2977d897e2bc847f451940ef02df7f1738e58

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+Three.js environment. The `<InstancedSprite>` component lets you bind to it through `ref`.
+
+The class extends the capabilities of the [troika's InstancedUniformsMesh](https://protectwise.github.io/troika/three-instanced-uniforms-mesh/). For an in-depth exploration of `InstancedSpriteMesh` and its features, refer to the documentation available at [InstancedSpriteMesh docs](https://three-kit.vercel.app/instancedsprite/01-instanced-sprite-mesh/).
+
+
+## Spritesheets
+
+### SpritesheetMetadata
+
+Object used in `buildSpritesheet` function has to be compliant with the `SpritesheetMetadata` type format. This type is structured to
```
ixxie, line 325: *The* object

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+  height: number
+  animations: {
+    name: string
+    frameRange: [number, number]
+  }[]
+}[]
+
+```
+
+#### Understanding SpritesheetMetadata
```
ixxie, line 342: Remove this heading; the contents of the subsection flow well with the previous content, so its not needed.

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+For improved developer experience when working with TypeScript, it is strongly recommended to use `as const` assertion in
+combination with `satisfies SpritesheetMetadata`. This approach not only ensures compliance with the `SpritesheetMetadata`
+type but also enables autocompletion for animation names within utility functions, which is highly recommended:
+
+
+### `buildSpritesheet()`
+
+`buildSpritesheet()` is a utility function for building a final texture and spritesheet object from a provided `SpritesheetMetadata` object or external source.
+Each `buildSpritesheet` method return an Promise that and has to be awaited. Promise returned by each method contains an object with a `spritesheet`
+ready for use in `<InstancedSprite>`.
```
ixxie, line 364: You can just say the methods are async.

I have to say its a bit weird to me that you need to make a call to access some methods.

Have you considered either:
1. a simple object with methods
2. a `Spritesheet` class with a constructor that accepts `arg: string | SpritesheetMetadata` and instantiates accordingly

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+
+#### buildSpritesheet().from(meta: SpritesheetMetadata)
```
ixxie, line 366: This title is horrible. How about `Building from SpritesheetMetadata`

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+      ]
+    }
+  ] as const satisfies SpritesheetMetadata
+
+  const result = buildSpritesheet.from<typeof meta>(meta)
+```
+
+![Tree sprite atlas](/images/docs/extras/instanced-sprite/useInstancedSpriteAutocomplete.png)
+
+#### buildSpritesheet().fromAseprite(asepriteDataUrl: string, spriteImageUrl: string)
```
ixxie, line 392: Same here. Don't use code snippets as titles, unless its a simple function / class name.

Do keep this snippet as a usage example, but put it in an inline or self-standing code block in the section content like a sane person.

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+
+### `buildSpritesheet()`
+
+`buildSpritesheet()` is a utility function for building a final texture and spritesheet object from a provided `SpritesheetMetadata` object or external source.
+Each `buildSpritesheet` method return an Promise that and has to be awaited. Promise returned by each method contains an object with a `spritesheet`
+ready for use in `<InstancedSprite>`.
+
+#### buildSpritesheet().from(meta: SpritesheetMetadata)
+Other than `spritesheet` promise, it also returns a `useInstancedSprite` hook. This hook can be enhanced with
+extra typescript support for autocompletion of animation names as such:
```
ixxie, line 368: Super confusing, both the text in itself and also the fact I don't see this hook in the code below.

apps/docs/src/content/reference/extras/instanced-sprite.mdx
```diff
@@ -0,0 +1,454 @@
...
+      animations: [{ name: 'death', frameRange: [0, 3] }]
+    },
+    {
+      url: '/textures/sprites/goblin/Idle.png',
+      type: 'rowColumn',
+      width: 4,
+      height: 1,
+      animations: [{ name: 'idle', frameRange: [0, 3] }]
+    }
+  ] as const satisfies SpritesheetMetadata
```
ixxie, line 443: Why does it need to be `as const satisfies SpritesheetMetadata`? Why not `const goblinSpriteMeta: SpritesheetMetadata = [...]`?

apps/docs/src/content/examples/animation/Complex sprite scene.mdx
```diff
@@ -0,0 +1,132 @@
...
+	useTask(()=>{
+		for(let i=0; i<count;i++){
+			updatePosition(i, [0,0,0])
+			sprite.animation.setAt(i, 'fly')
+		}
+	})
+</script>
+```
+
+## Various approaches
```
ixxie, line 102: Let's be more specific and title this `## Instancing approaches`

apps/docs/src/content/examples/animation/Complex sprite scene.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+This section goes over each component used in the scene and provides a short explanation of different approaches used with `<InstancedSprite/>` component.
+Every component is designed differently, aimed to present varied approaches to updating instance properties, loading and defining spritesheets.
+
+### From json (people)
```
ixxie, line 107: For this an all other subsections that follow, remove the `(sprite)` from the title (people in this case). You already open each section with a clear codeblocked file reference, and the extra noise in the right navbar is unhelpful.

apps/docs/src/examples/extras/instanced-sprite/UpdaterWalking.svelte
ixxie, line 1: Again I will say, I'm not a fan of the `UpdaterFoo` naming.

How about `behavior`? If I am reading it correctly, the [analogous unity concept is `StateMachineBehavior`](https://docs.unity3d.com/ScriptReference/StateMachineBehaviour.html).

I would suggest calling the files `DudeSprites.sveltes`, `TreeSpriteAtlas.svelte` and `WalkingBehavior.svelte`.

apps/docs/src/examples/extras/instanced-sprite/UpdaterWalking.svelte
```diff
@@ -0,0 +1,161 @@
...
+    if (e.key === 'd' || e.key === 'ArrowRight') dirs.right = true
+    if (e.key === 'w' || e.key === 'ArrowUp') dirs.up = true
+    if (e.key === 's' || e.key === 'ArrowDown') dirs.down = true
+  }
+
+  const handleKeyUp = (e: KeyboardEvent) => {
+    if (e.key === 'a' || e.key === 'ArrowLeft') dirs.left = false
+    if (e.key === 'd' || e.key === 'ArrowRight') dirs.right = false
+    if (e.key === 'w' || e.key === 'ArrowUp') dirs.up = false
+    if (e.key === 's' || e.key === 'ArrowDown') dirs.down = false
```
ixxie, line 130: Its basically impossible to find that one dude you can control with the key bindings. What is this, "where is Waldo" at difficulty insane?

I suggest removing this from the reference docs. In the complex example, we can keep it, with some modifications (see comments on that example for more on this).

apps/docs/src/examples/instancing/instanced-sprite-mania/App.svelte
ixxie, line 1: I think this example is cool, and that with minor tweaks it could be awesome.

## Naming & Structure

I would suggest grouping things in folders and naming as follows:

```
behaviors/
  FlyingBehavior.svelte
  FlyingBehaviorHook.svelte
  WalkingBehavior.svelte
sprites/
  DudeSprite.svelte
  GoblinSprites.svelte
  DemonSprites.svelte  <- renamed from Flyers because its too close to the Flying behavior.
```
I'm on the fence about whether or not to keep the `Behavior` and `Sprite(s)` text in the names. Your call.

## Streamlined Sprites

I would suggest the following changes, with later ones being more challenging and thus lower priority:

1. **Homogenous sprite distributions** - the different sprites are distributed in different ways. When you zoom out you can see their distributions are different. Unifying the algorithm for how they are positioned and moved would make it *look* better, and should allow code sharing between the files to simplify them. I prefer the disc-like distribution to the square: it fits the environment better.
2. **Lower density of sprites** - the quantity of sprites here is good, but their density is high, making it hard to see the details you've put into their individual movement.
3. **One dude vs. many monsters** - now that we can see what is going on and that everything is more uniform, we could potentially see the one dude moving with keyboard controls. Let's change the scene so there is only ONE dude, differentiating him from the evil in the world. The dude is the player, and is controlled with the keyboard controls. Now the `WalkingBehavior` is purely for the `GoblinSprites`. 
4. **Simplified algorithms** - the algorithms you use for positioning and behavior are super complicated and verbose. If there is a way to simplify the code, reuse it whenever possible, or even use a third-party library - please do so.
5. **Camera follows player** - for bonus points if the camera follows the dude, decoupling when you zoom out and reattaching when you start to move with the keyboard again.
6. **Minigame: kill the monsters** - if you are really into it, you could try and turn this example into a minigame, where you can actually kill the goblins (and maybe even the demons). 
 
Some of this could also be a separate future PR, if you don't feel like taking it that far for this PR. Alternatively you could immediately pull out this part from the PR so we can merge the reference docs, allowing you to focus on this bit after.

I like the idea of a minigame example because it gives a great starting point for anybody interested in using Threlte for a sprite-based game.

ixxie (changes requested):
Well, here you go. Lots of comments, but this is really just the finishing touches. I think the PR is in very good shape, and besides some of the (very optional) ideas I pitched for the complex example, you should be able to fix everything up quickly.

apps/docs/public/textures/sprites/cacodaemon.png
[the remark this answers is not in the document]
jerzakm: We have a few other sprites in that folder, like the Threlte one. Even the player-dude sprite is Michael's and I'm reusing it. I generally agree with any approach that sorts out our assets better, but a lot of these can be shared in the future if the need arises.

apps/docs/src/examples/instancing/instanced-sprite-mania/SpriteAtlasTrees.svelte
```diff
@@ -0,0 +1,128 @@
...
+        { name: 'red_2', frameRange: [10, 10] },
+        { name: 'red_3', frameRange: [11, 11] },
+        { name: 'red_4', frameRange: [20, 20] },
+        { name: 'red_5', frameRange: [21, 21] },
+        { name: 'red_6', frameRange: [22, 22] },
+        { name: 'red_7', frameRange: [23, 23] },
+        { name: 'dead_0', frameRange: [16, 16] },
+        { name: 'dead_1', frameRange: [17, 17] },
+        { name: 'dead_2', frameRange: [18, 18] },
+        { name: 'dead_3', frameRange: [19, 19] }
```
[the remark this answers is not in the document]
jerzakm: The sprites aren't in order unfortunately. I could edit the .png file but as we discussed maybe we just keep the dumb long array for clarity.

ixxie (approved):
LVFGTM

PR #715, opened 2023-11-11
Update OrbitControls.svelte to use .js

Its-Just-Nans:
Update OrbitControls.svelte to use .js


Its-Just-Nans:
Repro
https://stackblitz.com/edit/astro-qlqkxr

With client:only, the `npm run build` works

with client:load, the command crash with 

image

grischaerbe:
Hi @Its-Just-Nans,

We already had a PR that applied the "fix" to all imports from Three.js examples, but fortunately the issue can be fixed as explained [in this comment](https://github.com/threlte/threlte/pull/618#issuecomment-1726430969).

Therefore, I'll close the PR. Nevertheless thank you for your effort!

Its-Just-Nans:
Okay, I'm waiting for it.

I'm currently using the `client:only` props in Astro and it works, as explained above ^^

PR #721, opened 2023-11-13
THREE.WebGLRenderer: A WebGL context could not be created - powerPreference: "high-performance"

Its-Just-Nans:
Hi

With my GPU, the `powerPreference: "high-performance"` which is the default on threlte wasn't working

I know it's my GPU because when I deactivate `Hardware Acceleration` on the browser the webpage is loading correctly

So just to say, if you have this issue, maybe just add

```js
<Canvas rendererParameters={{ powerPreference: "default" }}>
 //....
</Canvas>
```

**Why does  `powerPreference: "high-performance"` is the default in threlte ?**

On the discord: https://discord.com/channels/985983540804091964/1173742823141548072/1173742823141548072

### DIFF
--- a/packages/core/src/lib/lib/useRenderer.ts
+++ b/packages/core/src/lib/lib/useRenderer.ts
@@ -54,7 +54,7 @@ export const useRenderer = (ctx: ThrelteContext) => {
     rendererParameters: WebGLRendererParameters | undefined
   ): void => {
     ctx.renderer = new WebGLRenderer({
-      powerPreference: 'high-performance',
+      powerPreference: 'default',
       canvas,
       antialias: true,
       alpha: true,


michealparks:
I would recommend keeping this `high-performance` since that is likely the desired value for most users. `"default"` in many browsers often means a low power mode, so this change could cause a degraded experience for many Threlte applications. 

As far as I understand this issue is uncommon enough that avoiding a low-power mode is more valuable. I'm curious though, what's your specific GPU model?

grischaerbe:
As @michealparks said, it's a recommendation to the browser that your app needs higher GPU power. The actual implementation is down to the OS and should not break your app. If it does, it's very likely an issue of your system.

Its-Just-Nans:
It's weird, all the default three Canvas are not working (in the docs for example).

I will give the reference of the GPU soon

Its-Just-Nans:
internal GPU:  AMD ATI 05:00.0 Cezanne
external GPU: NVIDIA GeForce RTX 3070 Mobile / Max-Q

```
VGA compatible controller: NVIDIA Corporation GA104M [GeForce RTX 3070 Mobile / Max-Q] (rev a1) (prog-if 00 [VGA controller])
	Subsystem: Acer Incorporated [ALI] GA104M [GeForce RTX 3070 Mobile / Max-Q]
```

PR #722, opened 2023-11-14
Add example for running multiple renderers inside threlte

BlueWinds:
As discussed in Discord, here is an example based on [le grisch's tip from back in May](https://discord.com/channels/985983540804091964/985983540804091967/1107961438246551552). The example shows how to run CSS2DRenderer in Threlte, applying HTML elements for use as flat labels on 3d objects inside a scene.

I have done my best to follow the style of the existing examples; please let me know if there's anything else you'd like to see.

Recording 2023-11-21 at 10 23 17


jerzakm:
That's super useful! Looks almost ready to merge at a first glance, could you just change the styling on the buttons? I think something went wrong here😄 

image

BlueWinds:
@jerzakm - Ah, whoops. I have a browser extension for dark themes, it was inverting the background color on my screen. Pushed a fix.

apps/docs/src/content/examples/renderers/Flat text with CSS2D.mdx
```diff
@@ -0,0 +1,78 @@
+In this tutorial, we'll walk you through the process of running an additional Three.js renderer in parallel with Threlte's `<Canvas>`, while still leveraging Threlte's built-in elements. Specifically, you'll learn how to run CSS2DRenderer to add flat labels to objects in a three-dimensional scene.
+
+This example can be easily adapted to use CSS3DRenderer instead, if you want the elements to live "inside" the scene, rather than flat across the surface.
+
+<Example path="renderers/CSS2DRenderer" />
+
+## How does it work?
+
+In this scene, we run two renderers - the default one provided by threlte, and also a new CSS2DRenderer which we'll initialize. Threlte's renderer runs on a canvas element as normal, while our new renderer runs in a `<div>` with absolute positioning on top of it.
```
ixxie, line 9: Capitalize Threlte.

Maybe reword:

In this scene, we run two renderers: the default one provided by Threlte, and a new CSS2DRenderer which we will be initializing manually.

apps/docs/src/examples/renderers/CSS2DRenderer/CssObject.svelte
```diff
@@ -0,0 +1,29 @@
+<script>
+  import { T } from '@threlte/core'
+  import { CSS2DObject } from 'three/examples/jsm/renderers/CSS2DRenderer'
+
+  export let pointerEvents = false
+
+  let element
+
+  const cssobject = (node) => {
```
ixxie, line 9: Let's camelCase: `cssObject`

apps/docs/src/examples/renderers/CSS2DRenderer/App.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+  import { Canvas } from '@threlte/core'
+  import Scene from './Scene.svelte'
+</script>
+
+<div id="css-renderer-target" class="absolute top-0 left-0" style="pointer-events: none;" />
```
ixxie, line 6: No Tailwind in examples: please move this CSS - including the pointer events - to a `<style>` block. We can use the `#css-renderer-target` id as a selector.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
```diff
@@ -0,0 +1,59 @@
...
+<T.Mesh position.y={1}>
+  <T.BoxGeometry args={[2, 2, 2]} />
+  <T.MeshStandardMaterial color="hotpink" />
+</T.Mesh>
+
+<CssObject position={[-1, 2, 1]} center={[0, 0.5]}>
+  <button on:click={clickHello}>
+    Hello - {helloCount}
+  </button>
+</CssObject>
```
ixxie, line 43: Let's move this template and the counter logic into a reusable component, called `CounterLabel.svelte` (or something like that - feel free to adjust).

This will help make this file more readable, and put the emphasis on the CSS2DRenderer setup.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+<CssObject position={[1, 2, 1]} center={[0, 0.5]}>
```
ixxie, line 45: Aesthetic suggestion (optional): once you make the `<CounterLabel>` component, make three labels instead of two: `Hello`, `CSS` and `Renderer`. 

Position them symmetrically around the cube, so the when you look the corner that is between them, they will form a triangle point down:
    
```
    +--------2 
   /        /|
  /        / |
 1--------0  |
 |        |  |
 |        |  +
 |        | /
 |        |/
 +--------3 
```

1: Hello
2: CSS
3: Renderer

0: The camera should look at this so the other points are symmetric around it.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
```diff
@@ -0,0 +1,59 @@
...
+  import { T, useFrame, useRender, useThrelte } from '@threlte/core'
+  import { OrbitControls } from '@threlte/extras'
+  import { CSS2DRenderer } from 'three/addons/renderers/CSS2DRenderer.js'
+  import CssObject from './CssObject.svelte'
+
+  // Set up the CSS2DRenderer to run in a div placed atop the <Canvas>
+  const element = document.querySelector('#css-renderer-target')
+  const cssRenderer = new CSS2DRenderer({ element })
+
+  const { size } = useThrelte()
```
ixxie, line 12: Consider moving this before line 9 so that all the `cssRenderer` lines are together.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
```diff
@@ -0,0 +1,59 @@
...
+  $: cssRenderer.setSize($size.width, $size.height)
+
+  useRender(async (ctx) => {
+    await tick()
+    ctx.renderer?.render(ctx.scene, ctx.camera.current)
+    cssRenderer.render(ctx.scene, ctx.camera.current)
+  })
```
ixxie, line 19: Lets destructure the context to `{renderer, scene, camera}` since this is more idiomatic and in line with other examples in the docs.

This could use some comments: why are we doing using `tick()`? What happens after?

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
```diff
@@ -0,0 +1,59 @@
...
+<CssObject position={[1, 2, 1]} center={[0, 0.5]}>
+  <button on:click={clickWorld}>
+    World - {worldCount}
+  </button>
+</CssObject>
+
+<style>
+  button {
+    background: rgb(41, 43, 45);
+    border-color: rgb(71 85 105);
```
ixxie, line 54: Let's use flashier colors to show off we can use CSS! Maybe even use a CSS variable and pass it to our new `<LabelCounter>` as a style prop, so we can have each label have a different color?

apps/docs/src/content/examples/renderers/Flat text with CSS2D.mdx
```diff
@@ -0,0 +1,78 @@
...
+    cssRenderer.render(ctx.scene, ctx.camera.current)
+  })
+</script>
+```
+
+This ensures that both renderers execute on each loop, after the scene updates.
+
+### Setting up CssObject
+
+The other integral part is a component that accepts DOM contents in the default slot and places them in the scene and renders them with the THREE.CSS2DRenderer:
```
ixxie, line 35: ...with the ThreeJS `CSS2DRenderer`.

apps/docs/src/examples/renderers/CSS2DRenderer/CssObject.svelte
```diff
@@ -0,0 +1,29 @@
...
+  use:cssobject
+  style:pointer-events={pointerEvents ? 'auto' : 'none !important'}
+  style:will-change="transform"
+>
+  <slot />
+</div>
+
+{#if element}
+  <T {...$$restProps} is={CSS2DObject} args={[element]} let:ref>
+    <slot name="three" {ref} />
```
ixxie, line 27: It would be neat if we could demonstrate the use of this slot without overcrowding the scene, but I don't have a good idea of how.

ixxie (changes requested):
Thank you for this wonderful contribution; this is great work! The example is clear and the text is well written, and the code is elegant.

I've added a bunch of suggestions for improvements for the pedagogy. Some are optional. Looking forward to merging this!

In addition:

- [x] Let's rename the example to `CSS2DRenderer Overlay`
- [x] Add a note about performance - I notice this is a tad laggy on the older computer I am using at the moment. This makes sense with two renderers, but should be noted. Edit: it seems this is more of an issue in Firefox than Chrome. I don't know why, but if you can track down a known issue with CSS2DRenderer and Firefox, it would be good to link that.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
BlueWinds: Gave them some nice css gradients.

BlueWinds (commented):
> Let's rename the example to CSS2DRenderer Overlay

Done in upcoming commit. Will push once I finish going through the rest of the comments.

>  Add a note about performance - I notice this is a tad laggy on the older computer I am using at the moment. This makes sense with two renderers, but should be noted. Edit: it seems this is more of an issue in Firefox than Chrome. I don't know why, but if you can track down a known issue with CSS2DRenderer and Firefox, it would be good to link that.

I did a bit of digging, and think that I might've come across the source of the performance issue. https://threejs.org/docs/#api/en/core/Object3D.matrixAutoUpdate:

> When this is set (default true), it calculates the matrix of position, (rotation or quaternion) and scale every frame and also recalculates the matrixWorld property

We can set `scene.matrixWorldAutoUpdate = false` and manually call `scene.updateMatrixWorld()` in order to avoid duplicating the work, since if true, both renderers will traverse the entire tree. Solution based on https://github.com/mrdoob/three.js/issues/24521 (the property has been renamed since that issue was closed).

I've added a description of this to the example text.

I'm also on Firefox, but a pretty powerful machine, so hadn't noticed any performance issues; would you mind retesting and letting me know if this has fixed it?

BlueWinds:
Still not the prettiest thing, but is a bit more eye-catching than the flat grey buttons. :D

image

apps/docs/src/examples/renderers/CSS2DRenderer/CssObject.svelte
BlueWinds: I'll be honest - I don't know what this slot is for or how to use it. I copied the code in from the original example and didn't much think about it.

ixxie:
> Still not the prettiest thing, but is a bit more eye-catching than the flat grey buttons. :D
> 
> image

If you make one end of the gradient a darker shade of the same hue, it would look nicer. Maybe try `#4e6cf4` and `#1338db`?

ixxie (commented):
Almost there! I spotted some last things here...

In addition, see my above comment about making the gradient prettier.

Otherwise, we are almost ready to go!

apps/docs/src/examples/renderers/CSS2DRenderer/CssObject.svelte
```diff
@@ -0,0 +1,29 @@
...
+  export let pointerEvents = false
+
+  let element
+
+  const cssObject = (node) => {
+    element = node
+    return {
+      destroy() { element = undefined }
+    }
+  }
```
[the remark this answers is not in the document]
BlueWinds: That... is a really good question. :D Changed to use bind:this.

apps/docs/src/examples/renderers/CSS2DRenderer/CssObject.svelte
BlueWinds: I've removed it from the example, since it doesn't seem necessary to get everything working.

BlueWinds:
> If you make one end of the gradient a darker shade of the same hue, it would look nicer. Maybe try `#4e6cf4` and `#1338db`?

Colors updated.

Feels like we're getting close here, good improvements all around.

BlueWinds:
Aha, figured out (by asking) what the `three` slot is for, and added an example usage to the scene. Now we have threlte components nested inside our CssObject. I added a ball to each corner we're labeling.

image

ixxie:
Excellent work, one last thing (and I feel a little bad bringing up such a tiny issue after you have done such great work):

image

Indentation here is off for some reason.

BlueWinds:
> Excellent work, one last thing (and I feel a little bad bringing up such a tiny issue after you have done such great work):
> 
> Indentation here is off for some reason.

Fixed! My editor seems to want to switch to tabs rather than spaces today for some weird reason! And no worries, I understand on a deep level needing the indentation to be correct, would drive me nuts too. :D

ixxie (approved):
LGTM! Great work, I hope to see more PRs from you!

PR #726, opened 2023-11-18
[core] - Eagerly set canvas size on start

michealparks:
It appears that resize observer callback will do their first execution lazily - like in a microtask or on the next frame (did not have time to investigate). This PR takes care of a regression noticed [in this discord thread](https://discord.com/channels/985983540804091964/1174829936620605501) by eagerly setting the store value before calling `observe()`.


packages/core/src/lib/hooks/useResize.ts
```diff
@@ -1,50 +0,0 @@
-import { browser } from '../lib/browser'
```
michealparks, line 1: This file wasn't being used anywhere, safe to remove?

grischaerbe:
Ph, good catch! Immediate merge 👍

PR #733, opened 2023-11-24
Set `pointerEvents` for `noTransform` divs

### LINKED ISSUE #732: `<HTML>`'s `pointerEvents` prop doesn't work when `transform` is `false`
Thank you all for this wonderful project!

## Explanation

When using `<HTML>` from `@threlte/extras`, with the prop `pointerEvent='none'`, it doesn't apply the style to the element.
However, when searching the [doc](https://threlte.xyz/docs/reference/extras/html)s, I see you guys used `transform={true}`, & upon adding that, `pointerEvents` now apply.

## Repro

```svelte
<script>
  import { HTML } from '@threlte/extras';
</script>

<div class='container'>
  <Canvas>
    <HTML pointerEvents='none'>
      <!-- i can still click this! -->
      <button>hello</button>
    </HTML>
  </Canvas>
</div>

<style>
  .container {
    position: relative;
    height: 100%;
    width: 100%;
  }
</style>
```

sxxov:
Fixes #732

### DIFF
--- a/.changeset/shy-keys-own.md
+++ b/.changeset/shy-keys-own.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
+Set `pointerEvents` for `noTransform` divs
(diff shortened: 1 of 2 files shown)


grischaerbe:
Thanks a lot for fixing this @sxxov!

grischaerbe (approved):

PR #735, opened 2023-11-26
Frame Scheduling

grischaerbe:
🚧 WIP 🚧


apps/docs/src/content/learn/basics/scheduling-tasks.mdx
```diff
@@ -0,0 +1,290 @@
+---
+title: Scheduling Tasks
+category: Basics
+order: -750
+---
+
+In 3D apps and games, a lot of work is done in functions that run on every frame
+– from rotating an object continously around an axis to reacting to user input.
+Web-based apps rely on the browsers `requestAnimationFrame` that runs a callback
```
michealparks, line 9:
```suggestion
Web-based apps rely on the browser's `requestAnimationFrame` that runs a callback
```

apps/docs/src/content/learn/basics/scheduling-tasks.mdx
```diff
@@ -0,0 +1,290 @@
...
+function when a new frame is rendered. When encapsulating logic into smaller
+parts (i.e. *components*), we run into the issue that we need to run multiple
+callbacks on every frame that may be dependent on each other. For instance, we
```
michealparks, line 12:
```suggestion
parts (i.e. *components*), we often need to run multiple
callbacks that may be dependent on each other. For instance, we
```

apps/docs/src/content/learn/basics/scheduling-tasks.mdx
```diff
@@ -0,0 +1,290 @@
...
+## Scheduler
+
+Every Threlte app has a single scheduler. It is accessible via [`useThrelte()`](/docs/reference/core/use-threlte):
+
+```ts
+const { scheduler } = useThrelte()
+```
+
+<Tip type="info">
+  Most of the times you won't need to interact with the scheduler directly. It is
```
michealparks, line 43:
```suggestion
  Usually you won't need to interact with the scheduler directly. It is
```

apps/docs/src/content/learn/basics/scheduling-tasks.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+### Task Dependencies
+
+A common use case for tasks is to run code after another task has been executed. Imagine
+a game where an object is transformed by user input in one task and a camera follows that
+object in another task. The camera task should be executed after the object has been
+transformed.
+
+To control the order in which tasks are executed in a stage, you can pass a
+`before` and `after` option to `useTask` The tasks passed to these options are
```
michealparks, line 196:
```suggestion
`before` and `after` option to `useTask`. The tasks passed to these options are
```

apps/docs/src/content/learn/basics/scheduling-tasks.mdx
```diff
@@ -0,0 +1,290 @@
+---
+title: Scheduling Tasks
+category: Basics
+order: -750
+---
+
+In 3D apps and games, a lot of work is done in functions that run on every frame
+– from rotating an object continously around an axis to reacting to user input.
```
michealparks, line 8:
```suggestion
In 3D apps and games, a lot of work is done in functions that run on every frame.
```
Recommend removing this example since a similar one is given at the end of this paragraph.

apps/docs/src/content/learn/basics/app-structure.mdx
```diff
@@ -0,0 +1,94 @@
...
+<T.Mesh rotation.y={rotation}>
+  <T.BoxGeometry />
+  <T.MeshBasicMaterial color="red" />
+</T.Mesh>
+
+<Player />
+<World />
+```
+
+## Context Not Available
```
DefinitelyMaybe, line 69: If this will go live with the useTask then adjust `useFrame` use

grischaerbe: Oh yeah, we have to change quite some stuff. It's mostly a drop-in replacement though. Also, we were able to make `useFrame` and `useRender` still work as they always did, but with the new scheduling backend.

apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+The hook `useTask` is used to create a
+[task](/docs/learn/basics/scheduling-tasks#tasks) that is used to run code on
+every frame.
+
+## Creating an Anonymous Task
+
+In its most basic form, `useTask` takes a function as its first argument. This
+function will be executed on every frame, starting on the next frame and
+receives the delta time representing the time since the last frame as its first
```
michealparks, line 23:
```suggestion
function will be executed on every frame, starting on the next frame. It receives a `delta`, representing the time since the last frame as its
```

apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -0,0 +1,201 @@
...
+  tasks](/docs/learn/basics/scheduling-tasks) page.
+</Tip>
+
+The hook `useTask` is used to create a
+[task](/docs/learn/basics/scheduling-tasks#tasks) that is used to run code on
+every frame.
+
+## Creating an Anonymous Task
+
+In its most basic form, `useTask` takes a function as its first argument. This
```
michealparks, line 21:
```suggestion
In its most basic form, `useTask` takes a callback function as its argument. This
```

apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+```ts
+const { start, stop, started, task } = useTask((delta) => {
+  // This function will be executed on every frame
+})
+```
+
+It returns an object with the following properties:
+- `start`: A function that starts the task. It will be executed on the next
+  frame. Note that by default a task is started automatically.
```
michealparks, line 35: I don't see the `autostart` option mentioned in these docs. Maybe it should be added for newcomers?

apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -0,0 +1,201 @@
...
+object in another task. The camera task should be executed after the object has been
+transformed.
+
+To control the order in which tasks are executed in a stage, you can pass a
+`before` and `after` option to `useTask`. The tasks passed to these options are
+called **dependencies** and can be a task itself, the key of a task or an array
+of tasks or keys. The referenced tasks must be in the same stage as the task you
+are creating.
+
+Task dependencies **do not need to be created yet** if they are passed by key.
```
michealparks, line 80:
```suggestion
Task dependencies **can be created in any order** if they are passed by key.
```

apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -0,0 +1,201 @@
...
+  If a task is passed by reference to the `before` or `after` option, the task
+  created by `useTask` will automatically be added to the same stage as the task
+  it depends on. If you pass a key instead and the task you want to reference is
+  **not** in [Threlte's `mainStage`](#default-stages), you will also need to pass the
+  stage, either by value or key.
+</Tip>
+
+## Examples
+
+### Update Objects
```
michealparks, line 94:
```suggestion
### Updating Objects
```

apps/docs/src/content/reference/core/use-task.mdx
grischaerbe: done ✅

packages/core/src/lib/frame-scheduling/Scheduler.ts
```diff
@@ -0,0 +1,99 @@
...
+    } = {
+      tasks: true
+    }
+  ) {
+    return {
+      stages: this.mapNodes((stage) => {
+        if (stage === undefined) throw new Error('Stage not found')
+        return {
+          key: stage.key.toString(),
+          ...{ tasks: include.tasks ? stage.getSchedule() : undefined }
```
michealparks, line 94:
```suggestion
          tasks: include.tasks ? stage.getSchedule() : undefined
```

grischaerbe: This will however add the key `tasks` to the object and a consumer who's just checking for the existence of the key might not realize that including the tasks was set to `false`. Spreading here will not add the key.

packages/core/src/lib/hooks/useTask.ts
```diff
@@ -0,0 +1,172 @@
...
+  options?: ThrelteUseTaskOptions
+): ThrelteUseTask
+export function useTask(
+  keyOrFn: string | ((delta: number) => void),
+  fnOrOptions?: ((delta: number) => void) | ThrelteUseTaskOptions,
+  options?: ThrelteUseTaskOptions
+): ThrelteUseTask {
+  if (!browser) {
+    return {
+      task: undefined as any,
```
michealparks, line 66: Could this be `undefined as Task`?

packages/core/src/lib/frame-scheduling/Scheduler.ts
michealparks: Oh! Makes sense, totally missed that

PR #738, opened 2023-11-28
Docs and SEO pass

michealparks:
* Improves lighthouse score for threlte.xyz

Untitled

* Improves some docs language, and replaces some cases of useFrame with useTask


grischaerbe (approved):
Awesome, thanks! ❤️

PR #739, opened 2023-11-28
chore: update extras components to useTask

DefinitelyMaybe:
First the good news. These components have been tested in the docs and work the same as before.

~~the interesting news. Some components errored out looking for the `after` prop but wasn't supplied any additional options. [HTML for ex](https://github.com/threlte/threlte/blob/e260da1ea4510d6ac553c64eed036e4f976e1c78/packages/extras/src/lib/components/HTML/HTML.svelte#L283C5-L283C5).~~

still to go:

- [x] AnimatedSpriteMaterial
- [x] Contact shadows
- [x] Float
- [x] HTML
- [x] instance
- [x] gamepad

### DIFF
--- a/packages/extras/src/lib/components/Align/Align.svelte
+++ b/packages/extras/src/lib/components/Align/Align.svelte
@@ -93,7 +93,7 @@
       invalidate()
       stop()
     },
-    { autoStart: false }
+    { autoStart: false, autoInvalidate: false }
   )
 
   /** Force a recalculation of the bounding box. */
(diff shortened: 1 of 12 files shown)


packages/extras/src/lib/components/Align/Align.svelte
```diff
@@ -80,16 +87,13 @@
...
+    () => {
       calculate()
       invalidate()
       stop()
     },
-    {
-      autostart: false,
-      invalidate: false
-    }
+    { autoStart: false }
```
michealparks, line 96:
```suggestion
    {
      autoStart: false,
      autoInvalidate: false
    }
```

michealparks:
Thank you for this! Could you add a major version changeset bump?

DefinitelyMaybe:
Thank you for the review 🙏

yes a changeset is also a good idea too

PR #740, opened 2023-11-28
chore: update docs to useTask

DefinitelyMaybe:
related: #739

This started as a findandreplace query using vscode so will need extra eyes to make sure it hasn't broke anything. Have done a first pass.

Also did a quick check on `useFrame((state, delta) => ` and nothing looks like it was actually using `state` 😄 

Doing the findandreplace has somehow also run a formatting pass... didn't know there was a bunch to do there, unless that's something my setup has inadvertently done

### DIFF
--- a/apps/docs/src/examples/core/use-frame/Scene.svelte
+++ b/apps/docs/src/examples/core/use-frame/Scene.svelte
@@ -1,17 +1,20 @@
 <script lang="ts">
-	import { BoxGeometry, MeshStandardMaterial } from 'three'
-	import { T, useFrame } from '@threlte/core'
+  import { BoxGeometry, MeshStandardMaterial } from 'three'
+  import { T, useTask } from '@threlte/core'
 
-	let rotation = 0
-	useFrame(() => {
-		rotation += 0.01
-	})
+  let rotation = 0
+  useTask(() => {
+    rotation += 0.01
+  })
 </script>
 
-<T.DirectionalLight position.y={10} position.z={10} />
+<T.DirectionalLight
+  position.y={10}
+  position.z={10}
+/>
 
 <T.Mesh
-	rotation={[rotation, rotation, rotation]}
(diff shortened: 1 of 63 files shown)


DefinitelyMaybe:
Oh wait! Can someone please check the plugins markdown page. I think it had `order:-infinity` in there and I might've just removed instead of properly translating that

apps/docs/src/examples/core/use-frame/Scene.svelte
grischaerbe, line 1: this example probably has to be moved since it's called `useFrame`

apps/docs/src/examples/core/use-render/Scene.svelte
grischaerbe, line 1: We remove this example entirely.

apps/docs/src/examples/renderers/CSS2DRenderer/Scene.svelte
grischaerbe, line 1: `useRender` is used here, so replace with a `useTask` added to the `renderStage`

grischaerbe: Or even better, we introduce a new stage here: a `cssRenderStage` that the task is added to and which also only runs its tasks when its needed (`shouldRender()`) and after the `renderStage`.

```ts
const { shouldRender, renderStage } = useThrelte()

const cssRenderStage = useStage('css-render-stage', {
	after: renderStage,
	callback: (delta, runTasks) => {
		if (shouldRender()) runTasks()
	}
})

useTask(() => {
	// css render here, but leave the actual rendering out.
}, { stage: cssRenderStage })
```

grischaerbe (approved):
Thanks so much @DefinitelyMaybe! I'll approve this one and work in the changes in a separate PR since it's minor changes and getting the docs on the way is super good 🙂

PR #752, opened 2023-12-01
<AnimatedSpriteMaterial> reverse animation direction

grischaerbe:
- Added support for reverse animation direction in `<AnimatedSpriteMaterial>`
- Added support for calling bound `play()` immediately through usage of `Promise.all` on the async writable stores
- Removed unused `playing` variable
- Now respects `autoplay` prop when a new animation is requested but the animation before ended and the material is set to `loop={false}`.


apps/docs/src/examples/extras/animated-sprite-material/Fire.svelte
```diff
@@ -0,0 +1,33 @@
...
+  const light = new PointLight('#FF893D', 100)
+
+  let rate = 1 / 10
+  let fixedStepTimeAccumulator = 0
+  useTask((delta) => {
+    fixedStepTimeAccumulator += delta
+    while (fixedStepTimeAccumulator >= rate) {
+      fixedStepTimeAccumulator -= rate
+      // random light intensity between 22 and 44
+      light.intensity = Math.random() * 24 + 22
```
michealparks, line 15: Ah! Love it. Fixed step strikes again!

michealparks (approved):
🚀

PR #758, opened 2023-12-04
🚧  Svelte 5 compatibility 🚧 

michealparks:
### Overview

This PR is up to explore potential problems with and solutions for migration to Svelte 5, and to provide a reference for Svelte and Threlte maintainers. 

### Current status (April 3, 2024)

⚠️ Partially working

* In the `@threlte/core` default sveltekit route, the scene is rendering. 
* Some docs examples are rendering.
* The T component works partially. ~Passing `$$restProps` to `<T>` changes the structure of incoming props to be a proxy, which when making a proxy T component and adding additional props ends up making all props undefined. No clear solution for this yet.~ Fixed
* `let:ref={someRef}` is not working
* [`createRawEventDispatcher`](https://threlte.xyz/docs/reference/core/utilities#createraweventdispatcher) and [`forwardEventHandlers`](https://threlte.xyz/docs/reference/core/utilities#forwardeventhandlers) cannot be made forward compatible with Svelte 5, because no Svelte 5 hook exists to query all events listeners from inside a component. The use of `restProps.$$events` will not work, because $props() is not accessible within a hook.
* `props.$$events` only works with svelte 5 components, so any component with instances of raw event dispatching needs to be converted to a svelte 5 component.
* The [`transitions`](https://threlte.xyz/docs/reference/extras/transitions) plugin is not compatible with Svelte 5 due to use of `svelte/internal` transitions functions.
* There is no way to make a version of Threlte that is compatible with both Svelte 4 and 5 due to threlte's heavy usage of `get_current_component()`.


tsdexter:
any additional work/plans to get threlte working with svelte 5?

jerzakm:
Michael has been doing a lot of experiments with svelte 5 and we're going to make a transition as smooth and as fast as reasonably possible. There isn't a build out there yet because we don't have the bandwith to maintain both svelte 5 and svelte 4 versions of the library at the same time.

tsdexter:
> Michael has been doing a lot of experiments with svelte 5 and we're going to make a transition as smooth and as fast as reasonably possible. There isn't a build out there yet because we don't have the bandwith to maintain both svelte 5 and svelte 4 versions of the library at the same time.

Thanks for the quick reply... that is great to hear that I can continue to rely on threlte going forward.

I'll keep an eye on it to see if there is anywhere I can help out.

michealparks:
Exactly! I've also been taking this pretty slow because there are still a lot of unknowns with svelte 5. Features are still being pushed into the alpha branch, and it's unclear if Svelte 5 will completely remove `svelte/internals` support, which would heavily affect how initial Svelte 5 support for Threlte would look. Once things settle and an RC of Svelte 5 lands then I assume we'll pick up the pace and release initial support pretty quickly.

michealparks:
Closing because of svelte 5 compatibility in the `@next` branch 🎉

tsdexter:
thank you @michealparks 🙏

PR #768, opened 2023-12-05
Allow reactive `<Portal>` ids

michealparks:
I currently have the following use case for `<Portal>`:

```svelte
<Portal {id}>
  <T.Mesh>...</T.Mesh>
</Portal>
```

Where `id` is reactive. The current portal doesn't support reactive ids, so this PR adds it!


grischaerbe (approved):
LGTM 🚀

PR #770, opened 2023-12-06
Gizmo visuals

grischaerbe:
@ethanlook, I hope you're fine with me doing some adjustments to the visuals. I corrected the colors of the canvas textures as well as adding a `renderOrder` to the sprites so that the overlap of the sprites with the "axis stems" do not create a transparency artifact.
I also added a mechanic that if you click on a target rotation that is already reached, the camera rotates towards the opposite side, so if you click on Y+ once, it's going to rotate to Y+. Then, if you click on Y+, it's going to rotate towards Y-. It's a common Gizmo mechanic and makes it intuitive to go to the opposite side of an axis.

### DIFF
--- a/.changeset/green-pianos-learn.md
+++ b/.changeset/green-pianos-learn.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': patch
+---
+
+Improved Gizmo visuals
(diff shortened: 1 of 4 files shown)


ethanlook:
Nice, I really like the improvements! The color looks much better.

PR #786, opened 2023-12-08
chore: update docs to use svelte-tweakpane-ui

DefinitelyMaybe:
Very much liking `Svelte-Tweakpane-UI` for the docs. The ability to use svelte features like `{#if ...}` blocks and `bind:value` while creating the UI is a huge bonus. For example:
```html
{#if gridType == 'polar' || gridType == 'circular'}
      <Slider
        bind:value={maxRadius}
        label="max radius"
        step={1}
        min={0}
        max={15}
      />
{/if}
```

Much more readable and easy to setup 🚀


grischaerbe:
Thanks so much @DefinitelyMaybe! If you mean that the panel is attached to the "host" window and not contained in the example viewport, then the fix is to add the prop `iframe` to the `<Example>`:

```mdx
<Example path="some/path" />
```

→

```mdx
<Example path="some/path" iframe />
```

This will embed the example in an iframe and the panel will only be allowed to float inside the example view. Be aware that you might need to make the `<div>` that holds the `<Canvas>` inside the example to fill out the whole "page" (i.e. iframe).

grischaerbe:
Check out the Gizmo example and markdown file for an example implementation 👍

DefinitelyMaybe:
Excellent! I hoped it was going to be a small thing. watch out `iframe` here I come 😄

grischaerbe:
We should probably make that the default behaviour at some point anyway.

DefinitelyMaybe:
Omgosh! as another bonus of this, there was some css which covered the area below the tweakpane which blocked touch/mouse input to the canvas - which now no longer is an issue 🎉 🎉

DefinitelyMaybe:
oops couple, more oversights. some extras iframes aren't showing. one moment

apps/docs/src/content/reference/xr/getting-started.mdx
```diff
@@ -29,11 +30,18 @@ The following adds a button to start your session and controllers inside an XR m
...
-
-<VRButton />
+<div>
+  <Canvas>
+    <Scene />
+  </Canvas>
+  <VRButton />
+</div>
+
+<style>
```
michealparks, line 40: Can we remove the div / styling in this case since this is just a snippet?

apps/docs/src/examples/animation/transitions/App.svelte
```diff
@@ -1,35 +1,44 @@
...
+  />
+  <Button
+    title="Run"
+    on:click={() => {
+      $buttonRun = !$buttonRun
+    }}
+  />
+</Pane>
+
+<div>
```
michealparks, line 32: The wrapping of the `<Canvas>` element with a div + the styling adds quite a bit of boilerplate when accumulated across all examples. Was this done because of issues that popped up when switching over to iframes? I wonder if there's a way to solve the problem in one place to avoid having to add this to every example...

michealparks:
This update is great! I noticed a few remaining issues when scanning through the docs:
-  Some examples are blank (noticed a few in the main examples page and in extras), I'm guessing because of switching to iframes they need to be updated to have a container div.
- When I tested the <Attractor> example the advanced buttons weren't working

One really nice thing that I noticed was that switching over to using iframes fixes a layout issue with the [immersive web emulator](https://chromewebstore.google.com/detail/immersive-web-emulator/cgffilbpcibhmcfbgggfhfolhkfbhmik) in the XR section!

apps/docs/src/content/reference/xr/getting-started.mdx
DefinitelyMaybe: sure thing 👍

DefinitelyMaybe:
Thank you! it has been quite an effort to re-check that all the functionality / embedded examples are working as before, I really appreciate the second pair of eyes on this 👍 I swear some of these were working but I re-checked in the vercel deploy annnd nope 😅 

Last ones to go:

- [ ] Rendering Points example
- [ ] Cursor Lines example
- [ ] Outlines example
- [ ] Interactive shader example 

Hey which part of the `<Attractor/>` advanced buttons? They _seem_ similar to the ones in the current deployment

DefinitelyMaybe:
On the cursor lines example, sometimes it works and sometimes I get this error: `The "position" attribute is likely to have NaN values.` printing out to console infinitely. I don't wont to tackled that within this PR

michealparks:
@DefinitelyMaybe Thanks for handling that! Found one last blank example:
- animated-sprite-material example 2

> Hey which part of the <Attractor/> advanced buttons

Woops! Nevermind, false alarm. Looks like it's working.

DefinitelyMaybe:
Have been through an exhaustive search through all pages except XR. I did check through each example/xr sources and all have a div and style tags so that should be good 👍 

found the look-at plugin example which needed its tags.

Future notes:

- the `useProgress` extras shows it's loading but doesn't display the helmet. There are a couple of warnings in the console too.
- the `rigidBody` rapier doesn't seem to be playing it's sound
- there is an update for theatre 😄

PR #789, opened 2023-12-10
Update Interactive shader.mdx

stijlmassi:
Previous link was dead, replaced with current working link.


grischaerbe:
Thanks so much! ❤️

PR #793, opened 2023-12-11
useLoader: Accept constructor arguments

grischaerbe:
- Enables the use of loaders that accept arguments

```ts
class SomeLoader {
  constructor(foo: string) {}
  loadAsync: (/* ... */) {}
}

const loader = useLoader(SomeLoader, {
  args: ['bar']
})
```

- Lazily creates loader before cache retrieval, so best case scenario doesn't create a loader at all

### DIFF
--- a/.changeset/afraid-schools-give.md
+++ b/.changeset/afraid-schools-give.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': patch
+---
+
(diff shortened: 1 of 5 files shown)


grischaerbe:
I added type tests in "useLoader.ts", not sure how to handle that in the future. Do long comments like this add to the bundle size?

PR #799, opened 2023-12-16
feat: add TrackballControls

MattFerraro:
Component Changes:
- Adds a TrackballControls.svelte component to extras
- Modifies TransformControls.svelte to be able to pause TrackballControls just like it pauses OrbitControls

Documentation changes:
- Adds TrackballControls.mdx page with working example
- Adds an example of using OrbitControls to OrbitControls.mdx (unrelated to the meat of this PR, just done for consistency)
- Modifies the TransformControls.mdx page to mention TrackballControls, and allow the user to switch between TrackballControls and OrbitControols in its example


ixxie:
Since you marked this as WIP, I converted it to a draft PR. When you feel its getting ready, you can convert it back @MattFerraro

michealparks:
> Fix the ambiguous indirect export: TrackballControls error that I see when I load the TrackballControls docs page

The issue is that you need to [export the TrackballControls here](https://github.com/threlte/threlte/blob/main/packages/extras/src/lib/index.ts); that should also fix your deploy.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
```diff
@@ -22,14 +22,17 @@
   type $$Slots = TransformControlsSlots
 
   export let autoPauseOrbitControls: $$Props['autoPauseOrbitControls'] = true
+  export let autoPauseTrackballControls: $$Props['autoPauseTrackballControls'] = true
   export let object: $$Props['object'] = undefined
 
   const { camera, renderer, invalidate, scene } = useThrelte()
 
-  const { orbitControls } = useControlsContext()
+  const { orbitControls, trackballControls } = useControlsContext()
```
ixxie, line 30: I'm not too familiar with this area of the codebase, but I wonder if we can avoid the repetition of code that occurs in the component in the current implementation.

Would there ever be a situation where `<TransformControls>` is in the context of _both_ `<OrbitControls>` _and_ `<TrackballControls>`? It seems unlikely to me. In this case, perhaps we can unify the logic?

This is not a binding comment since I am not sure I am right. I'd love @michealparks's feedback here also.

apps/docs/src/content/reference/extras/trackball-controls.mdx
```diff
@@ -0,0 +1,48 @@
...
+      extends:
+        {
+          type: 'TrackballControls',
+          url: 'https://threejs.org/docs/#examples/controls/TrackballControls'
+        }
+    }
+}
+---
+
+`<TrackballControls>` TrackballControls is similar to OrbitControls. However, it does not maintain a constant camera up vector. That means if the camera orbits over the “north” and “south” poles, it does not flip to stay "right side up"
```
ixxie, line 19: Could you add a little comment about the usecases for `<TrackballControls>` over `<OrbitControls>`?

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
```diff
@@ -0,0 +1,27 @@
...
+  import { TrackballControls } from '@threlte/extras'
+  import { BoxGeometry, MeshStandardMaterial } from 'three'
+</script>
+
+<T.PerspectiveCamera
+  makeDefault
+  position={[10, 5, 10]}
+  lookAt.y={0.5}
+>
+  <TrackballControls />
```
ixxie, line 12: Maybe we can also try a `svelte-tweakpane-ui` here to allow users the explore the settings of `<TrackballControls>`? Of course, a similar one should be added to `<OrbitControls>`.

Its likely the file will grow if you do this. In that event, maybe you can pull the `<T.PerspectiveCamera>` block into a separate `Camera.svelte` file to keep it clean.

I guess the tweakpane itself goes in the root `App.svelte` but see how we do it in other examples like `<Gizmo>`.

apps/docs/src/content/reference/extras/orbit-controls.mdx
```diff
@@ -18,6 +18,8 @@
 The component `<OrbitControls>` must be a direct child of a camera component and will mount itself to that camera.
```
ixxie, line 18: Maybe in the first line of this page, we should explain about the polar behavior which distinguished `<OrbitControls>` from `<TrackballControls>`. Then, we follow with a "see also..." sentence linking to `<TrackbacllControls>`. Marketing!

ixxie (changes requested):
Thank you @MattFerraro for the contribution! This is a very clean PR, so my comments are mostly enhancement requests and not any critical issues.

The main thing I am unsure about is the way `<TransformControls>` controls context is handled. I'd also like @michealparks to chime in on this.

apps/docs/src/examples/extras/transform-controls/Scene.svelte
```diff
@@ -9,16 +9,14 @@
   position={[10, 5, 10]}
   lookAt.y={0.5}
 >
-  <OrbitControls />
+  <TrackballControls />
```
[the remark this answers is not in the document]
MattFerraro: sure, done!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
```diff
@@ -0,0 +1,27 @@
...
+/>
+<T.AmbientLight intensity={0.3} />
+
+<T.GridHelper args={[10, 10]} />
+
+<T.Mesh
+  position.y={1}
+  geometry={new BoxGeometry(2, 2, 2)}
+  material={new MeshStandardMaterial()}
+/>
```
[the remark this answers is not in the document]
MattFerraro: done! I agree that Gizmo is even more valuable when using TrackballControls than when using OrbitControls because it is much easier to get disoriented. However it does highlight that Gizmo could probably be even more helpful if it set the camera up vector as well, at least when using the TrackballControl. I can handle that as a separate PR later.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
MattFerraro: yeah, the repetition feels icky but I don't yet know of a better way. My own rule of thumb is that repeating yourself up to 3 times is acceptable, but at the 4th time I try to find a better way 🤷 

Michael, if you have advice I'd be happy to follow it!

apps/docs/src/content/reference/extras/trackball-controls.mdx
MattFerraro: done!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
MattFerraro: sure, done!

I ended up pulling the camera out of `Scene.svelte` and putting it all in `App.svelte`. It's a lot of properties to pass around, but I could extract `Camera.svelte` if you feel strongly on this!

apps/docs/src/content/reference/extras/orbit-controls.mdx
MattFerraro: done!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
ixxie: Awesome work! I really like the feel of it.

Some notes:
- [ ] TailwindCSS shouldn't be used in examples.
- [ ] While I do see why you went with putting the camera directly in `App.svelte`, is contradicts our [officially recommended app structure](https://threlte-mytcm6ro7-threlte.vercel.app/docs/learn/basics/app-structure#recommended-app-structure) so I think its better to keep things consistent for pedagogic purposes.
- [ ] The tweakpane consumes a lot of lines and is of little interest pedagogically speaking, so lets move it to its own `Settings.svelte` or `Pane.svelte` component. When imported to `App.svelte`, put it under the `<Canvas>` in the template.
- [ ] Considering the limited screen real-estate, can you experiment and see if reducing the tweakpane's width is feasible?

apps/docs/src/content/reference/extras/trackball-controls.mdx
ixxie: Great writing! Super clear. 

Only comment is that links to components should include the both code styling and angled braces, e.g. [see links in the `<Sheet>` docs](https://threlte-mytcm6ro7-threlte.vercel.app/docs/reference/theatre/sheet). Same for the link you added in `<OrbitControls>`.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
ixxie: Its not so much repetition itself which bothered me, but a concern that this may cause other issues, like overhead or unexpected behavior. Let's wait for @michealparks's input.

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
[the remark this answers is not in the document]
ixxie: Great, and indeed its a good point the behavior is surprising with `<TrackballControlls>`.

Notes:
- [ ] Let's add some use the `<Gizmo>`'s padding props so it doesn't run up against the edge of the frame; if it takes too much space then you can make it smaller.
- [ ] `<OrbitControls>` deserves a `<Gizmo>` as well!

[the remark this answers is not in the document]
ixxie: Separate PR regarding `<Gizmo>` up vector handling seems like a good idea to me, but maybe discuss with @grischaerbe before submitting it. Note he is on vacation this week.

apps/docs/src/examples/extras/transform-controls/Scene.svelte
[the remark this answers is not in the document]
ixxie: Cool cucumbers!

Notes:
- [ ] No TailwindCSS allowed in examples
- [ ] Lets move the tweakpane below the `<Canvas>` in the template so the actual code of interest is in focus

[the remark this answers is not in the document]
MattFerraro: done!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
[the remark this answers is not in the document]
MattFerraro: done!

apps/docs/src/content/reference/extras/trackball-controls.mdx
MattFerraro: done!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
MattFerraro: Cool, I nuked all the TailwindCSS in OrbitControls, TrackballControls, TransformControls, and Gizmo. But there's lots of other occurrences in the docs which I'd prefer to change in a separate PR.

I was able to conform to the recommended app structure by splitting out `Settings.svelte` and including it directly in `Scene.svelte`. Does this new arrangement look good to you?

Reducing the tweakpane's width is straightforward:
```
<Pane
  width={150}
  ...
>
```
However, the automatically computed width seems to be the least wide that the TweakPane can be before it starts crunching the contents in a way that clips the edges, so I'm hesitant to introduce that change. Also, if it's a problem the user can collapse the whole TweakPane by clicking on the title of the Pane. Is that sufficient control for it to be a non-issue? Going deeper, the width of this Pane is dictated by how wide the Slider needs to be, but I could not find a straightforward way to control the width of a particular Slider. There is also a `scale` parameter we can play with, but it quickly makes the text hard to read on a non-retina screen. If width is an issue, can you include a screenshot? Maybe I'm just seeing something different than you are!

ixxie: One more request: can you double check the imports in all the files?
- Imports should be before any variables or props.
- Sometimes there are multiple imports from the same package in different lines - they need to be unified to one line.
- Import grouping: Threlte imports should come first, then Three imports and other 3rd party libs, then Svelte imports, and finally local imports. Groups should be separated with a blank line.

Sorry for this request since it should be something automated, but I haven't gotten around to investigate import ordering tools for Svelte.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
michealparks: Agreed that it feels awkward, but also agreed that we shouldn't overthink it until we're repeating ourselves many times. For now I think the type narrowing afforded by having two separate variables is more valuable. Having one `controls` store would make types more complicated.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
```diff
@@ -22,14 +22,17 @@
...
   const isDragging = writable(false)
   const useAutoPauseOrbitControls = writable(autoPauseOrbitControls) ?? true
   $: useAutoPauseOrbitControls.set(autoPauseOrbitControls ?? true)
+  const useAutoPauseTrackballControls = writable(autoPauseTrackballControls) ?? true
```
michealparks, line 34:
```suggestion
  const useAutoPauseTrackballControls = writable(autoPauseTrackballControls ?? true)
```
Boolean was accidently placed outside of the store args

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
```diff
@@ -22,14 +22,17 @@
...
   export let autoPauseOrbitControls: $$Props['autoPauseOrbitControls'] = true
+  export let autoPauseTrackballControls: $$Props['autoPauseTrackballControls'] = true
   export let object: $$Props['object'] = undefined
 
   const { camera, renderer, invalidate, scene } = useThrelte()
 
-  const { orbitControls } = useControlsContext()
+  const { orbitControls, trackballControls } = useControlsContext()
   const isDragging = writable(false)
   const useAutoPauseOrbitControls = writable(autoPauseOrbitControls) ?? true
```
michealparks, line 32:
```suggestion
  const useAutoPauseOrbitControls = writable(autoPauseOrbitControls ?? true)
```
I guess this was a preexisting issue 😂

michealparks, line 32: @ixxie @MattFerraro I wonder if in this case we should compress into a single variable. In the other case splitting is valuable due to types but here we could reduce cognitive complexity for the end user by changing this to:
```svelte
<TransformControls autoPauseControls />
```
which would pause any camera controls detected with the aforementioned context. That to me seems like a simpler API. It would also require more work in this PR and notifying people of a breaking change in the changeset, so no worries though if we'd want to push that to later.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
MattFerraro: whoa TIL that github has a suggested changes feature!

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
MattFerraro: lol, patched!

MattFerraro: yeah I think that's a better API, but I'd prefer to push it to later if that's okay!

apps/docs/src/examples/extras/trackball-controls/Scene.svelte
MattFerraro: Sounds good! I moved to `<style>` tags, extracted `Settings.svelte` from `Scene.svelte` and put it in `App.svelte`, and fixed all the imports. Ready for another look!

packages/extras/src/lib/components/controls/TrackballControls/TrackballControls.svelte
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!isCamera($parent)) {
+    throw new Error('Parent missing: <TrackballControls> need to be a child of a <Camera>')
+  }
+
+  export const ref = new ThreeTrackballControls($parent, renderer.domElement)
+
+  const { start, stop } = useTask(() => ref.update(), {
+    autoStart: true
+  })
```
michealparks, line 32:
```suggestion
  useTask(() => ref.update())
```
Can be simplified since `autoStart: true` is the default.

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
michealparks: It's such a great feature

packages/extras/src/lib/components/controls/TrackballControls/TrackballControls.svelte
```diff
@@ -0,0 +1,54 @@
...
+  {...$$restProps}
+  bind:this={$component}
+  on:change={invalidate}
+  on:create={({ ref, cleanup }) => {
+    trackballControls.set(ref)
+    ref.update()
+    cleanup(() => {
+      trackballControls.set(undefined)
+    })
+  }}
```
michealparks, line 51: Since we're instantiating `ThreeTrackballControls` in the `<script>` we could move all of this there to more immediately set these values:

```svelte
<script>
  import { onDestroy } from 'svelte'
   ...
   const { trackballControls } = useControlsContext()
   export const ref = new ThreeTrackballControls($parent, renderer.domElement)
   trackballControls.set(ref)
   onDestroy(() => trackballControls.set(undefined))
   ...
```
And it should be fine to just call `ref.update()` in `useTask` since that is called before the first frame is rendered.

michealparks (approved):
LGTM on my end after those last two small comments!

packages/extras/src/lib/components/controls/TrackballControls/TrackballControls.svelte
MattFerraro: done!

packages/extras/src/lib/components/controls/TrackballControls/TrackballControls.svelte
MattFerraro: cool, done!

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
MattFerraro: cool, gonna resolve this comment for now!

packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte
MattFerraro: fixed!

ixxie (approved):
Thanks for the hard and rapid work! Let's do this!

PR #800, opened 2023-12-17
[rapier] Upgrade to Threlte 7

michealparks:
This PR migrates all `@threlte/rapier` internal useFrame calls to useTask. I also added a test environment to `routes` 🎉


packages/rapier/src/lib/hooks/useFrameHandler.ts
```diff
@@ -160,6 +160,15 @@ export const useFrameHandler = (ctx: RapierContext, order?: number) => {
           rigidBodyDispatcher2
         } = getEventDispatchers(ctx, collider1, collider2)
 
+        if (
```
michealparks, line 163: Made a small optimization by early return here

grischaerbe (approved):
Awesome, thanks for this complete migration! LGTM 🚀

Issue #809, opened 2023-12-30
Gizmo click detection is off by a constant


MattFerraro:
I have a Gizmo on a page that includes an app bar and a tool bar:

![Screenshot 2023-12-30 at 10 16 38 AM](https://github.com/threlte/threlte/assets/1302431/7275d737-b73a-4ae3-9c7f-88ab20470cdb)

When I click on the X, -X, Z, -Y, buttons, the Gizmo does not respond. It behaves as though it is not receiving mouse clicks.

BUT, when I press the Y button, the camera moves to -Y. After some experimentation I found that the gizmo is receiving click events from the following locations:

![Screenshot 2023-12-30 at 10 16 38 AM](https://github.com/threlte/threlte/assets/1302431/e4836967-b433-49cc-b80c-247c9de6174d)

So I suspect there's a hardcoded assumption somewhere that the `<Canvas>` extends to the top of the viewport.

MattFerraro:
Digging in a bit, the assumption seems to be made on [this line](https://github.com/threlte/threlte/blob/a85cf5d3940e41ad667f4a191e2a01a03404aa21/packages/extras/src/lib/components/Gizmo/Gizmo.svelte#L104)

I'll see if I can make a patch!

MattFerraro:
PR is here: https://github.com/threlte/threlte/pull/810

DefinitelyMaybe:
closing for now. We have a [new gizmo](https://next.threlte.xyz/docs/reference/extras/gizmo) since #1254 was merged 🎉 

happy for the issue to be reopen by OP if needed

PR #810, opened 2023-12-31
fix: Gizmo click detection works properly when <Canvas> does not cover viewport

MattFerraro:
This PR addresses https://github.com/threlte/threlte/issues/809.

The `<Gizmo>` component does its own manual click detection because the `<Gizmo>` is not part of the `<Scene>`. To do that it creates its own dedicated div for listening to click events. To position that div correctly, the `<Gizmo>` component needs to understand the bounding rectangle of the DOM element that the renderer is attached to. If that DOM element is offset on the page by some X and Y, we need to respect those offsets when styling the `<Gizmo>`'s click div. Similarly, if the div that the `<Canvas>` is rendering to is shorter and thinner than the page overall, we need to account for that.

Here is one screenshot of the testing I performed to ensure this PR functions as intended even with padding on all four sides of the `<Canvas>`, and when placing the `<Gizmo>` in all four corners:
Screenshot 2023-12-30 at 11 32 16 PM

I did not include the changes I made to perform that testing in this PR because they make the example page ugly, but if you'd like to test it out you can check out https://github.com/threlte/threlte/commit/dc5c3aa3ab64b19312b094149142a0c59a3011a9 which is stored in my companion branch, `fix-gizmo-click-events-testing`

### DIFF
--- a/.changeset/funny-apricots-roll.md
+++ b/.changeset/funny-apricots-roll.md
@@ -1,0 +1,5 @@
+---
+'@threlte/extras': patch
+---
+
+fix Gizmo click events when the Canvas is not taking up the full viewport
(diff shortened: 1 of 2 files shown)


grischaerbe (approved):
Awesome, thanks for your in-depth explanation! LGTM 🚀

PR #813, opened 2024-01-07
Docs: small typo

lega0208:
Playing around with Threlte again after a long break, loving how far this has come ❤ you guys are awesome!


grischaerbe:
Thanks a lot @lega0208, good catch!

PR #821, opened 2024-01-13
Add port of Billboard component from drei

flo-bit:
Hi everyone! Really love threlte, after some getting used time, I thought I would try my hand at contributing starting with this port of drei's `Billboard` component.

Wasn't sure about the ordering of the examples, I added it to the end of the `components` part. This is my first time contributing to an open source project, so feedback very welcome!

### DIFF
--- a/packages/extras/src/lib/components/Billboard/Billboard.svelte
+++ b/packages/extras/src/lib/components/Billboard/Billboard.svelte
@@ -1,0 +1,50 @@
+<script lang="ts">
+	import { Group, Quaternion } from 'three';
+	import { T, useTask, useThrelte } from '@threlte/core';
+	import type { BillboardEvents, BillboardProps, BillboardSlots } from './Billboard.svelte';
+
+	type $$Props = Required<BillboardProps>;
+	type $$Events = BillboardEvents;
+	type $$Slots = BillboardSlots;
+
+	export let follow: $$Props['follow'] = true;
+	export let lockX: $$Props['lockX'] = false;
+	export let lockY: $$Props['lockY'] = false;
+	export let lockZ: $$Props['lockZ'] = false;
+
+	let inner: Group;
+	let localRef: Group;
+
+	const { camera } = useThrelte();
+
+	const q = new Quaternion();
+
+	useTask(() => {
+		if (!follow || !localRef) return;
+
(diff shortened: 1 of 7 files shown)


packages/extras/src/lib/components/Billboard/Billboard.svelte
```diff
@@ -0,0 +1,50 @@
...
+
+	const { camera } = useThrelte();
+
+	const q = new Quaternion();
+
+	useTask(() => {
+		if (!follow || !localRef) return;
+
+		// save previous rotation in case we're locking an axis
+		const prevRotation = localRef.rotation.clone();
```
michealparks, line 26: We can improve garbage collection pressure a bit by reusing an euler here:

```
const prevRotation = new Euler()

useTask(() => {
  ...
  prevRotation.copy(localRef.rotation)
  ...
})

File: packages/extras/src/lib/components/Billboard/Billboard.svelte [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+
+	let inner: Group;
+	let localRef: Group;
+
+	const { camera } = useThrelte();
+
+	const q = new Quaternion();
+
+	useTask(() => {
+		if (!follow || !localRef) return;
```
michealparks, line 23: Instead of doing an early return we can avoid calling this function altogether by starting and stopping it with reactive statements:

```ts
const { start, stop } = useTask(() => {
  ...
}, { autoStart: false })

$: if (follow && localRef) {
  start()
} else {
  stop()
}
```

packages/extras/src/lib/components/Billboard/Billboard.svelte
flo-bit: Thanks! Changed

packages/extras/src/lib/components/Billboard/Billboard.svelte
flo-bit: Changed!

michealparks: That line can be removed! There should be no need to perform that check every frame anymore 😄

grischaerbe (approved):
Awesome, thanks so much ❤️

PR #828, opened 2024-01-21
Fix minor error in your-first-scene.mdx

Dahoom152:
Tiny change, Just adjusted the page range to correctly highlight newly inserted lines.


Dahoom152:
The singular line also has to be changed, ill close this one and create a new one for that

PR #829, opened 2024-01-21
Fix for minor error in your-first-scene.mdx

Dahoom152:
Tiny change, Just adjusted the page range to correctly highlight newly inserted lines.


grischaerbe (approved):
Oh good catch, thanks for your contribution!

LGTM 🚀

PR #833, opened 2024-01-24
Upgrade package dependencies

michealparks:
Upgrades all dependencies for all packages that we ship to NPM.


grischaerbe (commented):
There are still 14 references to `@sveltejs/kit/vite` in this repo but I saw you were rewriting one in the XR package to `@sveltejs/vite-plugin-svelte`, do we need to do that change globally?

michealparks:
@grischaerbe Woops! Thanks for catching that. I believe `@sveltejs/kit/vite` is the correct import for sveltekit 2. I've reverted that change.

grischaerbe:
Do we need a version bump for this?

michealparks:
@grischaerbe I opted not to due to all updates being dev-dependencies, but curious if you disagree

packages/flex/package.json
```diff
@@ -14,35 +14,36 @@
...
+    "svelte2tsx": "^0.7.0",
+    "three": "^0.160.1",
+    "tslib": "^2.6.2",
+    "typescript": "^5.3.3",
+    "vite": "^5.0.12"
   },
   "dependencies": {
     "mitt": "^3.0.1",
-    "yoga-layout": "^2.0.0"
+    "yoga-layout": "^2.0.1"
```
DefinitelyMaybe, line 46: There's one minor here but I'm guessing not many, if any, would be worried about it

michealparks: Ah missed that one! I'll add a changeset to flex 👍

packages/xr/svelte.config.js
```diff
@@ -1,5 +1,5 @@
 import adapter from '@sveltejs/adapter-auto'
-import { vitePreprocess } from '@sveltejs/kit/vite'
+import { vitePreprocess } from '@sveltejs/vite-plugin-svelte'
```
michealparks, line 2: This import needed to be updated. The other packages use `svelte-preprocess`, a custom preprocessor and don't need to be updated.

Side discussion: should we remove `svelte-preprocess` and replace with this import? I could see this potentially causing issues for users if someone accidentally contributes a change that works with `svelte-preprocess` but a user is using the default `vitePreprocess`.

DefinitelyMaybe: There are two points where `<style lang="postcss">` are used. 1. `RightSidebar.svelte` and 2. a bunch of flex `DOM.svelte` files.

1. Would require inlining the tailwind classes
2. would just be an explicit copy paste where needed and would remove a bunch of warnings in the build output 👍 

+1 to inlining from me but maybe there was some reason they weren't inlined that I'm missing

michealparks: Oh I should have specified: I think it would make sense only to remove this plugin for packages that are shipped to NPM, since we ship the svelte files. For the docs I think it totally makes sense to keep this plugin + tailwind for DX.

PR #836, opened 2024-01-25
Astro update 

jerzakm:
Bumping astro to keep up with the changes


DefinitelyMaybe:
Hey @jerzakm !

This is a cool update from Astro's community 🎉 I'm seeing a reduction in kb transferred 😄 

Are there any other details to look into?

I noticed view transitions commented out. Hopefully that's coming back? Although I'm not noticing it within our current live so 🤷‍♂️

PR #838, opened 2024-01-28
Add port of r3f FakeGlowMaterial

flo-bit:
Small port of this [cool r3f glow effect](https://github.com/ektogamat/fake-glow-material-r3f) using a custom shader instead of postprocessing.

### DIFF
--- a/.changeset/pretty-carrots-know.md
+++ b/.changeset/pretty-carrots-know.md
@@ -1,0 +1,6 @@
+---
+'@threlte/extras': minor
+'@threlte/docs': minor
+---
+
+Add port of FakeGlowMaterial r3f component from ektogamat
--- a/apps/docs/src/content/reference/extras/fake-glow-material.mdx
+++ b/apps/docs/src/content/reference/extras/fake-glow-material.mdx
@@ -1,0 +1,63 @@
+---
+{
+  order: 11.5,
+  category: '@threlte/extras',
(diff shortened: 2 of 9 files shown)


DefinitelyMaybe (approved):
Thank you for the contribution! Docs LGTM!

The glsl is a carbon copy 👍 

Passing thought is how we might like folks to set up shadermaterials but I lack a strong opinion one way or the other.

flo-bit:
> Passing thought is how we might like folks to set up shadermaterials but I lack a strong opinion one way or the other.

Yeah I was thinking about that too, even set up a quick port of drei's shaderMaterials that the original used, but as a component it doesn't actually make that much sense, as it's not really that much easier/quicker to use:

```svelte
<T.ShaderMaterial uniforms={{ time: { value: 1.0 } }} {fragmentShader} {vertexShader} />
```

vs a port of drei's shaderMaterial as a svelte component:

```svelte
<ShaderMaterial uniforms={{ time: 1.0 }} {fragmentShader} {vertexShader} />
```

and the same goes for creating a shader material in the `script` tag, so I don't know (there is also something to be said for keeping the docs from overflowing imo).

Anyway I do have the port of drei's shaderMaterial basically done, so if you think it makes sense to add that, I can open another pull request and update this component anytime.

grischaerbe (approved):
Awesome, thanks for your contribution! LGTM 🚀

PR #840, opened 2024-01-29
Clean up: docs socials

DefinitelyMaybe:
Socials are back on mobile docs within the dropdown 🎉 The socials are no longer split up depending on screen size. Previously, larger screens had github at top and twitter + discord on rightsidebar.

Future notes: Even though this has added space to the top nav, will be looking into solutions for displaying packages at the top nav. More packages are on their way 😄


grischaerbe (approved):
LGTM, thanks so much! ❤️

PR #843, opened 2024-01-30
Add prettier check CI workflow

michealparks:
This PR adds a github action workflow that requires `prettier --check` to pass in order to merge a PR.

I've also run the formatter here on all `svelte,ts,md` files.


apps/docs/src/components/Hero/Reveal.svelte
```diff
@@ -1,26 +1,29 @@
...
-	$: p = clamp(mapLinear(progress, from, to, 0, 1), 0, 1)
+  $: p = clamp(mapLinear(progress, from, to, 0, 1), 0, 1)
 </script>
 
-<div class="reveal" style="--progress: {p};">
-	<slot />
+<div
+  class="reveal"
+  style="--progress: {p};"
+>
```
DefinitelyMaybe, line 15: small thing, I like the ending brackets `>` to not be on a line by themselves. though I think some folks like it this way instead 🤷

this way is the default but not sure why

package.json
```diff
@@ -11,7 +11,8 @@
     "install:packages": "pnpm install --filter='!./apps/*'",
     "package": "turbo run package --filter='./packages/*'",
     "package:graph": "turbo run package --filter='./packages/*' --graph",
-    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "format": "prettier --write \"**/*.{ts,svelte,md}\"",
+    "prettier": "prettier --check \"**/*.{ts,svelte,md}\"",
```
DefinitelyMaybe, line 15: hmm it's hit all the svelte files but what about the astro and mdx files?

DefinitelyMaybe (commented):
I'm so onboard with making this part of CI/CD 🚀

package.json
michealparks: I considered leaving that for a follow up since this already resulted in a lot of lint changes 😆

DefinitelyMaybe: fair enough,  must admit when I first opened the PR I was like.... "87 files changes for CI/CD" 😱 ...and then realized it was just the newly formatted files

apps/docs/src/components/Hero/Reveal.svelte
grischaerbe: I like having them on a single line, makes indentation clear.

DefinitelyMaybe: all good

package.json
grischaerbe: I'm also fine with a follow up for that. Formatting in mdx files is pretty tricky and sometimes prettier is just messing up the output, so it might not be desirable after all.

grischaerbe (approved):
Thanks for setting that up and formatting 👍 
So, if I understand correctly, the GH action is running whenever a PR is created and prettier is throwing an error whenever a malformatted file is found, resulting in a missing PR check that is visible right in the PR UI, right? If so, merge approval is granted :)

michealparks:
@grischaerbe Exactly! After this is merged I'll submit a test PR just to ensure that happens, and if there's an issue I can make a follow-up.

PR #844, opened 2024-01-30
Remove `svelte/internal` dependency for suspense

michealparks:
I figured this is one immediate task we could squash to get ahead of the svelte 5 migration. Let me know if this implementation has any issues and I'll be happy to update!

I couldn't detect any reason to continue to generate keys for components in this suspense implementation, and instead opted to instead generate sets of promises within a context. If I'm missing something then let me know!


grischaerbe (approved):
Great stuff @michealparks, I just added a check to do less work in case if there's no suspense context in `useSuspense` and replaced `any` with `unknown`, it's internal state anyway, so should not have a public effect 👍

PR #847, opened 2024-01-30
docs: remove button component

DefinitelyMaybe:
all uses of `<Button />` component had `href` props which means that they ended up as `<a/>` tags but with extra steps 😄 

notes: build fails because of astro config so this PR depends on: #846


DefinitelyMaybe:
argh. so going to redo this. turns out I missed an import in `ShowcaseEntry.astro`. Also thinking I might make a class to hold the common tailwind classes between all the buttons.

All of the commits came in because I tried to rebase the branch thinking it was the astro config 😓

PR #852, opened 2024-02-01
Add port of drei's MeshRefractionMaterial

flo-bit:
Mostly a direct port, with a few exceptions:

- drei splits the material into a ShaderMaterial and a component file, I kept everything in one file (but the fragment and vertex shaders are in their own file, so file length is comparable)
- drei has that interesting CubeCamera component, which threlte doesn't (yet) have, but I added examples of how you can use the material with a THREE.CubeCamera or a loaded texture (sadly just using the environment doesn't work for some reason)

I'm still conflicted regarding the ShaderMaterial that drei has, would make life a tiny bit easier, but not sure if it's really worth it, anyone else has an opinion?

As the documentation grows it may also make sense to split up the components section similar to how drei does it (e.g. a materials part) or sort it in some way.


packages/extras/package.json
```diff
@@ -69,6 +69,7 @@
     "types": "./dist/index.d.ts"
   },
   "dependencies": {
+    "three-mesh-bvh": "^0.7.1",
```
DefinitelyMaybe, line 72: I was wondering about this new dependency but yes, it is part of [drei's component](https://github.com/pmndrs/drei/blob/b03ed79a7aa7584f79fe689022d747d4272749fe/src/materials/MeshRefractionMaterial.tsx#L6). My guess is that's what it takes to make this kind of effect?

DefinitelyMaybe (commented):
docs LGTM 👍 

> As the documentation grows it may also make sense to split up the components section similar to how drei does it (e.g. a materials part) or sort it in some way.

Yes! Currently cleaning up docs to make a move into something more organized easier.

packages/extras/package.json
flo-bit: Ah, probably should have mentioned that in the PR, yeah that what makes the raycasting in the shader work fast enough for real time performance, could work without it, but then it would probably be _very_ slow

DefinitelyMaybe: yeah, I'm a fan of `three-mesh-bvh` 👍 

wanted to point it out for other reviewers

DefinitelyMaybe: Oh, also on reflection, it might be an idea for us to talk about the dependencies of extras components. It would be more work though and we'd have to check/update all our other components too

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,109 @@
...
+        })
+      )
+    }
+  })
+
+  const { camera } = useThrelte()
+
+  useTask(() => {
+    material.uniforms.viewMatrixInverse.value = $camera.matrixWorld
+    material.uniforms.projectionMatrixInverse.value = $camera.projectionMatrixInverse
```
michealparks, line 95:
```suggestion
    material.uniforms.viewMatrixInverse.value = camera.current.matrixWorld
    material.uniforms.projectionMatrixInverse.value = camera.current.projectionMatrixInverse
```
Accessing the `.current` prop may be a little more performant

packages/extras/package.json
michealparks: I'd prefer to depart from how drei does things here and instead opt for peer dependencies. This option would definitely require more documentation and education for setup but would also ultimately result in a better UX by reducing bundle size.

`three-mesh-bvh` is [~70kb](https://bundlephobia.com/package/three-mesh-bvh@0.7.1). It's a pretty commonly used library so if someone is using it and `MeshRefractionMaterial` with this dependency setup they're adding 140kb to their app, which is pretty costly.

If we use peer dependencies then the user would have to be instructed in the docs that installing `three-mesh-bvh` is required, but once they do so then they're only adding ~70kb of cost.

I understand that drei likely does things the way they do for convenience, but I think of the alternative as putting the end user first. Curious if anyone has any thoughts here.

DefinitelyMaybe: haha helping to guide towards a svelte'ier way of doing it 😄

flo-bit: I'm fine with both options, but if going the peer dependency route, the documentation should mention that in multiple places in big bold warning letters so people don't miss it imo.

michealparks: ^ Absolutely, I'd imagine we'd have to create a new component in the docs that very boldly spells out a dependency requirement

grischaerbe: I'm on @michealparks side here. Let's make it a peer dependency and add a `<Tip type="info">` as the first thing in the docs.

apps/docs/src/examples/extras/mesh-refraction-material/App.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  import { Canvas } from '@threlte/core'
+  import Scene from './Scene.svelte'
+</script>
+
+<div class="relative h-full w-full ">
+  <Canvas>
+    <Scene />
+  </Canvas>
+</div>
```
DefinitelyMaybe, line 10: Sorry, I missed this on my first review

```suggestion
<div>
  <Canvas>
    <Scene />
  </Canvas>
</div>
<style>
  div {
    height: 100%;
  }
</style>
```

packages/extras/package.json
DefinitelyMaybe: Nice! sorted. I think we'll be all good with this going forward because even if folks miss the info at the top, the node error should direct them to why/what to do as well.

DefinitelyMaybe (approved):

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte.d.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { ColorRepresentation, CubeTexture, ShaderMaterial, Texture } from 'three'
+
+/*
+  To use this component you need to install the seperate library `three-mesh-bhv`
+	please run `npm install three-mesh-bhv` before adding this component to your project.
+*/
+
+export type MeshRefractionMaterialProps = Props<ShaderMaterial> & {
+  envMap?: CubeTexture | Texture
+  /** Number of ray-cast bounces, it can be expensive to have too many, 2 */
```
michealparks, line 12:
```suggestion
  /**
   * Number of ray-cast bounces, it can be expensive to have too many
   * @default 2
   */
```
^^ JSDoc has a nice `@default` option that I think would make these prop descriptions easier to read

flo-bit: 👍 added

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,109 @@
...
+    // Add defines from chromatic aberration
+    if (aberrationStrength > 0) temp.CHROMATIC_ABERRATIONS = ''
+    if (fastChroma) temp.FAST_CHROMA = ''
+    return temp
+  }
+
+  $: if (envMap && aberrationStrength != null && fastChroma != null) {
+    defines = updateDefines()
+  }
+
```
grischaerbe, line 78:
```suggestion

  const updateDefines = (
    envMap: $$Props['envMap'],
    aberrationStrength: $$Props['aberrationStrength'],
    fastChroma: $$Props['fastChroma']
  ) => {
    const tempDefines = {} as { [key: string]: string }
    // Sampler2D and SamplerCube need different defines
    const isCubeMap = isCubeTexture(envMap)
    const w = (isCubeMap ? envMap.image[0]?.width : envMap.image.width) ?? 1024
    const cubeSize = w / 4
    const lodMax = Math.floor(Math.log2(cubeSize))
    const cubeSize = Math.pow(2, lodMax)
    const width = 3 * Math.max(cubeSize, 16 * 7)
    const height = 4 * cubeSize
    if (isCubeMap) tempDefines.ENVMAP_TYPE_CUBEM = ''
    tempDefines.CUBEUV_TEXEL_WIDTH = `${1.0 / width}`
    tempDefines.CUBEUV_TEXEL_HEIGHT = `${1.0 / height}`
    tempDefines.CUBEUV_MAX_MIP = `${lodMax}.0`
    // Add defines from chromatic aberration
    if (aberrationStrength > 0) tempDefines.CHROMATIC_ABERRATIONS = ''
    if (fastChroma) tempDefines.FAST_CHROMA = ''
    return tempDefines
  }

  $: defines = updateDefines(envMap, aberrationStrength, fastChroma)

```

grischaerbe: You could theoretically even put that function in a module script tag since it's stateless 👍

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,109 @@
...
+      )
+    }
+  })
+
+  const { camera } = useThrelte()
+
+  useTask(() => {
+    material.uniforms.viewMatrixInverse.value = camera.current.matrixWorld
+    material.uniforms.projectionMatrixInverse.value = camera.current.projectionMatrixInverse
+  })
```
grischaerbe, line 98:
```suggestion
  useTask(() => {
    material.uniforms.viewMatrixInverse.value = camera.current.matrixWorld
    material.uniforms.projectionMatrixInverse.value = camera.current.projectionMatrixInverse
  }, { autoInvalidate: false })
```

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,111 @@
...
+  is={material}
+  uniforms.envMap.value={envMap}
+  uniforms.bounces.value={bounces}
+  uniforms.ior.value={ior}
+  uniforms.fresnel.value={fresnel}
+  uniforms.aberrationStrength.value={aberrationStrength}
+  uniforms.color.value={new Color(color)}
+  uniforms.resolution.value={[$size.width, $size.height]}
+  {defines}
+/>
```
grischaerbe, line 111:
```suggestion
  useTask(() => {
    material.uniforms.viewMatrixInverse.value = camera.current.matrixWorld
    material.uniforms.projectionMatrixInverse.value = camera.current.projectionMatrixInverse
  }, { autoInvalidate: false })
  
  const colorObj = new Color(color)
  $: {
    colorObj.set(color)
    invalidate()
  }
</script>

<T
  is={material}
  uniforms.envMap.value={envMap}
  uniforms.bounces.value={bounces}
  uniforms.ior.value={ior}
  uniforms.fresnel.value={fresnel}
  uniforms.aberrationStrength.value={aberrationStrength}
  uniforms.color.value={colorObj}
  uniforms.resolution.value={[$size.width, $size.height]}
  {defines}
/>
```

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,117 @@
...
+      fresnel: { value: 0 },
+      bvh: { value: new MeshBVHUniformStruct() },
+      color: { value: new Color('white') },
+      resolution: { value: new Vector2() },
+      viewMatrixInverse: { value: new Matrix4() },
+      projectionMatrixInverse: { value: new Matrix4() }
+    }
+  })
+
+  const { size } = useThrelte()
```
grischaerbe, line 47:
```suggestion
  const { size, invalidate } = useThrelte()
```

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,117 @@
...
+      fresnel: { value: 0 },
+      bvh: { value: new MeshBVHUniformStruct() },
+      color: { value: new Color('white') },
+      resolution: { value: new Vector2() },
+      viewMatrixInverse: { value: new Matrix4() },
+      projectionMatrixInverse: { value: new Matrix4() }
+    }
+  })
+
+  const { size, invalidate } = useThrelte()
```
grischaerbe, line 47:
```suggestion
  const { size, invalidate, camera } = useThrelte()
```

packages/extras/src/lib/components/MeshRefractionMaterial/MeshRefractionMaterial.svelte
```diff
@@ -0,0 +1,117 @@
...
+      material.uniforms.bvh.value = new MeshBVHUniformStruct()
+      material.uniforms.bvh.value.updateFrom(
+        new MeshBVH($parent?.geometry.clone().toNonIndexed(), {
+          strategy: SAH
+        })
+      )
+    }
+  })
+
+  const { camera } = useThrelte()
```
grischaerbe, line 93:
```suggestion
```

grischaerbe:
Thanks a lot! I hope to be able to see it in the future :face_holding_back_tears:

PR #855, opened 2024-02-04
Remove tailwind classes from examples

flo-bit:
Removed most tailwind classes from the examples.

Only example I found that I left as-is for now that still has tailwind classes is extras/transitions, as that uses a few more classes for a custom button, which maybe should be switched to tweakpane instead?

### DIFF
--- a/.changeset/hip-lemons-invite.md
+++ b/.changeset/hip-lemons-invite.md
@@ -1,0 +1,5 @@
+---
+"@threlte/docs": minor
+---
+
+Remove tailwind classes from examples
--- a/.changeset/hip-lemons-invite.md
+++ b/.changeset/hip-lemons-invite.md
@@ -1,5 +1,0 @@
----
-'@threlte/docs': minor
----
-
-Remove tailwind classes from examples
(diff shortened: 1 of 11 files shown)


DefinitelyMaybe (approved):
Love this! ❤️ 

I don't think the docs needs to worry about changeset tho, lemme double check with team, I could be wrong 😅

michealparks:
No changset needed for docs since it's not versioned / shipped to NPM 😁

michealparks:
One last note: we're soon going to enforce prettier for all PRs to pass, @flo-bit csn you run "pnpm run format" on all your open PRs at the root level of the monorepo?

michealparks:
@flo-bit I've just added prettier to our CI checks. You'll have to do the following to clear up issues, let me know if you have problems!

* Merge latest main into your branches
* Run `pnpm i` to upgrade prettier
* Run `pnpm format` from the root `threlte` repo
* Prettier check should be green after push to origin 🎉

flo-bit:
> @flo-bit I've just added prettier to our CI checks. You'll have to do the following to clear up issues, let me know if you have problems!
> 
> * Merge latest main into your branches
> * Run `pnpm i` to upgrade prettier
> * Run `pnpm format` from the root `threlte` repo
> * Prettier check should be green after push to origin 🎉

All good from my side 👍

PR #857, opened 2024-02-06
Interactivity plugin cleanup and optimizations

michealparks:
I've been wanting to clean up the interactivity after modeling the XR repo's pointer controls after it, and I was able to get to it this week!

I've done a few simple cleanup / optimization tasks here:

* Organize contexts and use symbols for keys
* Use a weakmap instead of userData for event dispatching
* Other small cleanup and optimization that I'll list in comments

### DIFF
--- a/packages/extras/src/lib/interactivity/context.ts
+++ b/packages/extras/src/lib/interactivity/context.ts
@@ -6,8 +6,15 @@ import {
 } from '@threlte/core'
 import { getContext, setContext } from 'svelte'
 import { Vector2, Raycaster } from 'three'
-import type { IntersectionEvent, ThrelteEvents } from './types'
+import type { IntersectionEvent, ThrelteEvents, DomEvent } from './types'
 import { getDefaultComputeFunction } from './defaults'
+
+export type FilterFunction = (
+  items: THREE.Intersection[],
+  context: InteractivityContext
+) => THREE.Intersection[]
+
+export type ComputeFunction = (event: DomEvent, context: InteractivityContext) => void
 
 export type InteractivityOptions = {
   enabled?: boolean
@@ -25,15 +32,6 @@ export type InteractivityOptions = {
    */
   filter?: FilterFunction
 }
-
-export type DomEvent = PointerEvent | MouseEvent | WheelEvent
-
-export type FilterFunction = (
-  items: THREE.Intersection[],
-  context: InteractivityContext
-) => THREE.Intersection[]
-
-export type ComputeFunction = (event: DomEvent, context: InteractivityContext) => void
 
 export type InteractivityContext = {
   enabled: CurrentWritable<boolean>
--- a/packages/extras/src/lib/interactivity/context.ts
+++ b/packages/extras/src/lib/interactivity/context.ts
@@ -1,4 +1,9 @@
-import { currentWritable, type CurrentWritable, type createRawEventDispatcher, useThrelte } from '@threlte/core'
+import {
+  currentWritable,
+  type CurrentWritable,
+  type createRawEventDispatcher,
+  useThrelte
+} from '@threlte/core'
 import { getContext, setContext } from 'svelte'
 import { Vector2, Raycaster } from 'three'
 import type { IntersectionEvent, ThrelteEvents } from './types'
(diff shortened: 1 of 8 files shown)


packages/extras/src/lib/interactivity/context.ts
```diff
@@ -0,0 +1,97 @@
+import {
```
michealparks, line 1: Moved all context-related logic here

packages/extras/src/lib/interactivity/context.ts
```diff
@@ -0,0 +1,94 @@
...
+  setContext<InteractivityContext>(contextKey, context)
+
+  return context
+}
+
+interface HandlerContext {
+  dispatchers: WeakMap<THREE.Object3D, ReturnType<typeof createRawEventDispatcher<ThrelteEvents>>>
+}
+
+const handlerContextKey = Symbol('interactivity-handler-context')
```
michealparks, line 84: Created a new context for dispatchers

packages/extras/src/lib/interactivity/defaults.ts
```diff
@@ -1,8 +1,10 @@
-import { memoize, useThrelte, watch } from '@threlte/core'
-import type { ComputeFunction, State } from './types'
+import { useThrelte, watch, type CurrentWritable } from '@threlte/core'
+import type { ComputeFunction } from './context'
 
-export const getDefaultComputeFunction = (state: State): ComputeFunction => {
-  const camera = memoize(useThrelte().camera)
+export const getDefaultComputeFunction = (
+  target: CurrentWritable<HTMLElement>
```
michealparks, line 5: This function takes only the target store to simplify context creation

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -61,60 +58,46 @@ export const setupInteractivity = (state: State) => {
...
-      // Filter out duplicates
-      .filter((item) => {
-        const id = getIntersectionId(item as Intersection)
-        if (duplicates.has(id)) return false
-        duplicates.add(id)
-        return true
-      })
-
-    if (state.filter) hits = state.filter(hits, state)
+    const hits = context.raycaster.intersectObjects(context.interactiveObjects, true)
```
michealparks, line 67: The `intersectObjects` method sorts by distance and contains no duplicates as part of its behavior, allowing us to simplify `getHits()`

DefinitelyMaybe: [can confirm](https://threejs.org/docs/index.html?q=ray#api/en/core/Raycaster.intersectObjects), [source link too](https://github.com/mrdoob/three.js/blob/841d2e791d3e8a2463322c5ca31b16956828b91c/src/core/Raycaster.js#L66)

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -1,15 +1,10 @@
...
-  return object.userData._threlte_interactivity_dispatcher as
-    | ReturnType<typeof createRawEventDispatcher<ThrelteEvents>>
-    | undefined
-}
+import { getHandlerContext, type InteractivityContext } from './context'
+import type { DomEvent, Intersection, IntersectionEvent } from './types'
 
 function getIntersectionId(event: Intersection) {
-  return (event.eventObject || event.object).uuid + '/' + event.index + event.instanceId
+  return `${(event.eventObject || event.object).uuid}/${event.index}${event.instanceId}`
```
DefinitelyMaybe, line 7: nice! this looks clearer to me as to what's going on here

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -61,60 +58,46 @@ export const setupInteractivity = (state: State) => {
     }
   }
 
-  const enabled = memoize(state.enabled)
+  const enabled = memoize(context.enabled)
 
   const getHits = (): Intersection[] => {
-    const duplicates = new Set<string>()
+    if (!enabled) return []
```
grischaerbe, line 64:
```suggestion
    if (!enabled.current) return []
```

grischaerbe (changes requested):
I have a tiny change, apart from that it looks fabulous!

packages/extras/src/lib/interactivity/setupInteractivity.ts
michealparks: Ah! Good catch. Should have gone with an explicit comparison 🤦

PR #858, opened 2024-02-06
Use `watch()` in useEvents

michealparks:
While investigating how to upgrade `@threlte/core`'s event dispatching to svelte 5 I ran into a store subscription that could be simplified with `watch()`.

This PR simplifies the subscriptions made in the `useEvents` hook!

### DIFF
--- a/packages/core/src/app.d.ts
+++ b/packages/core/src/app.d.ts
@@ -9,7 +9,3 @@ declare namespace App {
   // interface Session {}
   // interface Stuff {}
 }
-
-declare module '*.svelte' {
-  export { SvelteComponentDev as default } from 'svelte/internal'
-}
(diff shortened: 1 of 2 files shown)


packages/core/src/app.d.ts
```diff
@@ -9,7 +9,3 @@ declare namespace App {
   // interface Session {}
   // interface Stuff {}
 }
-
-declare module '*.svelte' {
```
michealparks, line 13: As far as I can tell, this isn't needed for typechecking. I might be missing context about why it was added, but if we can remove it then migration to svelte 5 will be just a little simpler 🎉

grischaerbe: I think this was before the types around `SvelteComponent` stabilized, so good catch 👍

grischaerbe (approved):
LGTM 👍

PR #859, opened 2024-02-06
Text3D component

flo-bit:
Wrapper around three's TextGeometry with included font loading.

There is one issue I haven't been able to fix, which is that wrapping it into an `<Align>` component doesn't seem to work when the font is loaded by the component (if you load it yourself it works). 

I'm guessing the align happens before the font is loaded and the bounding box is recalculated. Any tips for fixing that?

Drei's version works by traversing all children and adding those that are of type `string` or `number` but from what I know that's not possible in svelte? I added a text prop to the component instead.

Also drei's version adds three props `lineHeight`, `letterSpacing` and `smooth`, I haven't added those yet, what's the opinion there?


grischaerbe:
Thanks @flo-bit, great stuff ❤️ 

- If you add an event to the `<Text3D>` component that fires when the font has been loaded (and the mesh generated), you can re-align, see the example in the docs: https://threlte.xyz/docs/reference/extras/align#slot-props.
- `text` prop is fine, Svelte can't access `children` like react can.
- Yes to more options! :)

grischaerbe:
Also: `<Align>` should theoretically re-align by itself when you use the `auto` prop:

```svelte
<Align auto>
  <Text3D text="hello there" />
</Align>
```

flo-bit:
> * If you add an event to the `<Text3D>` component that fires when the font has been loaded (and the mesh generated), you can re-align

Added the event, just called it rendered and passing in the mesh for now (see ["centering text" part in the docs](https://threlte-gpvjwovo7-threlte.vercel.app/docs/reference/extras/text-3d)) as I couldn't find an example for what events like this should be named/should pass as parameters.

> Also: `<Align>` should theoretically re-align by itself when you use the auto prop:

`<Align auto>` is not working for me here for some reason ¯\\\_(ツ)\_/¯

> * Yes to more options! :)

I added `smooth` as a parameter reusing the toCreasedNormals function from RoundedBoxGeometry (maybe move that to the lib folder?)

Could not for the life of me figure out how drei does `lineHeight` and `letterSpacing`, they just seem to pass the prop on to TextGeometry but that doesn't actually accept those parameters? Any pointers much appreciated ([Text3D source from drei](https://github.com/pmndrs/drei/blob/503f4f7b9a715dfdd33b9d0a07e6f08b93894903/src/core/Text3D.tsx))

There also seems to also be a problem with changelog and prettier, I could commit all changed files when I run `pnpm format` (`Changelog.md`, `tsconfig.json` and `app.html` in packages/extras) but I didn't want to clutter the commit with file changes that don't really belong.

DefinitelyMaybe:
> maybe move that to the lib folder?

Inlining ain't so bad. Sometimes functions diverge and then you need to add conditionals. Admittedly, either way can get hairy

packages/extras/src/lib/components/Text3D/Text3D.svelte
```diff
@@ -0,0 +1,83 @@
...
+    T,
+    forwardEventHandlers,
+    useLoader,
+    type AsyncWritable,
+    asyncWritable
+  } from '@threlte/core'
+  import { createEventDispatcher } from 'svelte'
+
+  import type { Mesh } from 'three'
+  import { TextGeometry, FontLoader, Font } from 'three/examples/jsm/Addons.js'
```
michealparks, line 12: We should use individual import paths here due to this issue: https://github.com/threlte/threlte/pull/871

apps/docs/src/content/reference/extras/text-3d.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### Loading a font yourself
+
+You can also load the font yourself and pass it to the component, like so:
+
+```svelte title="Text3D.svelte"
+<script lang="ts">
+	import { T, useLoader } from '@threlte/core'
+	import { Text3D } from '@threlte/extras'
+	import { FontLoader } from 'three/examples/jsm/Addons.js'
```
michealparks, line 61: Same note concerning https://github.com/threlte/threlte/pull/871

michealparks:
>I added smooth as a parameter reusing the toCreasedNormals function from RoundedBoxGeometry (maybe move that to the lib folder?)

Back when I added the toCreasedNormals it was a copy of a the three/examples function, but without an unrelated `console.warn` that would spam the user every time they created a `RoundedBoxGeometry`. Since then Three.js has fixed / removed that warning, so I'd recommend we delete the file in Threlte and just have both of these components use the Three.js version!

```ts
import { toCreasedNormals } from `three/examples/jsm/utils/BufferGeometryUtils`
```

michealparks:
> There also seems to also be a problem with changelog and prettier, I could commit all changed files when I run pnpm format (Changelog.md, tsconfig.json and app.html in packages/extras) but I didn't want to clutter the commit with file changes that don't really belong.

I've updated prettier to not format autogenerated files like Changelog.md to reduce confusion. If you merge / rebase main the problem should go away.

flo-bit:
> >I added smooth as a parameter reusing the toCreasedNormals function from RoundedBoxGeometry (maybe move that to the lib folder?)
> 
> 
> 
> Back when I added the toCreasedNormals it was a copy of a the three/examples function, but without an unrelated `console.warn` that would spam the user every time they created a `RoundedBoxGeometry`. Since then Three.js has fixed / removed that warning, so I'd recommend we delete the file in Threlte and just have both of these components use the Three.js version!
> 
> 
> 
> ```ts
> 
> import { toCreasedNormals } from `three/examples/jsm/utils/BufferGeometryUtils`
> 
> ```
> 
> 

Sounds good to me, do you want me to add that change of RoundedBoxGeometry to this PR though or open a new one?

michealparks:
Should be fine to change in this PR! Also the correct import should be `three/examples/jsm` instead of three/addons` (like in the case of `Font`) because some bundlers don't recognize that import. We also don't want to include file extensions in imports for consistency.

grischaerbe:
Hey @flo-bit, are you okay with this to be a geometry rather than a mesh? So the markup would look like this:

```svelte
<T.Mesh>
  <Text3DGeometry />
  <T.MeshStandardMaterial />
</T.Mesh>
```

The reason for this is that by being a geometry, it would also support instancing, usage in a `THREE.BatchedMesh` or usage in `THREE.Points`. The API can stay like this and I think features like font loading and beveling still make it a great fit for extras and would align with `<RoundedBoxGeometry>`. 👍

flo-bit:
> Hey @flo-bit, are you okay with this to be a geometry rather than a mesh? So the markup would look like this:
> 
> ```svelte
> <T.Mesh>
>   <Text3DGeometry />
>   <T.MeshStandardMaterial />
> </T.Mesh>
> ```
> 
> The reason for this is that by being a geometry, it would also support instancing, usage in a `THREE.BatchedMesh` or usage in `THREE.Points`. The API can stay like this and I think features like font loading and beveling still make it a great fit for extras and would align with `<RoundedBoxGeometry>`. 👍

Sure, sounds good to me, I'll change that.

PR #866, opened 2024-02-09
Remove children binding from HierarchicalObject component

michealparks:
In Svelte 5 `children` is a reserved property for snippets. When testing Svelte 5 integration with Threlte one thing that breaks is the HierarchicalObject component because of the `children` binding.

I propose that we remove it to prepare for Svelte 5, since HierarchicalObject isn't used in the community often, and it seems like there's little documentation behind this binding.

I may be missing context about this binding, so I'd love to hear if there's a problem with this!

### DIFF
--- a/.changeset/good-items-add.md
+++ b/.changeset/good-items-add.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": minor
+---
+
+Remove children binding from HierarchicalObject component
--- a/apps/docs/src/content/reference/core/hierarchical-object.mdx
+++ b/apps/docs/src/content/reference/core/hierarchical-object.mdx
@@ -52,17 +52,6 @@ In this example the component is responsible for
 - adding children to the scene graph (i.e. as children to the mesh)
 - removing children from the scene graph
 
-```svelte
-@@ -86,23 +117,41 @@ The component `<HierarchicalObject>` also provides a useful binding to collect a
-
-<HierarchicalObject
-  object={mesh}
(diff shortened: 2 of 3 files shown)


DefinitelyMaybe:
I'd go for a rename or adding a prefix. [The portal component](https://threlte.xyz/docs/reference/extras/portal) uses it.

I think this is an opportunity to step back and think about the wider picture. How are T components added to the scene graph and how do we let user's adjust that when needed

Should it be sorted within each T component? We have the attach prop which kinda does this

michealparks:
@DefinitelyMaybe the only thing removed here is the `children` binding:

```html
<HierarchicalObject bind:children />
```

No components within the monorepo use this binding, and from a quick search I don't see any public community repos using it either: https://github.com/search?q=path%3A*.svelte+bind%3Achildren&type=code

I'd argue it's a chance to remove something largely unused.

All other features of the component are kept, which should keep `<Portal />` working as usual.

Agreed that it would be really cool to take a bigger look at the scene graph features Threlte provides, but I'd like to keep this PR focused on making the migration to svelte 5 easier 😄

DefinitelyMaybe:
Oooh thank you. That makes much more sense. Don't mind me 😅

grischaerbe (approved):
Actually, I'd love to refactor the whole hierarchy handling, and this is a great step forwards, making it easier to replace the implementation while keeping the API 👍 

LGTM 🚀

PR #874, opened 2024-02-12
Replace all instances of SvelteComponentTyped

michealparks:
This is just a small maintenance PR to prepare for Svelte 5. Here I've replaced all instances of the deprecated `SvelteComponentTyped` type with `SvelteComponent`. I didn't include a changeset since this was just internal type maintenance.


grischaerbe (approved):
LGTM 🚀

PR #875, opened 2024-02-12
Next: docs: clean up Hero folder

DefinitelyMaybe:
* removed unused imports/exports
* unused `let:Sync`
* HeroWrapper flagged `top-[66svh]` vs `top-[66vh]` with `'top-[66svh]' applies the same CSS properties as 'top-[66vh]'.` ... I know svh and vh are different, is it a worry here?


apps/docs/src/components/Hero/HeroWrapper.svelte
```diff
@@ -164,9 +164,7 @@
       in={2.7}
       out={3.9}
     >
-      <h2
-        class="fixed left-0 top-[66svh] top-[66vh] flex w-screen flex-col items-center justify-center"
-      >
+      <h2 class="fixed left-0 top-[66vh] flex w-screen flex-col items-center justify-center">
```
grischaerbe, line 167: The Tailwind extension is falsely complaining about it. If a browser doesn't know [dynamic viewport units](https://caniuse.com/viewport-unit-variants), it will fall back to regular units. Since dynamic viewport units have good support nowadays, we can probably reside to only use svh, but I wouldn't drop it, otherwise scrolling on an iPhone might result in jumping text (because of the appearing/disappearing status bar).

grischaerbe (changes requested):
I left a small note, otherwise LGTM :)

apps/docs/src/components/Hero/HeroWrapper.svelte
DefinitelyMaybe: good points! thank you

Issue #804, opened 2023-12-21
Crash when deallocating canvas with <Sky /> element


flo-bit:
[See a minimal example here](https://stackblitz.com/edit/nkthtn-wqdgqa?file=src%2Fexample%2FApp.svelte) toggle sky on, then toggle canvas off -> crash

Toggling the canvas/scene on or off without a sky component works without problems, when toggling the canvas off while it has a `<Sky />` I get the following runtime error:

```
TypeError: Cannot read properties of undefined (reading '0')
    at deallocateRenderTarget (chunk-EL5WXWEC.js?v=7bfaad47:15318:68)
    at WebGLCubeRenderTarget.onRenderTargetDispose (chunk-EL5WXWEC.js?v=7bfaad47:15277:5)
    at WebGLCubeRenderTarget.dispatchEvent (chunk-EL5WXWEC.js?v=7bfaad47:709:18)
    at WebGLCubeRenderTarget.dispose (chunk-EL5WXWEC.js?v=7bfaad47:2352:10)
    at @threlte_extras.js?v=7bfaad47:20042:50
    at run (chunk-7JRVFVIN.js?v=7bfaad47:28:10)
    at Array.forEach (<anonymous>)
    at run_all (chunk-7JRVFVIN.js?v=7bfaad47:34:7)
    at destroy_component (chunk-7JRVFVIN.js?v=7bfaad47:2148:5)
    at Object.destroy [as d] (Scene.svelte:5:39)
```

MikeBellika:
I get the same crash report, except it happens whenever the app hot reloads

grischaerbe:
fixed in #877 🎉

Issue #876, opened 2024-02-12
Using Sky causes exception if the user leaves the web page


clibequilibrium:
Using sky in this code raises exception when user navigates back or forward from 3D web page

```svelte
<script>
	import { T, useTask } from '@threlte/core';
	import { OrbitControls, Sky, interactivity } from '@threlte/extras';
	import { spring } from 'svelte/motion';
	import { DEG2RAD } from 'three/src/math/MathUtils.js';
	interactivity();
	const scale = spring(1);
	let rotation = 0;
	useTask((delta) => {
		rotation += delta;
	});
</script>

<Sky elevation={0.5} />

<T.PerspectiveCamera position={[0, 7, 18]} fov={60} near={1} far={20000} makeDefault>
	<OrbitControls
		enableZoom={true}
		maxPolarAngle={85 * DEG2RAD}
		enableDamping
		target={[0, 2.5, 0]}
	/>
</T.PerspectiveCamera>

<T.DirectionalLight position={[0, 10, 10]} castShadow />
<T.Mesh
	rotation.y={rotation}
	position.y={1}
	scale={$scale}
	on:pointerenter={() => scale.set(1.5)}
	on:pointerleave={() => scale.set(1)}
	castShadow
>
	<T.BoxGeometry args={[1, 2, 1]} />
	<T.MeshStandardMaterial color="hotpink" />
</T.Mesh>
<T.Mesh rotation.x={-Math.PI / 2} receiveShadow>
	<T.CircleGeometry args={[4, 40]} />
	<T.MeshStandardMaterial color="white" />
</T.Mesh>

```

![image](https://github.com/threlte/threlte/assets/105135724/60e93f8d-5b17-4622-9dfd-f58be4efee07)

at ``deallocateRenderTarget()``

The user is unable to get back to the 3D scene unless they reload the web page to see 3D scene again. Commenting out ``<Sky>`` fixes the issue.

clibequilibrium:
Possible related to https://github.com/threlte/threlte/issues/804

grischaerbe:
fixed in #877 🎉

clibequilibrium:
> fixed in #877 🎉

ty so much !

PR #877, opened 2024-02-12
`<Sky>` fix

grischaerbe:
Fixed a bug where `<Sky>`'s renderTarget wasn't able to be disposed because the canvas context did not exist anymore.

### DIFF
--- a/.changeset/cyan-elephants-jog.md
+++ b/.changeset/cyan-elephants-jog.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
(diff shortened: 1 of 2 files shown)


grischaerbe:
fixes #804  and #876

DefinitelyMaybe:
TIL js optional chain will still throw

grischaerbe:
It's not the check for `undefined`, as it in fact isn't `undefined`, but because the canvas context is already destroyed, something fails while calling `renderTarget.dispose()` :/

PR #879, opened 2024-02-13
Refactor T component, add optimizations

michealparks:
More Svelte 5 prep in this PR. Here I've done a small refactor on the T component by pulling out most logic, so that if we end up temporarily supporting both a Svelte 4 and 5 component, maintenance will be easier.

I also included a few tiny optimizations! 🎉 Details in comments.

### DIFF
--- a/.changeset/pretty-pears-lay.md
+++ b/.changeset/pretty-pears-lay.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": patch
+---
+
+Small `<T>` optimizations


packages/core/src/lib/components/T/utils/utils.ts
```diff
@@ -0,0 +1,38 @@
+import type { Object3D } from 'three'
+import type { AnyClass, MaybeInstance } from '../types'
+import type { DisposableThreeObject } from '../../../types'
+
+const classRegex = /^\s*class\s+/
+
+// Type Guards
+const isClass = (input: unknown): input is AnyClass => {
+  if (typeof input !== 'function') {
```
michealparks, line 9: optimization: return early to avoid the string conversion and regex test

packages/core/src/lib/components/T/utils/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    return false
+  }
+
+  return classRegex.test(input.toString())
```
michealparks, line 13: optimization: cache the regex object. on my machine this showed around a 10x speed improvement for this line!

packages/core/src/lib/components/T/utils/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const argsIsConstructorParameters = (args: unknown): args is ConstructorParameters<AnyClass> => {
+  return Array.isArray(args)
+}
+
+export const determineRef = <Type>(is: unknown, args: unknown): MaybeInstance<Type> => {
+  if (isClass(is)) {
```
michealparks, line 21: optimization: avoid calling `isClass` twice if no args are present

packages/core/src/lib/hooks/useParent.ts
```diff
@@ -1,9 +1,22 @@
...
-  return getContext<ThrelteParentContext>('threlte-hierarchical-parent-context')
+  return getContext<ThrelteParentContext>(key)
+}
+
+export const setParent = (context: ThrelteParentContext) => {
+  return setContext(key, context)
+}
+
+export const createParentContext = (ref: Object3D | undefined) => {
+  const context = createObjectStore(ref)
```
michealparks, line 19: This is new and worth calling out: before the T component was just passing a writable to this context, but  `createObjectStore`, but `HierarchicalObject` was passing `createObjectStore(ref)`.

I fixed this inconsistency since it seems like what we would want, but I'm curious if there's anything I'm potentially missing.

packages/core/src/lib/components/T/utils/utils.ts
grischaerbe: Oh yeah!

grischaerbe (approved):
Thanks! Looks good to me :) We could at some point decide to not have any subcomponents in `<T>`, I think that would make things faster again. Svelte 5 will probably also speed things up, so good times ahead!

packages/core/src/lib/hooks/useParent.ts
grischaerbe: Afaik, the objectStore is just a store that doesn't emit updates when the uuid of a newly set object doesn't change, so all good 👍

PR #881, opened 2024-02-15
create threlte package clean up

DefinitelyMaybe:
To dev or not to dev that is the question


grischaerbe (approved):
Awesome! This alignes create-threlte with the install instructions: https://threlte.xyz/docs/learn/getting-started/installation

LGTM :rocket:

PR #885, opened 2024-02-25
docs: update example and remove useTweakpane hook

DefinitelyMaybe:
Last step before removing the hook would be updating the stackoverflow files/scripts which use it.


grischaerbe (approved):
Nice! LGTM 🚀

PR #891, opened 2024-02-27
Refactor import fix

DefinitelyMaybe:
bum! checked out build output. looks like it has to be `../index`. tried `$lib` first which the ts output was good but building docs failed because of it.


grischaerbe:
Fixed in 36f8eef217c516bf68fc39e91644e8e69b94f685

PR #895, opened 2024-02-27
Docs: Fixes typo

luroc:
Description of Alpha map stated "black is invisible and black is visible". Commit fixes this to "white is visible"


jerzakm:
Good catch, thanks!

jerzakm (approved):

PR #896, opened 2024-02-28
Next: Docs: clean up some of the styles/class usage

DefinitelyMaybe:
I didn't know [class:list](https://docs.astro.build/en/reference/directives-reference/#classlist) was a thing 😄


grischaerbe (changes requested):
I agree, we should probably use the Astro internally provided function 👍 

Can you remove the imports for `c`? Then we're good to merge :)

DefinitelyMaybe:
Not quite that simple.

There's this function call
```js
const getClasses = (): string => {}
```
within the `Tip.astro` file that gets called within a `c('...', getClasses())`.

I'll need to unpack it and figure out a solution, and for the moment this was the low hanging fruit.

Theres also
```js
import { c } from '$lib/classes'
```
within svelte files, and there it's either the templating strings or `c()` stays

grischaerbe:
Ah, I meant to just remove the imports from the Astro files where it's not needed anymore 👍

DefinitelyMaybe:
oh haha, so did a quick vscode search and it turns out `Tip.astro` is the last c import in astro file.

Happy to give it a shot and come back to you, otherwise merge as is 😄

DefinitelyMaybe:
figured it out for `Tip.astro`, made use of template strings `'${}'`

Also adjust `RightSidebar.svelte`, I've been consistently running into the error below because of preprocessing.

```bash
09:45:13 [ERROR] Error while preprocessing C:/Users/blah/Documents/threlte/apps/docs/src/components/Menu/RightSidebar/RightSidebar.svelte - C:/Users/blah/Documents/threlte/apps/docs/src/components/Menu/RightSidebar/RightSidebar.svelte:3:5: The `text-faded` class does not exist. If `text-faded` is a custom class, make sure it is defined within a `@layer` directive.
```

grischaerbe (approved):
Awesome, LGTM 🚀

PR #897, opened 2024-02-29
Fixes for Three.js r162 compatibility

michealparks:
This PR fixes compatibility with the latest three revision by using a wildcard import where missing imports exist.

~~It also updates all repos within the monorepo to use r162.~~

Edit: I reverted updating three in the docs. This breaks the homepage animation for some reason.


packages/rapier/package.json
```diff
@@ -36,6 +36,7 @@
     "svelte-check": "^3.6.3",
     "svelte-preprocess": "^5.1.3",
     "svelte2tsx": "^0.7.0",
+    "three": "^0.162.0",
```
DefinitelyMaybe, line 39: Did the rapier package not need three.js?

michealparks: This is just for the local dev environment (running the svelte kit app locally as a testbed). Doesn't affect anything else!

michealparks: This was previously missing, but should be somewhat pinned like it is in the other repos.

PR #898, opened 2024-03-02
Remove encoding-related code to allow compatibility with r162

michealparks:
This PR removes encoding-related code and bumps the required Three.js version to greater than or equal to r152.

I also fixed the shader deprecation warning for grid.


grischaerbe (approved):
Awesome, thanks so much. I think strictly speaking a dependency update of Three.js is a breaking change, since Three.js' update is also breaking, but I think we can agree that it would be a good practice to update from r13X to 15X.

So, LGTM 🚀

PR #901, opened 2024-03-03
Examples import fix

### LINKED ISSUE #883: Type inference for `useGltf` always returns `any`
I'm trying to use the `useGltf` hook from `@threlte/extras` to load a model. I specified the GLTF scene graph objects but the `AsyncWritable` is always of type `any` and I can't get autocomplete working in VSCode. 

Screenshot 2024-02-19 190447

```ts
  type GLTFResult = {
    nodes: {
      Suzanne: THREE.Mesh
    }
    materials: {}
  }

  const gltf = useGltf<GLTFResult>('/path/to/model.glb', { useDraco: true })
```

Thanks in advance!

grischaerbe:
Added file extensions to all imports that are not part of the usual `'three'` import, I grouped the commits by packages and apps for easier rollback and review. This should also solve issues around `any` types (fixes #883).


apps/docs/src/examples/getting-started/Scene.svelte
```diff
@@ -12,7 +12,7 @@
     MeshStandardMaterial,
     PerspectiveCamera
   } from 'three'
-  import { degToRad } from 'three/src/math/MathUtils'
+  import { degToRad } from 'three/src/math/MathUtils.js'
```
DefinitelyMaybe, line 15: For anyone else wanting to double check the [exports](https://github.com/mrdoob/three.js/blob/c2593ed3db121b17590068c638d5dc115e7496f9/src/math/MathUtils.js#L337)

yes, both `degToRad` and `DEG2RAD` are exports 🤷‍♂️

grischaerbe: Oh, this is not a suggestion, nevermind then :)

DefinitelyMaybe: regex is good 👍

PR #880, opened 2024-02-13
Docs: clean up Astro component scripts

DefinitelyMaybe:
This PR brings our Astro component scripts inline with [how astro does them](https://docs.astro.build/en/basics/astro-components/#the-component-script).

copied patterns were repeated. Anyone know where the `{ }` pattern came from?

```html
---
{
    'category': '@threlte/extras'
    'name': '<Align>'
}
---
```

vs

```html
---
category: '@threlte/extras'
name: '<Align>'
---
```

This PR will also make a move to astro+starlight easier. I ran into a wall of errors because of it when trying to make it happen myself.

PR #911, opened 2024-03-16
Add `<ImageMaterial>` component

michealparks:
Untitled


apps/docs/src/content/reference/extras/image.mdx
```diff
@@ -0,0 +1,56 @@
+---
+{
```
DefinitelyMaybe, line 2: I would like us to not continue the `{ }` brackets within our components. It was giving me troubles when I was mucking around with starlight

michealparks: Updated! The non-brackets version is what astro recommends, correct? If so, is there any progress on moving all of our docs to using that format? I'm happy to help in any way I can to help move that along so that we're using the correct format everywhere.

DefinitelyMaybe: Yeah, I couldn't find the brackets version in Astro docs when I looked into this. Almost there on the progress, have a look over #880 when you get some time

DefinitelyMaybe (approved):

ixxie:
Awesome work, and I like your taste in art. The Threatened Swan is in the Rijks Museum in Amsterdam, and when I was visiting there during high school excursion I chose that painting to write my essay about.

grischaerbe:
@michealparks Would it make sense to make this a material instead of a mesh?

Issue #708, opened 2023-10-28
Better navigation for extras sidebar


jerzakm:
This is madness.

![image](https://github.com/threlte/threlte/assets/16734228/e6b13464-eeef-4575-9cf5-737ec25d9ac0)

Any suggestions on how to improve it?

1. /extras starting/overview page that is a grid with tiny screenshots and component names?

2. Sub categories? For this I currently see these as possibilities:
- Audio
- Camera & controls
- Shadows
- Scene/Environment(?) (Environment, Sky, Grid)
- Utils (Portal, Suspense)
- Others - everything else would go to others, and then once a certain category grows we make a new one and move these out.

michealparks:
Catagories + alpha sorting within categories would make sense to me!

ixxie:
I agree, categories and alphabetical sorting is good. I think it would also let us see opportunities for eventually breaking up extras into distinct packages if it gets too big.

grischaerbe:
Done, I guess?

PR #912, opened 2024-03-17
Added categories to extras sidebar

edsunman:
I've split the extras components into categories based on the drei categories and ordered alphabetically.

Addresses this issue:
https://github.com/threlte/threlte/issues/708


ixxie:
@edsunman thank you for separating this into a distinct PR! I'm gonna have a look at it at some point this week (its been a busy one!).

ixxie:
I generally really like the direction you've taken here, but there are some specific issues I'd like to address.

The main issues is a mix of domain- and type-based categories. I think we should get rid of *hooks*, *lifecycle* and *plugins* as distinct categories: their contents should be distributed in the other categories. If necessary, new categories can be added, fitting the domain orientated approach the other categories have. As a concrete example of benefits, I would like to see all suspense related stuff together, and some of the audio hooks with the audio components.

The lesser issues relate to problematic "et cetra" categories: *abstractions* and *misc*. This is always a challenge with categorization, but maybe we can sort out a good solution. I really like the *staging* category but I'm conflicted since it would be nice to have shadows together.

I'm wondering if we should have a *content* category: `<HTML>`, `<Text>`, `<Text3DGeometry>` and the upcoming `<Image>` could all go there.

Also, while I like the specificity of the *shaders* category, I wonder if its a bit too narrow. In my version I had a category called *effects*, intentionally decouple from implementation details. If I was browsing extras, I would put for example `<Edges>` in the same category mentally, since I would be looking for cool effects to augment my models.

This is all very loose feedback, and nonbinding. I don't think this kind of thing has a clear answer. I'd love to also hear the perspectives of the rest of the team.

edsunman:
@ixxie thanks for the feedback! I agree with removing the sub-categories and just using domain based categories. I felt it was a bit messy with the components listed with angle brackets < >. What do you think about removing the angle brackets in the sidebar but keeping them in the title?

https://github.com/threlte/threlte/assets/14927523/a00ade00-2226-4e19-8b1e-a54c2486ebff

I don't have any strong opinions on what the categories are called so happy to go with what you suggest!

ixxie:
The problem is that all the other docs have the `<AngledBrackets>` notation. I personally like it, but I'm open to a change. However, changing this in all the docs is maybe something beyond the scope of this PR and should be discussed with @grischaerbe maybe.

The `content` category in the video now also combined you previous `staging` category. I actually liked that category as it was! Maybe you can separate them? If you could then deploy that I could have a look.

edsunman:
Good point about keeping the brackets to match the rest of the docs. I've put them back in and deployed so you can take a look.

Do you think the 'content' could be renamed to 'objects' to make it a bit more specific?

ixxie:
Looks better!

Regarding *content* - I meant the include the kind of stuff that lets you put text and images in the scene. I was thinking that `<Environment>`, `<Stars>`, `<Sky>` and `<Grid>` would be more like *staging* stuff.

Regarding *effects and shaders* I'm somehow not entirely happy with the name. I also realized calling it merely *effects* is going to be very confusing in Svelte 5. How about *special effects*? I'm open to other ideas...

I really would like to have all suspense related stuff contiguous in the menu. So either we break the alphabetical ordering rule just for this, or we split the category. What do you think?

As a final note, it feels like `<Billboard>` and `<Align>` are somehow out of place in *staging*. What was your reasoning to put them there? It may be that the concept of "staging" is very specific and I don't get it, but I was thinking of stuff you put on a stage (like theater set design).

edsunman:
I created a google doc so it's easy to discuss and quickly move things around without pushing lots of changes:

https://docs.google.com/document/d/1PPylp7XjcJS9HElB5jT7r-kPhuxTVcjCbzNi8OSQwZw/edit?usp=sharing

Staging is quite a broad term that can refer to the set dressing but also to how actors are positioned and move around the stage. So I was thinking that _content_ would refer to the objects in the scene and _staging_ would refer to how those things are lit and positioned, with things like `<Float>` , `<Billboard>` etc positioning the objects. I think you are right that environments and backdrops like `<Environment>`, `<Stars>` and `<Sky>` would make more sense in staging so I have moved those. 

With _effects and shaders_ I agree it's tricky to think of a succinct category name. How about _visual effects_?

I think we are getting there! The number of categories and things in each category feels quite nicely balanced.

jerzakm:
> I created a google doc so it's easy to discuss and quickly move things around without pushing lots of changes:
> 
> https://docs.google.com/document/d/1PPylp7XjcJS9HElB5jT7r-kPhuxTVcjCbzNi8OSQwZw/edit?usp=sharing
> 
> Staging is quite a broad term that can refer to the set dressing but also to how actors are positioned and move around the stage. So I was thinking that _content_ would refer to the objects in the scene and _staging_ would refer to how those things are lit and positioned, with things like `<Float>` , `<Billboard>` etc positioning the objects. I think you are right that environments and backdrops like `<Environment>`, `<Stars>` and `<Sky>` would make more sense in staging so I have moved those.
> 
> With _effects and shaders_ I agree it's tricky to think of a succinct category name. How about _visual effects_?
> 
> I think we are getting there! The number of categories and number of things in each category feels quite nicely balanced.

Thanks! I took a stab at it as well. I think the proposal in that doc was a bit vague and arbitrary about where some of the things were put in. I really don't like the"Visual effects" category because it's going to be a generic toss-in-everything-that-looks-good bag. 

It might be my personal preference, but I'd look for familiar naming so I proposed sorting into categories like: Objects, Geometries, Materials, Lighting & Shadows. 

Not everything fits perfectly (yet?)

edsunman:
@jerzakm thanks for your input! Naming categories more specifically (geometries, materials) does make things clearer. There will be awkward things like meshline that fall between categories but I suppose that's what misc is for?

> That being said, we should decide on something in the next day or two and just merge it. Tweaking docs is not a breaking change any ANYTHING is better than what we have now.

Agree with this! @ixxie do you want to have a look at our suggestions in the google doc and make a final decision on a category names and I'll make the changes?

ixxie:
I'm very happy with @edsunman's categories and names as they are now in the doc. The problem with being more specific @jerzakm, is that we end up with many categories. This both extends the length of the sidebar, and makes things harder to find since you need to parse more headers. As Ed mentions earlier, its a balance: you want Goldilocks categories which are neither too big, nor too small.

edsunman:
@ixxie I've pushed those changes so the order matches now the google doc.

ixxie (approved):
LGTM, thanks @edsunman!

PR #914, opened 2024-03-22
Added <Mask> component

edsunman:
I've had a go at porting drei's <Mask> component. It is fairly simple and works alongside a useMask hook to set the stencil properties on chosen material. 

https://github.com/threlte/threlte/assets/14927523/7508a9e2-5e49-47fc-8afd-f9b1715aad64

### DIFF
--- a/.changeset/curly-colts-hug.md
+++ b/.changeset/curly-colts-hug.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": minor
+"@threlte/docs": minor
+---
+
+Added Mask component to Extras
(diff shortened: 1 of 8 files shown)


DefinitelyMaybe:
@edsunman thank you so much for the PR, I really like it!

Where could I go to read up about `stencil buffer`'s

edsunman:
@DefinitelyMaybe To be honest this was mostly a copy and paste job from the drei component so I'm not too familiar with the stencil buffer myself, other than playing around with parameters to create the example above. I had a go at porting <MeshPortalMaterial> but it was going way over my head so decided to work with <Mask> instead 😂

DefinitelyMaybe (approved):
LGTM 😄 Had a quick chat to team, we'll merge now and later on we can improve the docs re explanations/understanding of stencil buffers 👍

Or nobody's worried about it and I'm just overthinking it 🤷‍♂️ it happens 😅

PR #921, opened 2024-04-01
Add public `createThrelteContext()` export

michealparks:
### Overview

This PR adds a new export, `createThrelteContext()`, which creates a `ThrelteContext` and `ThrelteInternalContext`. A few pieces of logic involved in creating the context were moved from the `<Canvas>` component here.

This change was made primarily to support creating contexts for the upcoming test renderer and to allow us to more easily make the `<HUD>` and `<View>` components.

I think that I've moved around everything necessary, but if anyone finds any issues with this approach I'd be happy to restructure it.

I also did not add any documentation for this function, as it's fairly advanced and I can't yet think of a use case outside what I mentioned above, but I'd be up for adding them if it makes sense.

### DIFF
--- a/.changeset/fuzzy-clouds-develop.md
+++ b/.changeset/fuzzy-clouds-develop.md
@@ -1,5 +1,5 @@
 ---
-"@threlte/core": patch
+"@threlte/core": minor
 ---
 
 add createThrelteContext export


packages/core/src/lib/lib/contexts.ts
```diff
@@ -236,10 +249,10 @@ export const createContexts = (options: {
...
     renderMode: currentWritable(options.renderMode),
     autoRender: currentWritable(options.autoRender),
-    scheduler: undefined as any, // will be set later
-    mainStage: undefined as any, // will be set later
-    renderStage: undefined as any, // will be set later
-    autoRenderTask: undefined as any, // will be set later
+    scheduler,
+    mainStage,
+    renderStage,
+    autoRenderTask,
```
DefinitelyMaybe, line 255: thats a nice improvement 👍

grischaerbe (approved):
LGTM 🚀

PR #922, opened 2024-04-01
Improve MeshLineGeometry performance, fix interactivity `NaN` bug

michealparks:
### Overview

Back when we were playing with making a `@threlte/xr` minigame, I noticed that there were some performance issues / large GC events with `MeshLineGeometry` on my Quest 2. This PR attempts to tackle some of them by removing all array copies within that component, which should produce noticeable changes within the `setPoints` function since it gets called quite often during animations. Also, by using the `BufferAttribute.set*` functions I was able to reduce the overall amount of code there.

When playing around with these optimizations I noticed that the `cursor-outlines` example sometimes crashes on startup. Turns out there was a divide by zero bug in the `interactivity` plugin. That's fixed as well.


DefinitelyMaybe (approved):
Top notch 👍

edsunman:
Looks great, code is much neater!

Issue #920, opened 2024-04-01
Incorrect font in docs


edsunman:
A very minor issue... I think the font in the docs is supposed to be Inter (specified in tailwind config) but it is currently being overridden by: _system-ui, -apple-system, BlinkMacSystemFont, etc,_. This is not very noticeable on Apple devices because the font is similar but on Windows the system font is not as nice!

I think it is leaking out from [here](https://github.com/threlte/threlte/blob/6503288959dbcaeccfe28af96091fbdd0def6dba/packages/flex/src/routes/%2Bpage.svelte#L110) and/or [here](https://github.com/threlte/threlte/blob/6503288959dbcaeccfe28af96091fbdd0def6dba/apps/docs/src/examples/flex/examples/App.svelte#L96).

Can this font-family property be removed? I'm not sure if it is needed for the `<Flex>` examples.

DefinitelyMaybe:
```css
:global(body) {
    background: #333333;
    /*margin: 0; etc...*/
  }
  ```
:grimacing: not the best setup for I agree.

I think all examples become iframes so I'm not 100% whether this is the source of the overriding.

I'm on a windows machine myself, is there some steps to reproduce the effect?

edsunman:
https://github.com/threlte/threlte/assets/14927523/ac0947f0-18ca-426a-9c9a-f3466ff6ce7c

To replicate if you disable that CSS selector in dev tools the correct font will display.

Interestingly this problem does not occur when running `pnpm run dev` locally. I needed to build the project and run `pnpm run preview` to replicate the issue. So maybe Astro is pulling out that CSS at build time and applying it globally?

DefinitelyMaybe:
oh! I see, and basically a search across the whole repo for that body css comes up with the files you mentioned earlier.

Right, don't mind me. jump straight into making a PR and I'll merge 😅

grischaerbe:
How did I not notice that? 🫣 Happy to merge any PR 👍

PR #924, opened 2024-04-02
Docs: font fix

edsunman:
Fix for this issue: https://github.com/threlte/threlte/issues/920

I've removed the font-family properties that are being applied globally. I couldn't see anywhere in the flex examples where they were being used.


grischaerbe (approved):
Thanks so much @edsunman ❤️ !

PR #926, opened 2024-04-05
Remove Svelte component proxies from theater package

michealparks:
### Overview

In Svelte 5 proxying a component becomes a lot more complicated, and mutating / adding / removing props sent to the component from the proxy often breaks (like in the case of $$restProps). This PR gets ahead of the Svelte 5 migration by removing the svelte component proxies created in the Theater package and replaces the props they add with an internal context.

### DIFF
--- a/.changeset/curly-trees-train.md
+++ b/.changeset/curly-trees-train.md
@@ -1,0 +1,5 @@
+---
+"@threlte/theatre": patch
+---
+
+Remove svelte component proxies.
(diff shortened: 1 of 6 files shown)


grischaerbe (approved):
\*Technically\* this is a breaking change:

```svelte
<SheetObject let:Transform key="a">
  <Transform>
    <T.Mesh />
    <SheetObject let:Transform={Transform2} key="b">
      <!-- This will now reference sheet object with key "b" -->
      <Transform>
        <T.Mesh />
      </Transform>
    </SheetObject>
  </Transform>
</SheetObject>
```

I highly doubt there's a usage of this pattern out in the wild, what do you think? If you're fine with it, feel free to merge 👍  Maybe we can add a little note about this pattern change in the changelogs? For that, we'd need to adapt the changeset.

Apart from that as always perfect 👍

michealparks:
Closing since we'll get this sorted in the next branch

Issue #933, opened 2024-04-11
[next] Do not invalidate the frame in `<Grid>`


grischaerbe:
This usage of `useTask` invalidates every frame.

https://github.com/threlte/threlte/blob/45fae01f883ce8bd7802bca93f9c5be142faa0bd/packages/extras/src/lib/components/Grid/Grid.svelte#L208-L218

PR #931, opened 2024-04-10
[next] Initial next branch cleanup

michealparks:
Just gettin' some more tedious stuff out of the way 🧹 

This PR mostly consists of repetitive cleanup following the initial next branch:
* Migrate more components to runes mode. All that's left after this is `theater`, `xr`, and some `extras` that need care.
* Fix incorrect type issues with `ref` bindings with many extras components.
* Remove usage of `$$Props`, since the `$props` rune makes it unnecessary.
* Make sure all repos are using the latest and same dependency versions.

And a few one-off fixes:
* Fix animated sprite material bug.
* Add a few missing component `ref` bindings.

PR #934, opened 2024-04-12
[next] Grid invalidate fix

michealparks:
Fixes https://github.com/threlte/threlte/issues/933

Removed the `invalidate()` call in the Grid component's useTask so that it won't invalidate every frame.

Invalidation should occur when the camera moves or when the grid position changes, which if I recall will usually involve their own invalidate calls. 

I also synced the Grid.svelte file with #931 to avoid conflicts.

Let me know if there are any issues with this solution and I'll fix!

### DIFF
--- a/.changeset/odd-comics-begin.md
+++ b/.changeset/odd-comics-begin.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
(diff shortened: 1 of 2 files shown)


grischaerbe (approved):
LGTM 🚀 Thanks!

PR #955, opened 2024-05-02
[next] Docs update loading assets page

DefinitelyMaybe:
Admittedly, when I first saw this attach function, I didn't really understand it.

```html
<T.MeshStandardMaterial
    map={texture2}
    attach={(parent, self) => {
      if (Array.isArray(parent.material)) parent.material = [...parent.material, self]
      else parent.material = [self]
    }}
  />
```

I've created a second alternative, which hopefully helps others figure it out sooner 😅 It's also a example of how we can mix and match 'normal' three.js code 👍


DefinitelyMaybe:
prettier is only worried about xr files

```shell
Checking formatting...
[warn] packages/xr/src/lib/plugins/pointerControls/hook.ts
[warn] packages/xr/src/lib/plugins/pointerControls/setup.ts
[warn] packages/xr/src/lib/plugins/teleportControls/hook.ts
[warn] Code style issues found in 3 files. Run Prettier to fix.
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

this PR is fine 😅

PR #959, opened 2024-05-08
docs: Add warning for <MASK>

### LINKED ISSUE #958: stencil needs to be manually enabled starting from threejs 0.163.0 for Mask to work
Hey, I noticed the Mask component suddenly stopped working for me so I did some digging. Turns out the stencil buffer, starting from Three.js 0.163.0, is now disabled by default, breaking the Mask component.

To re-enable it, the appropriate renderer parameter needs to be set to true like so:

```
<Canvas rendererParameters={{ stencil: true }}>
  <Mask id={1}>
...
```

Creating this issue for two reasons:

1. Hope that others having the same issue come across it.
2. To find an approach how to deal with it in Threlte.

Personally I see two possible solutions.

1. Add a hint in the docs that tells users to make sure that the stencil buffer is enabled in the renderer.
2. The Mask component automatically taking care of enabling it, e.g. in the onMount method.

Not sure if there could be some unwanted side effects from enabling the stencil buffer when going with approach 2 though.

Let me know what you think! :slightly_smiling_face:   

References:
- [Three.js 0.163.0 Release Notes](https://github.com/mrdoob/three.js/releases/tag/r163)
- [PR setting stencil to false by default](https://github.com/mrdoob/three.js/pull/27900)

DefinitelyMaybe:
closes #958

Due to recent changes in three.js. Also prettier did some formatting.

### DIFF
--- a/apps/docs/src/content/reference/extras/mask.mdx
+++ b/apps/docs/src/content/reference/extras/mask.mdx
@@ -10,9 +10,27 @@
       extends: { type: 'Mesh', url: 'https://threejs.org/docs/index.html#api/en/objects/Mesh' },
       'props':
         [
-          { name: 'id', type: 'number', required: false,default: '1', 'description': 'Each mask must have an id, you can have compound masks referring to the same id.' },
-					 { name: 'colorWrite', type: 'boolean', required: false,default: 'false', 'description': 'If colors of the masks own material will leak through.' },
-					  { name: 'depthWrite', type: 'boolean', required: false,default: 'false', 'description': 'If depth  of the masks own material will leak through.' }
+          {
+            name: 'id',
+            type: 'number',
+            required: false,
+            default: '1',
+            'description': 'Each mask must have an id, you can have compound masks referring to the same id.'
+          },
+          {
+            name: 'colorWrite',
+            type: 'boolean',
+            required: false,
+            default: 'false',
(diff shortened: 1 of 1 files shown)


apps/docs/src/content/reference/extras/mask.mdx
```diff
@@ -10,16 +10,35 @@
...
+            'description': 'If depth  of the masks own material will leak through.'
+          }
         ]
     }
 }
 ---
 
 Masks use the stencil buffer to cut out areas of the screen. This component is a port of [drei's `<Mask>` component](https://github.com/pmndrs/drei#mask).
 
+<Tip type="warning">Starting from [three.js v163](https://github.com/mrdoob/three.js/releases/tag/r163), stencil is set to `false` by default. To enable stencil's, set your canvas's renderer params like so `<Canvas rendererParameters={{ stencil: true }}>`. In prior versions of three.js the default is `true` already.</Tip>
```
michealparks, line 41:
```suggestion
<Tip type="warning">The Mask component requires Three.js to render with a stencil buffer. As of [r163](https://github.com/mrdoob/three.js/releases/tag/r163), stencil is set to `false` by default. To enable the stencil buffer, set it in your canvas's renderer params: `<Canvas rendererParameters={{ stencil: true }}>`. In prior versions the default is `true` already.</Tip>
```

michealparks (approved):

PR #961, opened 2024-05-12
Added KTX2Loader.Dispose to prevent memory leak

clibequilibrium:
Hi, loading multiple GLTF files that has KTX2 textures results in a memory leak. For complex scenes it goes as far as 1GB of data.
image

This PR disposes KTX2loader instances after the gltf is loaded. 

image

I am also curious what is your opinion of disposing draco loader ? I have 20-50 objects that are draco compressed and Draco Loader also stays in memory resulting in 120 MB of memory (4 workers, 30 mb each).

Not disposing KTX2loaders

image

vs

Disposing removes 2GB of data making Threlte very mobile friendly when dealing with GLTFs that are draco compressed and have KTX textures 🙌

image

EDIT: I figured that draco loader only uses 30 mb per web worker. So on mobile I can simply setWorkerLimit to minimize extra overhead. So this PR targets KTXLoader only.

### DIFF
--- a/.changeset/yellow-needles-jump.md
+++ b/.changeset/yellow-needles-jump.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
+dispose ktx2Loader's after load to prevent over-use of memory
--- a/packages/extras/src/lib/hooks/useGltf.ts
+++ b/packages/extras/src/lib/hooks/useGltf.ts
@@ -54,6 +54,7 @@ export function useGltf<
     } {
   const { renderer } = useThrelte()
   const opts = typeof urlOrOptions === 'string' ? options : urlOrOptions
+  const ktx2Loader =  opts?.ktxTranscoderPath ? new KTX2Loader() : null;
   const loader = useLoader(GLTFLoader, {
     extend(loader) {
       if (opts?.useDraco) {
@@ -79,7 +80,6 @@ export function useGltf<
       }
 
       if (opts?.ktxTranscoderPath) {
-        const ktx2Loader = new KTX2Loader()
         ktx2Loader.setTranscoderPath(opts?.ktxTranscoderPath)
         ktx2Loader.detectSupport(renderer)
         loader.setKTX2Loader(ktx2Loader)
@@ -90,6 +90,7 @@ export function useGltf<
   const load = (url: string) => {
     return loader.load(url, {
       transform(result) {
+        ktx2Loader?.dispose();
         return {
           ...result,
           ...buildSceneGraph(result.scene)
--- a/packages/extras/src/lib/hooks/useGltf.ts
+++ b/packages/extras/src/lib/hooks/useGltf.ts
@@ -80,9 +80,9 @@ export function useGltf<
       }
 
       if (opts?.ktxTranscoderPath) {
-        ktx2Loader.setTranscoderPath(opts?.ktxTranscoderPath)
-        ktx2Loader.detectSupport(renderer)
-        loader.setKTX2Loader(ktx2Loader)
+        ktx2Loader!.setTranscoderPath(opts?.ktxTranscoderPath)
+        ktx2Loader!.detectSupport(renderer)
+        loader.setKTX2Loader(ktx2Loader!)
       }
     }
   })
--- a/packages/extras/src/lib/hooks/useGltf.ts
+++ b/packages/extras/src/lib/hooks/useGltf.ts
@@ -54,7 +54,7 @@ export function useGltf<
     } {
   const { renderer } = useThrelte()
   const opts = typeof urlOrOptions === 'string' ? options : urlOrOptions
-  const ktx2Loader =  opts?.ktxTranscoderPath ? new KTX2Loader() : null;
+  const ktx2Loader = opts?.ktxTranscoderPath ? new KTX2Loader() : null
   const loader = useLoader(GLTFLoader, {
     extend(loader) {
       if (opts?.useDraco) {
@@ -90,7 +90,7 @@ export function useGltf<
   const load = (url: string) => {
     return loader.load(url, {
       transform(result) {
-        ktx2Loader?.dispose();
+        ktx2Loader?.dispose()
         return {
           ...result,
           ...buildSceneGraph(result.scene)


DefinitelyMaybe:
Hey thank you so much for making this PR and your discussion over on discord ❤️ 

Looks like a good improvement.

clibequilibrium:
> Hey thank you so much for making this PR and your discussion over on discord ❤️
> 
> Looks like a good improvement.
> 
> Would you like me to have a look into prettier's warnings?

Thanks a lot, please do. I confess I did this PR in vscode.dev web version, so I did not compile to see the error. For some reason the GitHub action didn't show what is the issue too.

michealparks:
Thank you! Great fix, could you [add a changeset when you get a chance](https://github.com/threlte/threlte/blob/main/CONTRIBUTING.md#changesets)? That'll trigger shipping a new package to NPM when this gets merged.

DefinitelyMaybe:
Fair enough. What're our options? Could we:

- Make sure only ever one KTX2/Draco/etc... is used?
- Know when all models are loaded and dispose them at that point?

clibequilibrium:
> Fair enough. What're our options? Could we:
> 
> * Make sure only ever one KTX2/Draco/etc... is used?
> * Know when all models are loaded and dispose them at that point?
> * other?

I think it would be better to expose custom KTX loader as an option, same way that custom dracoLoader can be passed to useDraco. But by default <GLTF/> uses a single DracoLoader, it's only KTXloaders that get leaked. 

@grischaerbe yes it would allocate a loader per gltf and safely dispose it. I did not encounter any problems on a scene with lots of GLTF models 20+ (draco compressed + KTX).

michealparks:
I think that what @grischaerbe is concerned about is `useGltf`'s alternative function signature:

```ts
const { load } = useGltf(options)
```

When it's used this way `load` can be called any number of times, so diposing the loader after the initial call will cause problems.

> Make sure only ever one KTX2/Draco/etc... is used?

I like this solution, since - unless someone can correct me - there shouldn't be an issue with `useGltf` sharing one loader instance. Something like:

```ts
const ktx2Loader: KTX2Loader | undefined

export const useGltf = (...) => {
  // Add this as an option, can be boolean, transcoder path, or loader instance
  if (useKtx) {
    if (useKtx === true && ktx2Loader === undefined) {
      // initialize loader and save it to the global ktx2Loader var
    } else if (typeof useKtx === 'string') {
      // initialize single loader transcoder path, don't save it to the global loader var
    } else {
      // use loader instance
    }
  }
}
```

What do y'all think?

This is suggestion is also a breaking change, so maybe if we like it then perhaps it should go into [next]?

clibequilibrium:
> I think that what @grischaerbe is concerned about is `useGltf`'s alternative function signature:
> 
> 
> 
> ```ts
> 
> const { load } = useGltf(options)
> 
> ```
> 
> 
> 
> When it's used this way `load` can be called any number of times, so disposing the loader after the initial call will cause problems.
> 
> 
> 
> > Make sure only ever one KTX2/Draco/etc... is used?
> 
> 
> 
> I like this solution, since - unless someone can correct me - there shouldn't be an issue with `useGltf` sharing one loader instance. Something like:
> 
> 
> 
> ```ts
> 
> const ktx2Loader: KTX2Loader | undefined
> 
> 
> 
> export const useGltf = (...) => {
> 
>   // Add this as an option, can be boolean, transcoder path, or loader instance
> 
>   if (useKtx) {
> 
>     if (useKtx === true && ktx2Loader === undefined) {
> 
>       // initialize loader and save it to the global ktx2Loader var, then use it
> 
>     } else if (useKtx === true) {
> 
>       // global loader is initialized, use it
> 
>     } else if (typeof useKtx === 'string') {
> 
>       // initialize single loader transcoder path, don't save it to the global loader var, use it
> 
>     } else {
> 
>       // use loader instance
> 
>     }
> 
>   }
> 
> }
> 
> ```
> 
> 
> 
> What do y'all think?
> 
> 

My only worry with complex scenes. It would spawn x number of web workers, so by default the limit is 4. On a complex scenes that would be 8 workers total (4 draco, 4 ktx). After scene is loaded they would stay in memory causing 300MB overhead. Maybe disposing them and recreating once would work. Because every subsequent load would not require all 8 workers.

michealparks:
> It would spawn x number of web workers
@clibequilibrium In my example above by default (the first `if` block) only one KTX2Loader instance would be created for the entire app, so only web worker would be spawned. If users wanted to supply their own instance then it's the same deal, so I think the solution would scale pretty well!

clibequilibrium:
> > It would spawn x number of web workers
> 
> @clibequilibrium In my example above by default (the first `if` block) only one KTX2Loader instance would be created for the entire app, so only web worker would be spawned. If users wanted to supply their own instance then it's the same deal, so I think the solution would scale pretty well!

No no I understand. But even a single loader can spawn up to 4 workers by default , because default worker limit is 4. 
Imagine a scene where there is 1 gltf with KTX , then a single KTXLoader will only create 1 worker.

But in a scene with 10 to 20 gltf files, KTX loader spawns up to 4 web workers (each of those is 30 mb).
So the scenario I am describing is of course having single ktx loader is just on a complex scene it would spawn 4 workers, but then won't dispose them.

DefinitelyMaybe:
I was wondering where this limit of 4 comes from, [it's the default value](https://threejs.org/docs/index.html?q=loader#examples/en/loaders/KTX2Loader.setWorkerLimit)

michealparks:
> but then won't dispose them.

Gotcha, thanks for clarifying. I believe that concern would be handed by my proposal as well? If the user wants only one loader and also wants to control its disposal they could do this:

```ts
// myKtxLoader.ts
export let ktx2Loader = new KTX2Loader()
```
```ts
// MyComponent.svelte
import { ktx2Loader } from './myKtxLoader.ts'

useGltf('./file1.glb', { useKtx: ktx2Loader })
useGltf('./file2.glb', { useKtx: ktx2Loader })
useGltf('./file3.glb', { useKtx: ktx2Loader })

// Later, when all files are loaded
ktx2Loader.dispose()
```

clibequilibrium:
> > but then won't dispose them.
> 
> Gotcha, thanks for clarifying. I believe that concern would be handed by my proposal as well? If the user wants only one loader and also wants to control its disposal they could do this:
> 
> ```ts
> // myKtxLoader.ts
> export let ktx2Loader = new KTX2Loader()
> ```
> 
> ```ts
> // MyComponent.svelte
> import { ktx2Loader } from './myKtxLoader.ts'
> 
> useGltf('./file1.glb', { useKtx: ktx2Loader })
> useGltf('./file2.glb', { useKtx: ktx2Loader })
> useGltf('./file3.glb', { useKtx: ktx2Loader })
> 
> // Later, when all files are loaded
> ktx2Loader.dispose()
> ```



> > but then won't dispose them.
> 
> Gotcha, thanks for clarifying. I believe that concern would be handed by my proposal as well? If the user wants only one loader and also wants to control its disposal they could do this:
> 
> ```ts
> // myKtxLoader.ts
> export let ktx2Loader = new KTX2Loader()
> ```
> 
> ```ts
> // MyComponent.svelte
> import { ktx2Loader } from './myKtxLoader.ts'
> 
> useGltf('./file1.glb', { useKtx: ktx2Loader })
> useGltf('./file2.glb', { useKtx: ktx2Loader })
> useGltf('./file3.glb', { useKtx: ktx2Loader })
> 
> // Later, when all files are loaded
> ktx2Loader.dispose()
> ```

This would absolutely work!

grischaerbe:
Thanks @clibequilibrium for your effort ❤️ 
We decided to go for @michealparks implementation of having multiple hooks in #969.

clibequilibrium:
> Thanks @clibequilibrium for your effort ❤️ We decided to go for @michealparks implementation of having multiple hooks in #969.

Awesome ! please make sure you backport the memory leak fix to Svelte 4 branch, otherwise it's an outstanding issue for anyone who would use Threlte as is right now.

PR #962, opened 2024-05-13
[next] Remove `useframe` and `useRender` hooks

michealparks:
✂️


grischaerbe (approved):
Nice! Thanks so much ❤️ 
I adapted the spaceship example to use the render stage.

LGTM 🚀

michealparks:
Ah, thanks for catching that!

PR #966, opened 2024-05-15
[next] Remove forwardEventHandlers usage

michealparks:
This PR removes our remaining usage of `forwardEventHandlers` in the docs and in the `gltf` package.

This fixes event forwarding in the `custom-abstractions` and `handling-events` lessons and `svelte` files generated by the GLTF package, which currently do not work due to `forwardEventHandlers` being removed.

### DIFF
--- a/apps/docs/src/examples/test/Box.svelte
+++ b/apps/docs/src/examples/test/Box.svelte
@@ -1,32 +1,32 @@
 <script lang="ts">
-  import { forwardEventHandlers, T } from '@threlte/core'
+  import { T } from '@threlte/core'
   import { useCursor } from '@threlte/extras'
   import { spring } from 'svelte/motion'
 
-  let color = 'white'
+  let { ...props } = $props()
+
+  let color = $state('white')
 
   const scale = spring(1)
-  const component = forwardEventHandlers()
 
   const { onPointerEnter, onPointerLeave } = useCursor()
 </script>
 
 <T.Group
   scale={$scale}
-  {...$$restProps}
-  bind:this={$component}
+  {...props}
 >
   <T.Mesh
     position.y={0.5}
-    on:pointerenter={onPointerEnter}
-    on:pointerleave={onPointerLeave}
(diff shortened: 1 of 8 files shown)


apps/docs/src/examples/test/Box.svelte
```diff
@@ -1,32 +1,32 @@
...
-  const component = forwardEventHandlers()
 
   const { onPointerEnter, onPointerLeave } = useCursor()
 </script>
 
 <T.Group
   scale={$scale}
-  {...$$restProps}
-  bind:this={$component}
+  {...props}
```
DefinitelyMaybe, line 17: nice. that's a good improvement on ergonomics imo

apps/docs/src/examples/test/Box.svelte
```diff
@@ -1,32 +1,32 @@
...
 >
   <T.Mesh
     position.y={0.5}
-    on:pointerenter={onPointerEnter}
-    on:pointerleave={onPointerLeave}
     on:pointerenter={() => {
```
DefinitelyMaybe, line 21: oh, what was the situation with `on:___` vs `on___`. Is that something to think about here or not yet?

apps/docs/src/examples/test/Box.svelte
michealparks: Agreed!

apps/docs/src/examples/test/Box.svelte
michealparks: I set up the next branch so that users wouldn't have to migrate or change any events yet (as long as they're not doing things like forwarding events). 

The idea was to defer thinking about how to migrate events to threlte 9. @grischaerbe may want to chime in though

michealparks: In this specific example, multiple duplicate events listeners won't work anymore due to svelte's internals, we could potentially add more signaling to users by giving specific error messages

PR #972, opened 2024-05-22
[next] Add Speculation API

grischaerbe:
This one is for @jerzakm  :)
It adds the Speculation API for instant loads. Currently, all pages are set to prerender moderately eager.


jerzakm:
I was checking for `data-astro-prefetch` attribute, and we already seem to have it in most places where it makes sense, like here:
https://github.com/threlte/threlte/blob/next/apps/docs/src/components/Menu/LeftSidebar/LeftSidebarCategory.svelte#L45

So LGTM .🚀🚀🚀

PR #1002, opened 2024-06-23
Replace events with callback props

michealparks:
### Overview

* Events: Replaces all events with callback props. It replaces all instances of `on:event` with `onevent`
* Snippets: Replaces all `<slots>` with snippets and replaces all `let:prop` directives with snippet props.


apps/docs/src/examples/rapier/rigid-body/Particle.svelte
```diff
@@ -35,10 +35,11 @@
     }
   })
 
-  const fireSound = (e: ContactEvent) => {
+  const fireSound: ContactEvent = (event) => {
     if ($muted) return
-    const volume = clamp((e.detail.totalForceMagnitude - 30) / 1100, 0.1, 1)
+    const volume = clamp((event.totalForceMagnitude - 30) / 1100, 0.1, 1)
     const audio = audios.find((a) => a.volume >= volume)
+    console.log('here')
```
DefinitelyMaybe, line 42:
```suggestion
```

packages/extras/src/lib/components/Grid/Grid.svelte.d.ts
```diff
@@ -18,7 +18,6 @@ export type GridProps = Props<Mesh> & {
   fadeDistance?: number
   fadeStrength?: number
   side?: Side
-  children?: GridSnippet
```
DefinitelyMaybe, line 21: I've noticed a removing the optional `children` prop a couple of times, we didn't need it did we because `children` is a reserved prop right?

packages/extras/src/lib/components/InstancedSprite/InstancedSprite.svelte.d.ts
```diff
@@ -4,7 +4,7 @@ import type {
   SpritesheetFormat
 } from '@threejs-kit/instanced-sprite-mesh'
 import type { Props } from '@threlte/core'
-import { SvelteComponent } from 'svelte'
+import { SvelteComponent, type Snippet } from 'svelte'
```
DefinitelyMaybe, line 7:
```suggestion
import type { SvelteComponent, Snippet } from 'svelte'
```

packages/extras/src/lib/components/RoundedBoxGeometry/RoundedBoxGeometry.svelte
```diff
@@ -46,17 +40,18 @@
...
+<T
+  is={geometry}
   bind:ref
-  args={[shape, params]}
-  on:create={({ ref, cleanup }) => {
-    ref.center()
-    toCreasedNormals(ref, creaseAngle)
-    props.$$events?.create?.({ ref, cleanup })
-  }}
   {...props}
```
DefinitelyMaybe, line 54: I like this change 😄

packages/extras/src/lib/components/InstancedSprite/InstancedSprite.svelte.d.ts
DefinitelyMaybe: I think this is a bit nit picky of me

packages/extras/src/lib/components/Grid/Grid.svelte.d.ts
michealparks: It's already provided by the threlte `Props` export, so we don't need to explicitly add it unless the children are providing more properties than the standard `ref` 🎉

grischaerbe (approved):
Holy cow, what a work 💪 
LGTM although I didn't check extremely thoroughly. We still can fix upcoming issues downstream 👍

PR #1003, opened 2024-06-24
[next] more docs clean up

DefinitelyMaybe:
there was some `import { Mesh } from '@threlte/core'` which is no longer a thing. `interactive` prop removed too.


michealparks (approved):
🚀

Issue #949, opened 2024-04-19
[next] Gizmo does not accept clicks


grischaerbe:

DefinitelyMaybe:
closing for now. We have a [new gizmo](https://next.threlte.xyz/docs/reference/extras/gizmo) since #1254 was merged 🎉 

happy for the issue to be reopen by OP if needed

Issue #849, opened 2024-01-31
`<Gizmo />` not reacting to click events


janosh:
I may have a similar issue as #809 where the `<Gizmo />` component does not react to click events. I tried clicking all around it to see if the active area was misplaced but didn't manage to find it. I did this with both `@threlte/extras@8.5.0` including the fix in #810 and `@threlte/extras@8.1.1` without it. Neither case worked for me.

Still actively debugging this over at https://elementari.janosh.dev/structure. Relevant commit: https://github.com/janosh/elementari/pull/46/commits/e6d7ca96e1.

janosh:
can confirm my issue is similar to https://github.com/threlte/threlte/issues/809. the `Gizmo` `clickTarget` is far outside the canvas.

![Screenshot 2024-03-22 at 09 31 49](https://github.com/threlte/threlte/assets/30958850/51eed969-6d30-499b-89f7-6d067e6150c4)

not easy to debug this as changing the code in `node_modules/.pnpm/@threlte+extras@8.10.3_svelte@4.2.12_three@0.162.0/node_modules/@threlte/extras/dist/components/Gizmo/Gizmo.svelte` has no effect, maybe because `vite` actually loads an optimized/transpiled version of the code from somewhere else?

michealparks:
Yep, I think the only way to realistically debug this is to clone + download the threlte repo and run the docs or extras playground locally. I've been meaning to look at this issue myself since I use gizmo a lot but unfortunately haven't had the time lately :(

DefinitelyMaybe:
closing for now. We have a [new gizmo](https://next.threlte.xyz/docs/reference/extras/gizmo) since #1254 was merged 🎉 

happy for the issue to be reopen by OP if needed

PR #1009, opened 2024-06-24
[next] Separate contexts by domain

### LINKED ISSUE #998: [next] Domain-specific contexts
Currently, Threlte uses two contexts, one of which is part of the public API.

In an effort to support advanced patterns like `<HUD>` or `useViewport` we want to separate these two contexts to a multitude of contexts which can be created at any point in the component tree. We're therefore opting for domain-specific contexts:

- Scene Context
	- scene
- Camera Context
	- camera
- Renderer Context
	- colorSpace
	- toneMapping
	- dpr
	- shadows
	- colorManagementEnabled
	- useLegacyLights
	- renderer
- Scheduler Context
	- scheduler
	- renderMode
	- autoRender
	- frameInvalidated
	- advance
	- autoInvalidations
	- resetFrameInvalidation
	- mainStage
	- renderStage
	- autoRenderTask
	- shouldRender
- Disposal Context
	- dispose
	- collectDisposableObjects
	- addDisposableObjects
	- removeDisposableObjects
	- disposableObjects
	- shouldDispose
- Cache Context
	- remember
	- clear
- (Plugins Context)
	- plugins

Every context has at least two methods:

- `createXContext` to create the context
  - `(args) => ContextType`
- `useX` to use the context
  - `() => ContextType`

On top of that, there will be convenience functions that create *all* contexts at once:

- `createThrelteContext`

This PR intentionally leaves out the size management that Threlte currently does as it will not be used in Threlte 8.

grischaerbe:
Fixes #998 

This PR also addresses a flaw in the design of the `<Canvas>` component: Since the `<Canvas>` markup only contains a `<canvas>` element, there's no way to have authority over added HTML elements, since they cannot be children to a `<canvas>` element. This PR introduces a new layout:

```html
<div>
  <canvas></canvas>
</div>
```

which gives us authority over elements being added as children to the `<div>` element. This paves the way for #809, #849, #949


packages/core/src/lib/context/fragments/renderer.svelte.ts
```diff
@@ -0,0 +1,210 @@
...
+  if ('setAnimationLoop' in context.renderer) {
+    const setAnimationLoop = context.renderer.setAnimationLoop as (
+      callback: (time: number) => void
+    ) => void
+    setAnimationLoop((time) => {
+      dispose()
+      scheduler.run(time)
+      resetFrameInvalidation()
+    })
+  }
```
michealparks, line 170: When is this not true?

packages/core/src/lib/context/fragments/renderer.svelte.ts
```diff
@@ -0,0 +1,210 @@
...
+  onDestroy(() => {
+    if ('dispose' in context.renderer) {
+      const dispose = context.renderer.dispose as () => void
+      dispose()
+    }
+  })
+
+  $effect(() => {
+    context.colorManagementEnabled.set(options.colorManagementEnabled ?? true)
+  })
```
michealparks, line 181:
```suggestion
```
duplicate ✂️

packages/core/src/lib/context/fragments/renderer.svelte.ts
grischaerbe: `context.renderer` is now extending the generic `THREE.Renderer` which currently is this:

```ts
export interface Renderer {
    domElement: HTMLCanvasElement;

    render(scene: Object3D, camera: Camera): void;
    setSize(width: number, height: number, updateStyle?: boolean): void;
}
```

michealparks: Ah!

michealparks (approved):
😍

PR #1010, opened 2024-06-24
[next] updated astro and mdx dependencies

### LINKED ISSUE #1004: [next] Fix Manual Install Guide
For some reason it's broken and only shows "undefined".

DefinitelyMaybe:
closes #1004

Also thought that `ManualInstallGuide` shouldn't be auto imported into all pages so I've removed that from the config and inserted an import into the install page


grischaerbe (approved):
Oh awesome! Good catch with the import 👍 
LGTM 🚀

PR #1011, opened 2024-06-25
[Next] docs clean up

DefinitelyMaybe:
in an effort to reduce console warnings 😄 Also noticed mask needed updating. Already liking the way to initialize a three renderer with different params 👍


grischaerbe (approved):
Awesome, LGTM 🚀 
Thanks for fixing the renderers!

Issue #1089, opened 2024-08-29
Nice to have: Gizmo props for custom labels 


spikespiegel91:
Hello! 
It would be nice to be able to change the Gizmo default axis label text (X, Y , Z) with a custom user label. This would be useful in case you want to show "N" and "E" in the Gizmo (North, East) for geographic applications. 

Currently the axis labels are hardcoded text ('X', 'Y', 'Z') but adding yLabel?: string  to GizmoProps in Gizmo.d.ts and setting the props in Gizmo.svelte doesn't take much and would male this possible.

Here is an example of using a custom label in the Y axis: 
![image](https://github.com/user-attachments/assets/2864735c-d670-411c-ba0a-1d6d7dfa46fd)

![image](https://github.com/user-attachments/assets/b0b52b1f-9129-4b18-bc3e-83d782b7e297)

DefinitelyMaybe:
Hey! That's an easy as idea for the Gizmo component. I'm sure some folks might like to adjust their labelling to like X -> Y or Z -> Y. If compass directions work for you then you do you 😄

PR #1090, opened 2024-08-29
Nice to have: Gizmo props for custom labels

spikespiegel91:
Following this discussion:
https://github.com/threlte/threlte/issues/1089#issue-2495485226

This is a minor update of Extras/Gizmo component to add xLabel, yLabel and zLabel as props (insted of hardcoded 'X', 'Y', 'Z' values) to this component. 

Please forgive me if I messed something up or of this is not the propper way to proceed with this change. This is actually my first PR in Github, but I think Threlte is pretty neat!

### DIFF
--- a/.changeset/forty-bugs-divide.md
+++ b/.changeset/forty-bugs-divide.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": minor
+---
+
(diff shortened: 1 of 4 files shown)


DefinitelyMaybe:
please add a changeset so that packaging will pick up the update 👍

DefinitelyMaybe (approved):

PR #1091, opened 2024-08-29
change `oncreate` event api to return optional cleanup function

### LINKED ISSUE #1088: [next] eliminate cleanup function in `oncreate` prop in favor of returning optional cleanup function
currently to run a function on cleanup you do something like this

```svelte
<T.Mesh
  oncreate={({ref, cleanup}) => {
    cleanup(() => {
      console.log(ref);
    });
  }}
>
  <T.BoxGeometry />
</T.Mesh>
```

i think `oncreate` should return an optional cleanup function similar to svelte 5's `$effect`. `$effect` runs an optional cleanup function when a component is destroyed (and before the effect runs again)

```svelte
<script>
  $effect(() => {
    return () => {
      console.log('goodbye');
    }
  });
</script>
```

`oncreate` could mimic this behavior. something like this:

```svelte
<T.Mesh
  oncreate={(ref) => {
    return () => {
      console.log(ref);
    }
  }}
>
<T.BoxGeometry />
</T.Mesh>
```

this isn't exactly the same as `$effect` because it won't run reactively (and it shouldn't) but i think it's a little simpler overall.

i took a look at the [implementation](https://github.com/threlte/threlte/blob/next/packages/core/src/lib/components/T/utils/useCreateEvent.ts)

it looks like you could do something like this in `dispatchCreateEvent`:

```typescript
const dispatchCreateEvent = () => {
  // ...
  const cleanup = oncreate?.(ref);
  if (cleanup !== undefined) {
    cleanupFunctions.push(cleanup);
  }
  // ...
}
```

but i'm not entirely sure it's that simple.

if this change did occur, it would have to be added to the migration guide and could cause some friction when people upgrade to the latest version of threlte. they've already got to replace `on:create` with `oncreate` unless there will be a migration script to help with that.

joshwashywash:
closes #1088

also updates the docs to accommodate and match the api change

haven't tested yet and there isn't a changeset but i can add one if requested.

### DIFF
--- a/.changeset/brave-snails-beam.md
+++ b/.changeset/brave-snails-beam.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": major
+---
+
+Change oncreate API to return a cleanup function instead of passing it to a function argument
--- a/.changeset/forty-cheetahs-own.md
+++ b/.changeset/forty-cheetahs-own.md
@@ -1,0 +1,5 @@
+---
+"create-threlte": minor
+---
+
+Adapt oncreate API design
--- a/.changeset/tender-taxis-explode.md
+++ b/.changeset/tender-taxis-explode.md
@@ -1,0 +1,5 @@
+---
+"@threlte/studio": patch
+---
+
+Adapt oncreate API design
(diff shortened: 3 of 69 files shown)


grischaerbe:
Hey @joshwashywash, thanks for your contribution! I hope you're fine with me refining the PR a bit. The new `oncreate` event can only be passed once to the component, so we don't need to maintain an array of cleanup functions. I also adapted all docs occurences and packages.

joshwashywash:
awesome sounds good to me. thanks @grischaerbe

PR #1093, opened 2024-09-07
[next] scene graph refactoring

grischaerbe:


apps/docs/src/content/learn/advanced/migration-guides.mdx
```diff
@@ -81,6 +81,64 @@ to the component you wish to forward events to.
...
+    return () => {
+      console.log('detaching', ref, parent, parentObject3D)
+    }
+  }}
+/>
+```
+
+#### `attach={false}`
+
+When passing `false` as the attach prop, the component will not be attached to
```
joshwashywash, line 116:
```suggestion
If `false` is passed to the `attach` prop, the component will not be automatically attached to
```

apps/docs/src/content/learn/advanced/migration-guides.mdx
```diff
@@ -81,6 +81,64 @@ to the component you wish to forward events to.
...
+<T
+  is={mesh}
+  attach={false}
+/>
+```
+
+#### `attach={object3D}`
+
+When passing an object3D as the attach prop, the component will be attached to
+the provided object, essentially acting as a portal.
```
joshwashywash, line 130:
```suggestion
If an object3D instance is passed to the `attach` prop, the component will be attached to the instance, essentially acting as a portal.
```

apps/docs/src/content/reference/core/t.mdx
```diff
@@ -297,46 +297,71 @@ The following attaches a material to the material property of a mesh and a geome
...
       return () => {
         parent.shadow.camera = null
       }
     }}
   />
-</T>
+</T.DirectionalLight>
+```
+
+- You may also pass an `Object3D` to the `attach` prop. This allows you to
```
joshwashywash, line 339:
```suggestion
- You may also pass an object3D instance to the `attach` prop. This allows you to
```

joshwashywash (approved):
i like it. 👍🏻 

i just made some suggestions for wording/phrasing in the docs.

one thing i think we could do eventually is go through and normalize how we refer to `T` objects/components. i think we use _component_ and _object_ interchangeably. i propose _object_ is for anything that is created in the script tag; anything "threejs-y" and _component_ is for things in the markup, mostly `T` components.

grischaerbe:
Thanks for your review! ❤️ 

> one thing i think we could do eventually is go through and normalize how we refer to `T` objects/components. i think we use _component_ and _object_ interchangeably. i propose _object_ is for anything that is created in the script tag; anything "threejs-y" and _component_ is for things in the markup, mostly `T` components.

I agree, i find it hard to clearly communicate the difference between component and object instance or even find the correct and most distinguishable wording for that. Any ideas?

PR #1104, opened 2024-09-20
Docs remake

DefinitelyMaybe:
~Still need to merge content 😄 _~

~Content is in but needs to be merged with some recent changes.~

~Build working locally. Will check if just vercel deploy needing to be re-initializing~. 

Looks like the vercel deploy will need to be re-initialized as a working copy of this PR is available [here](https://threlte-docs-shell.vercel.app/)

~Examples needing some TLC~ Merging next branch as lots of examples have been updated

On hold till after threlte 8


DefinitelyMaybe:
Well I did get it to build on me machine 😅 however this could also be a similar issue to what I experienced in my own deployment earlier - literally deleted the vercel project and re-initialized it with the same src and it worked again... lets cross that bridge later

DefinitelyMaybe:
Will redo from current main instead of trying to work through all the merge conflicts

PR #1105, opened 2024-09-21
[next] <Detailed>

joshwashywash:
Named after drei's [\<Detailed\> component](https://drei.docs.pmnd.rs/performances/detailed)

Differs from drei's in that you specify the distance for the mesh as a prop. drei's requires that you pass an array of distances to the parent. I thought it might be annoying to keep the children and distances array in sync.

```svelte
<Detailed>
  <T.Mesh />
  <T.Mesh distance={5} />
</Detailed>
```

The first child mesh is will be visible until the camera is 5 units away from the mesh. At that point, the first mesh is swapped out for the second.

If no distance is provided, as in the example above, it will default to 0.

Individual meshes can define their own hysteresis values. It is an optional prop and defaults to 0 if it is not provided. Details on what hysteresis is can be fount in [three's LOD docs](https://threejs.org/docs/#api/en/objects/LOD).

---

Note that although i haven't removed the LOD example, it can definitely be removed if this gets merged because it'll be redundant.

### DIFF
--- a/.changeset/little-pants-cover.md
+++ b/.changeset/little-pants-cover.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": minor
+"@threlte/docs": minor
+---
+
+adds a <Detailed> component that can be used for displaying different level-of-detail for scene objects
--- a/apps/docs/src/content/reference/extras/detailed.mdx
+++ b/apps/docs/src/content/reference/extras/detailed.mdx
@@ -19,7 +19,7 @@ Notice how as the mesh's distance from the camera increases, a mesh with less ve
   creation tools.
 </Tip>
 
-Children of <Detailed> accept a distance prop. This prop determines the visibility of each child based on its distance from the camera. If not specified, distance defaults to 0.
+Children of `<Detailed>` accept a distance prop. This prop determines the visibility of each child based on its distance from the camera. If not specified, distance defaults to 0.
 
 ```svelte
 <Detailed>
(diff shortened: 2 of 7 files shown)


joshwashywash:
I changed the api to be more flexible. it could be argued that the previous design:

```svelte
<Detailed distances={[0, 5, 10]}>
  <T.Mesh />
  <T.Mesh />
  <T.Mesh />
</Detailed>
```

is simpler but it's definitely not as flexible and may be troublesome to keep `distances` and the children in sync. You also don't get to set individual hysteresis with the above design.

DefinitelyMaybe:
good stuff 👍 

I'm happy enough with the API. I'm sure folks could go either way and perhaps trialing it out this way will help converge towards a good default in time 😄 

From here, add a changeset and then I think we're good to go

joshwashywash:
Okey dokey. I’m gonna do one more cleanup for the docs tomorrow and add a changeset and we’ll be good to go 😎

joshwashywash:
@DefinitelyMaybe do you think we should get rid of the LOD example [here](https://threlte.xyz/docs/examples/geometry/lod) as well? this example is more or less redundant now.

joshwashywash:
changeset added.

i also tweaked the example to be more "automatic". rather than having to scroll the camera to see the lod changes, the lod is now moved along a sine wave to and from the camera.

DefinitelyMaybe:
Yes please to another PR removing the LOD example on the next branch 👍 

haha my bad, I didn't catch something but no biggie - docs doesn't need a changeset, its just for the `packages/*` stuff

joshwashywash:
ah okay. i'll remember that in the future 👍🏻

PR #1106, opened 2024-09-24
[next] <BakeShadows>

joshwashywash:
adds a `<BakeShadows>` component which can be used to improve performance by "freezing" shadows. https://drei.docs.pmnd.rs/performances/bake-shadows

When inside a suspense boundary, `<BakeShadows>` will wait until the boundary has unsuspended before freezing shadows. See docs for an example.

### DIFF
--- a/.changeset/metal-elephants-perform.md
+++ b/.changeset/metal-elephants-perform.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": patch
+"@threlte/docs": patch
+---
+
+Adds a `<BakeShadows>` component that "freezes" shadows for scenes where shadows only need to be calculated once or very infrequently.'
--- a/.changeset/metal-elephants-perform.md
+++ b/.changeset/metal-elephants-perform.md
@@ -1,6 +1,5 @@
 ---
 "@threlte/extras": patch
-"@threlte/docs": patch
 ---
 
 Adds a `<BakeShadows>` component that "freezes" shadows for scenes where shadows only need to be calculated once or very infrequently.'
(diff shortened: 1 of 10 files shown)


DefinitelyMaybe:
Cool cool. changeset and then good to go 👍

joshwashywash:
i was thinking @DefinitelyMaybe should we deviate from drei here and call this component `<FreezeShadows>`? [Baking](https://en.wikipedia.org/wiki/Texture_mapping#Baking) is a common term used in graphics but i don't think it's the right word to use here. "Freezing" indicates a little more clearly what the component is actually doing.

DefinitelyMaybe:
Hmm it is an opportunity to be more concise. There's a bunch of different options. One is that keeping BakedShadows means folks from R3F will know what it is. Another is asking the three.js community what they would call it. Could also just called it whatever feels like it fits best, chatgpt gave me a couple of others like: StaticShadows, LockedShadows and FixedShadows.

In short, no right answer here. I'm tempted by keeping it with BakedShadows for ease of any R3F folks that try Threlte out but not firm on that. Not sure how the rest of the team feels, if you don't mind it taking longer we could wait for their input?

joshwashywash:
> Hmm it is an opportunity to be more concise. There's a bunch of different options. One is that keeping BakedShadows means folks from R3F will know what it is. Another is asking the three.js community what they would call it. Could also just called it whatever feels like it fits best, chatgpt gave me a couple of others like: StaticShadows, LockedShadows and FixedShadows.
> 
> In short, no right answer here. I'm tempted by keeping it with BakedShadows for ease of any R3F folks that try Threlte out but not firm on that. Not sure how the rest of the team feels, if you don't mind it taking longer we could wait for their input?

yeah I agree. R3F is like the standard for all this kind of stuff so we should probably stay aligned with them.

I merged the __next__ brinch into this one and cleaned up the docs once last time and added a changeset. Should be all set now.

.changeset/metal-elephants-perform.md
```diff
@@ -0,0 +1,6 @@
+---
+"@threlte/extras": patch
+"@threlte/docs": patch
```
DefinitelyMaybe, line 3:
```suggestion
````

DefinitelyMaybe (approved):

PR #1110, opened 2024-09-29
[next] meshBounds

joshwashywash:
port of drei's [meshBounds](https://drei.docs.pmnd.rs/performances/mesh-bounds)

a couple of things i'm unsure of

1. if you look at the implementation, `useMeshBounds` isn't necessarily a hook but i matched the syntax and api of hooks so that it keeps the same _feel_. the `useMeshBounds` is just a function that returns a function. the user would never see this but i think it's probably best to keep the same look and feel as the other hooks. so instead of having an export like this:

```typescript
export const meshBounds = () => {}
```

it's this:

```typescript
export const useMeshBounds = () => () => {}
```

and it's meant to be called like a hook

```svelte
<script>
  const meshBounds = useMeshBounds();
</script>

<T.Mesh raycast={meshBounds}>
  <T.BoxGeometry />
</T.Mesh>
```

2. drei's implementation checks that a material is defined on the object and exits early if there is not one defined - https://github.com/pmndrs/drei/blob/71864076c35c3b4f201cd9518da873eff7dc90a1/src/core/meshBounds.tsx#L12 and if you look at three's default raycast function for meshes - https://github.com/mrdoob/three.js/blob/37d6f280a5cd642e801469bb048f52300d31258e/src/objects/Mesh.js#L148 they do the same thing. i'm not sure why they're checking if the mesh has a material. it seems like it wouldn't matter for raycast calculations but if someone has insight on that, i'd like to know.

3. it's difficult to quantize how much more performant this `useMeshBounds` is so i'm wondering if including an example that better demonstrates its effect is worth diving into. at the moment i don't really have a good idea in mind but this is something that can come later if there's enough of an interest in it.

4. lastly concerning the "exists" check in the loop in the example's `Scene.svelte`, i had to do this to satisfy a typescript rule and svelte 5's warning that you can't update state in a loop (paraphrasing). it looks kinda strange and i'm wondering if people won't understand why that check is there. essentially array access could give you back something that is undefined and you're supposed to check for it. i'm wondering if we should leave that part out and just let the typescript warning persist. the ts rule at play here is `noUncheckedIndexAccess`

[here](https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&target=99&jsx=0#code/C4TwDgpgBAKgTgSzAGwgHhgPigXigbRgBpYSYBdAbgChRIoAFAewGcFgEmA7XWRFdFwCuAWwBGEOJhq1w0ALIQWACwBCTIVwAmLXgG9qUKGFbtOXAFyNTHbjQC+1agGNuLYFDEbtLK4pXqmjr45Lz4hlB6xjbmVvgAzCQATCQAjKH2RBFRJmy2lgSpySTxGdRUTq5c7sZWzHnmYQAMJC1QTRXUAGZMcFAAFKgeCLxtyLxeQSwAdKhcAObAypRQI2hQyCsjANR4qQCUkREA9MdQXEwAqlzOyhDOANYQWgCS2hAAHgCCzs5KughdMAhHAuM8oNwIpMfPgEORprkzNxeGAZEZTlAWEwoAB3aDzJjAYAAQygWmxS0BESqNWhWgm3mCcJoRhpHk+gOAujwdKgAEIcHgghAuggwVoWasugMOe4WIcDEYjHSETFkXhURFHPYgA)'s a simplified version of what's going on:


DefinitelyMaybe (approved):
Even though at first glance I was wondering why would there be this function/component, it has actually taught me more about three.js 😅

Over time I think this function can help with a page talking about performance and the ways that can be achieved. @jerzakm has also been working on some performance related components so will be interesting to see where the rubber-meets-the-road, so to speak, like in what use-cases will this function suffice vs no actually you need a BVH algorithm and then maybe it'll work 🤷

grischaerbe:
Thanks @joshwashywash for your contribution! Looks good! I'm somewhat undecided on whether it should be a hook or just a regular function and I guess a regular function is what I would prefer.
I think it could be interesting to add an example that uses Threlte's plugin interface to add the `meshBounds` function to every object in the scene or adds that via a prop, like so:

```ts
const meshBounds = useMeshBounds()

injectPlugin('meshBounds', ({ ref, props }) => {
  if (ref instanceof THREE.Mesh && 'raycastMeshBounds' in props && !!raycastMeshBounds) {
    ref.raycast = meshBounds
  }
})
```

```svelte
<T.Mesh raycastMeshBounds />
```

joshwashywash:
> Thanks @joshwashywash for your contribution! Looks good! I'm somewhat undecided on whether it should be a hook or just a regular function and I guess a regular function is what I would prefer. I think it could be interesting to add an example that uses Threlte's plugin interface to add the `meshBounds` function to every object in the scene or adds that via a prop, like so:
> 
> ```ts
> const meshBounds = useMeshBounds()
> 
> injectPlugin('meshBounds', ({ ref, props }) => {
>   if (ref instanceof THREE.Mesh && 'raycastMeshBounds' in props && !!raycastMeshBounds) {
>     ref.raycast = meshBounds
>   }
> })
> ```
> 
> ```svelte
> <T.Mesh raycastMeshBounds />
> ```

yeah i think that sounds like a good idea. i kinda like the explicitness of `raycast={meshBounds}` but it would be annoying to have to apply it per object so i think a plugin example like that would be nice to have. i'm curious how it would behave. I don't mind if you add that to the pr and make any tweaks that you think are necessary.

grischaerbe (approved):
🚀

PR #1114, opened 2024-10-13
add MarchingCubes "lava lamp" example

joshwashywash:
based on the threejs example example [here](https://threejs.org/examples/#webgl_marchingcubes). 

[drei makes this a component](https://drei.docs.pmnd.rs/abstractions/marching-cubes) but in my opinion it's too restrictive to be a component and works best as an example.

i elaborate a little [here](https://discord.com/channels/985983540804091964/986233334747254854/1294518941833625610)

i tried to make the example pretty comprehensive but the MarchingCubes api is a little _strange_ in some places.

~~one thing i'd like to clear is the flickering that occurs when adjusting the resolution slider but i'm not sure how to achieve that in the current design.~~

Let me know if this needs a changeset and i'll add one.


joshwashywash:
ok i thought of a way to really make this simpler and solve the problem with flickering. will push changes soon.

DefinitelyMaybe (approved):

PR #1118, opened 2024-10-13
convert &lt; and &gt; characters to < and > in search results

### LINKED ISSUE #1024: Search doesn't show < and > correctly
Search command palette on documentation website shows html tags instead of < and >

See: 
threlte

joshwashywash:
closes #1024

to test, open the search modal and type __text__. verify that __\<Text>__ is shown and not __\&lt;Text\&gt;__.

### DIFF
--- a/apps/docs/src/components/Search/Markprompt.tsx
+++ b/apps/docs/src/components/Search/Markprompt.tsx
@@ -51,6 +51,11 @@ function Search({ visible }: { visible: boolean }) {
               visible: false
             }}
             search={{
+              getTitle(ref) {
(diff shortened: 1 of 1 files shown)


DefinitelyMaybe:
i mean yes but also there's #1104

joshwashywash:
ah you're right. i'll close this in favor of that ticket.

DefinitelyMaybe:
hmm @joshwashywash maybe we should just merge this?

joshwashywash:
yeah i think it's small enough not to really cause any disruption with #1104

PR #1120, opened 2024-10-14
Camera-Controls update

DefinitelyMaybe:
expand that pane! previously it was stuck on false

### DIFF
--- a/apps/docs/src/examples/camera/camera-controls/App.svelte
+++ b/apps/docs/src/examples/camera/camera-controls/App.svelte
@@ -15,7 +15,6 @@
 <Pane
   title="Camera Controls"
   position="fixed"
-  expanded={false}
 >
   <Button
     title="rotate theta 45deg"


kitschpatrol:
Hey! I'm the author of Svelte Tweakpane UI and I've noticed this issue in some of the Threlte examples.

It came up on my end in an [issue here](https://github.com/kitschpatrol/svelte-tweakpane-ui/issues/4) a while back, and I found that I can't handle the case of passing a  boolean literal directly to the prop without permanently "locking" the expand / collapse state of certain pane types.

If you want a pane to be initially collapsed but still be user-openable...

Instead of this:
```svelte
<script lang="ts">
	import { Pane } from 'svelte-tweakpane-ui';
</script>

<Pane expanded={false}>
```

Do this:
```svelte
<script lang="ts">
	import { Pane } from 'svelte-tweakpane-ui';
	let expanded = false;
</script>

<Pane bind:expanded>
```

Not thrilled to require a work-around, but can't dig further at the moment.

kitschpatrol:
Actually I can open a quick PR to fix this everywhere.

kitschpatrol:
I think I caught them all here:
https://github.com/threlte/threlte/pull/1121

DefinitelyMaybe:
closing in favor of #1121

PR #1121, opened 2024-10-14
Tweakpane expansion work-around

kitschpatrol:
This PR prevents Tweakpane `<Pane>` components from being locked in an expanded or collapsed state. This was affecting several examples and documentation pages, I'd guess the original authors intended to define the initial state of a pane's expansion vs. its perpetual state.

`expanded` might be a cleaner name for the bound variable and could simplify the binding site, but with all the things going on in the Threlte examples `paneExpanded` seemed more distinctive.

This is related to https://github.com/kitschpatrol/svelte-tweakpane-ui/issues/4. Pardon the need for a work-around, I'll ping back here if I figure out a more elegant fix in the `svelte-tweakpane-ui` library itself at a later date.

### DIFF
--- a/apps/docs/src/examples/camera/camera-controls/App.svelte
+++ b/apps/docs/src/examples/camera/camera-controls/App.svelte
@@ -6,6 +6,7 @@
   import { DEG2RAD } from 'three/src/math/MathUtils.js'
 
   let camera
+  let paneExpanded = false
 
   $: if ($cameraControls) {
     camera = $cameraControls._camera
@@ -15,7 +16,7 @@
 <Pane
   title="Camera Controls"
   position="fixed"
-  expanded={false}
+  bind:expanded={paneExpanded}
 >
   <Button
     title="rotate theta 45deg"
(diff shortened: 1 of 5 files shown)


DefinitelyMaybe (approved):

DefinitelyMaybe:
haha omgosh yeah thats a much better solution! No worries at all and thank you so much for making this PR 👍 

I don't consider this a work-around, instead my own laziness at the initial implementation. Really `expanded={false}` not changing on click is technically correct and well implemented 😄  `bind:expanded={paneExpanded}` should've been my first thought.

kitschpatrol:
Thank you! And thanks for all your work on Threlte!

Just FYI, the work-around is no longer required in projects compiled with Svelte 5 and Svelte Tweakpane UI >=1.4.0, but it's probably a good idea to keep the explicitly-bound `expanded` approach in the examples for Svelte 4 support.

(Even though "legacy" Svelte components are supposed to compile transparently in Svelte 5, I've found some "breaking" [bug fixes](https://github.com/sveltejs/svelte/issues/4265) that can subtly change how reactivity works... in this case to our benefit. 😅)

PR #1123, opened 2024-10-15
[next] <Resize>

joshwashywash:
adds a port of drei's [Resize](https://drei.docs.pmnd.rs/staging/resize) component. which is a simple component that will scale down a group by its maximum dimension.

differs slightly from drei's in that you pass an `axis` prop to choose an axis instead of passing `height`, `width`, or `depth` boolean props which is ambiguous if you pass more than one. making `axis` a union only allows for one of these options to be passed.

edit:

i don't think this is reactive to children by design.

what i mean is if the group's children change, it won't recalculate and rescale things. it's only effective over the props.


DefinitelyMaybe:
Hey nice work with the port 👍 

I'm think I might be misunderstanding something about the component. I've adjust the scene to demonstrate, why is the larger red box smaller than the hotpink ones?

joshwashywash:
> Hey nice work with the port 👍
> 
> I'm think I might be misunderstanding something about the component. I've adjust the scene to demonstrate, why is the larger red box smaller than the hotpink ones?

this code?

```svelte
<Resize position.x={-5}>
  <T.Mesh>
    <T.BoxGeometry args={[5, 5, 5]} />
    <T.MeshStandardMaterial color="red" />
  </T.Mesh>
  <T.Mesh position.z={5}>
    <T.BoxGeometry />
    <T.MeshStandardMaterial color="red" />
  </T.Mesh>
</Resize>
```

so the bounding box that gets calculated for that group ends up being a box like this:

```ts
{
  max: {x:  2.5, y:  2.5, z:  5.5},
  min: {x: -2.5, y: -2.5, z: -2.5},
  // ...
}
```

the max dimension is z with a length of `5.5 - (-2.5) = 8.` then the entire group gets scaled by `1/8`. the group's z gets a length of 1 since `8/8 = 1`. that's expected. the big box and the little box and the distance between them must all fit into this length of 1.

it's a little hard to explain but i made a screenshot for it with a box helper around the group

resize-thing

---

the docs may need to explain this better or maybe the example can demonstrate it better.

DefinitelyMaybe:
Ah that makes more sense. It's not about each individual object being resized instead it's the `Box` created around the objects within the component, yes?

grischaerbe:
Thanks @joshwashywash for another excellent contribution 🎉 
I have some remarks or suggestions about the API design and the docs (wording) and wanted to have your opinion on that.

- The docs start with "Scales down a group of objects", but this component may also scale up, right? So essentially it "normalizes" the scale on a particular axis in world space, right?
- It always scales to world space `1`, right? So if I want to scale it to let's say `3`, I would need to wrap it again in a `<T.Group scale={3}>`. I'd love to be able to do that on that component too to reduce the scene graph depth.
- The docs tell me the respective axes of `"width"`, `"height"`, `"depth"`, is there a reason `dimension` is not just `"x" | "y" | "z"`? This would make looking up the "translation" obsolete.

I imagine an API like this would also be possible:

```svelte
<Resize x={1} />
<Resize x={3} />

<!-- Advanced, based on CSS object-fit -->
<Resize x={3} y={1} fit="contain" />
<Resize x={3} y={1} fit="cover" />
```

I see two possibilities:
- Per TS only allow a single dimension and error if more than one is provided. Error if no axis is provided. No `object-fit` correspondence, just scaling to fit the provided axis. If no axis is provided, nothing happens.
- Multiple axes may be provided. Two possible `fit` properties allow to either `"contain"` or `"cover"` the axes. If no axis is provided, nothing happens.

I know the implementation is a bit more challenging with this feature, but would make the component ultimately more convenient to use, what do you think?

Bonus: What do you think about reusing the logic that the `<Align>` component uses to [recalculate whenever a child `<T>` component mounts](https://threlte-cry0z8dp8-threlte.vercel.app/docs/reference/extras/align#auto)?

joshwashywash:
> [!Warning]
> i was trying to get the formatting of this comment to look good. hopefully it's not too hard to parse.

---

> * The docs start with "Scales down a group of objects", but this component may also scale up, right? So essentially it "normalizes" the scale on a particular axis in world space, right?

oh yeah you're right. if there's a bounding box, for example 

```ts
// let original =
{
  x: 0.5,
  y: 0.6,
  z: 0.1
}
```

that will get scaled up to

```ts
// let normalized =
{
  x: 0.5 / 0.6,
  y: 0.6 / 0.6,
  z: 0.1 / 0.6
}
// => 
{
  x: 0.83,
  y: 1,
  z: 0.67
}
```

i can fix the wording on that.

> * It always scales to world space `1`, right? So if I want to scale it to let's say `3`, I would need to wrap it again in a `<T.Group scale={3}>`. I'd love to be able to do that on that component too to reduce the scene graph depth.

you can set the scale of the `Resize`

```ts
<Resize scale={3}>
</Resize>
```

This would scale the children of <Resize /> and then scale them up so that the max axis is 3 units, the other two have been normalized 0 -> 1 and then multiplied by 3 and their proportions are the same.

using the same units from above

```ts
// let scaledBy3 = 3 * normalized =
{
  x: 3 * 0.83,
  y: 3 * 1,
  z: 3 * 0.67
}
// =>
{
  x: 2.49
  y: 3,
  z: 2.01
}
```

the proportions of `scaledBy3`, `normalized` and the `original` are all the same give or take some rounding errors.

is that what you mean?

> * The docs tell me the respective axes of `"width"`, `"height"`, `"depth"`, is there a reason `dimension` is not just `"x" | "y" | "z"`? This would make looking up the "translation" obsolete.

oh yeah that's a lot nicer. i can get that fix in as well.

> I imagine an API like this would also be possible:
> 
> ```svelte
> <Resize x={1} />
> <Resize x={3} />

is this the same as

```svelte
<Resize scale.x={1} />
<Resize scale.x={3} />
```

or am i missing something?

i think it would stretch things out if we did it using `scale.x` this way though since the scaling is not applied to each axis.

> 
> <Resize x={3} y={1} fit="contain" />
> <Resize x={3} y={1} fit="cover" />
> ```
> 
> I see two possibilities:
> 
> * Per TS only allow a single dimension and error if more than one is provided. Error if no axis is provided. No `object-fit` correspondence, just scaling to fit the provided axis. If no axis is provided, nothing happens.
> * Multiple axes may be provided. Two possible `fit` properties allow to either `"contain"` or `"cover"` the axes. If no axis is provided, nothing happens.
> 
> I know the implementation is a bit more challenging with this feature, but would make the component ultimately more convenient to use, what do you think?

i think i like the second one better but i think i need to see it in practice to really have a strong opinion.

> Bonus: What do you think about reusing the logic that the `<Align>` component uses to [recalculate whenever a child `<T>` component mounts](https://threlte-cry0z8dp8-threlte.vercel.app/docs/reference/extras/align#auto)?
> 
> ```svelte
> <Resize x={1} auto />
> ```

i thought about that too, it would be nice if it automatically rescaled when the children changed. i think that would be a good feature to have but it makes me wonder if the `auto` prop changes, is there a way to remove the plugin?

in `Align.svelte` it's

```ts
  if (auto) injectAlignPlugin(scheduleAligning)
```

but if auto switches from true to false, is there a way to remove the plugin or disable it? the `check if auto is true` logic might have to live in the plugin itself. maybe? 🤔 

or if there was a way to do something like:

```ts
$effect(() => {
  if (auto) {
    inject()
  } else {
    uninject()
  } 
})
```

i don't think anyone would want to turn off and on `auto` so maybe we just let it be set once but its still something to think about.

or something like that.

joshwashywash:
oh yeah, i was considering renaming the `dimension` prop to `axis` but i think we should keep the name `dimension` since this component deals with dimensioning and scale.

you know, i'm kinda wondering why this prop is even offered. i can't think of a use case where you wouldn't want to scale by a dimension other than the maximum. i think it's simple feature to offer but i'm worried it doesn't really have a purpose and i don't want people to think they need to use it or need to consider it.

grischaerbe:
So, here's my proposal:

```svelte
<!-- Normalized respective to max dimension -->
<Resize />

<!-- Normalized respective to max dimension, scaled up by a factor of 3 -->
<Resize scale={3} />

<!-- Normalized respective to x axis -->
<Resize axis="x" />

<!-- Automatically recalculates whenever child components mount/unmount -->
<Resize auto />
```

So basically very close to your first proposal, but 
- `dimension` → `axis`
- type of `dimension`: `"width" | "height" | "depth"` → `"x" | "y" | "z"`
- Added `auto` support

What do you think?

joshwashywash:
yeah i like all those ideas. `dimension` -> `axis` and the `auto` support are the only two missing. dimension to axis will be a quick and easy thing. i'll work on `auto` support and use `Align` as a guide.

joshwashywash:
i'l also tweak the first example to show off the axis prop.

joshwashywash:
@DefinitelyMaybe @grischaerbe tell me what you all think of the revised example. i think it more clearly shows `<Resize>`'s purpose. i'm not sure if we want to keep the camera repositioning when _resize_ is unchecked. i think it's cool but probably not necessary.

DefinitelyMaybe:
Much better! I like the camera repositioning 👍

joshwashywash:
@grischaerbe i added a autoresizing thing to the component but i'm not quite sure how to test it properly. i wrote a small example and put it at the top of the Resize mdx page. i don't intend for it to be included in the final merge but i put it there if there's anything you want to test.

i'm also not sure how $state is handled in a callback so i think i have it coded correctly but i'd like you to comb over it and see if there's anything that may be done incorrectly.

i also added a `onresize` callback prop that can be used the same way `onalign` is used in `Align.svelte`.

the only thing missing i think is a binding to the resize function so that you can bind to and force resizing. something like this:

```svelte
<script>
let resize = $state()
</script>

<T.Mesh
  onclick={() => {
    resize?.()
  }}
>
</T.Mesh>

<Resize bind:resize>
</Resize>
```

joshwashywash:
i like what you came up with. I kinda think we should have an example of normalizing multiple objects like with the "fox, duck, and flower" example from before. I'm just concerned that people will think that they need to use <Resize> with <Align>. I kinda wanted an example similar to this one https://codesandbox.io/p/sandbox/6yg0i3 but i liked the idea of showing how the resize made everything the same size and being able to toggle it on and off.

Other than that, I think this one is good to go. Let me know what you think.

joshwashywash:
oh it looks like there's some additional props that aren't documented. i'll take care of that real quick

joshwashywash:
@grischaerbe  i made some tweaks to the code. the main thing was to make  `stage` reactive.

one question involves the `resize` function that is made available to the parent.

right now we have two way to access it - binding to it or using it through snippet props

## through binding

```html
<script>
  let resize
</script>

<Resize bind:resize />
```

`<Resize>` will set `resize` but the parent has the ability to set it to something else at a later time.

```html
<script>
  let resize
  
  setTimeout(() => {
    resize = () => console.log('hello')
  }, 5000);
</script>

<Resize bind:resize />
```

the resizing in the component still refers to `scheduleResizing` so that won't break but should the ability to set `resize` after it has been set within `<Resize>` be a concern?


## through snippet props

in `<Resize>` we have:

```
{@render children?.({ref, resize: scheduleResizing })}
```

so here you can get the resize out like so

```html
<script>
  let outsiteResize
</script>

<Resize>
  {#snippet children({resize})}
    <T.Group on:create={() => {
      outsiteResize = resize
    }}>
    </T.Group>
  {/snippet}
</Resize>
```

that's very weird but also possible

---

my question is should we provide resize through children like that? you can already bind to it, you don't really need to access it inside the snippet block, or is there a situation where having it through snippet props would be useful? Is there a case that isn't already handled by getting it through the bindable prop? If not then I think we can safely remove it from the children snippet.

grischaerbe:
> my question is should we provide resize through children like that?

I think it's okay to offer both. I just had a second look at component exports, since `resize` should be a read-only bindable, Svelte calls this a component export. I think we do not declare that on `$props()` but just `export const align = scheduleAligning`. Here are [the docs](https://svelte.dev/docs/svelte/v5-migration-guide#Breaking-changes-in-runes-mode-Bindings-to-component-exports-are-not-allowed). The types are different then though, we need to use the new `Component` type then. This is all still new to me and this PR is a good testbed to try out the new mechanics, since pretty much everything that Svelte offers is used :)

grischaerbe:
Concerning the example: I agree, the use case of downloaded models being in all shapes and sizes is a totally valid one. Maybe we can make it without the complexity of the camera movement, i.e. less exaggerated? I'll revert this change 👍

joshwashywash:
okay i think it's finally done. i can't run the examples cuz i think stackblitz is using the `module` prop and my local astro is using `context="module"`. I think that's the issue. Feel free to fix up anything that may have broken 🙏🏻 . Otherwise we're ready to go!

PR #1131, opened 2024-10-17
[next] `<LinearGradientTexture>` and `<RadialGradientTexture>`

joshwashywash:
another port from drei

drei combines these into one [component](https://drei.docs.pmnd.rs/abstractions/gradient-texture) but i wanted to keep them separate to avoid having props that aren't always used. in other words, if you combine them then you have extra logic in the component for checking what "type" of gradient you are and it adds ambiguity. for example, drei's allows you to pass in props that are only used for the radial version when you might only be using the linear version.

there's a lot of code but there's nothing overly cryptic.

i decided to extract some of the shared logic and normally there would only be .svelte files that are exported by @threlte/extras but this time around i'm exporting two additional type -> `GradientStop` and `RadialGradientOuterRadius`. these are exported to make it easier to type things that will eventually get passed as props to the gradient components. the downside to this is that these types are opaque when you try to view them in an editor. you have to jump to their definition to see what they truly are.

i tried to make the examples very comprehensive. let me know if there's anything weird or off about them.


joshwashywash:
you can actually get some really cool textures out of the radial gradient one

Screenshot 2024-10-17 at 4 38 26 PM

Screenshot 2024-10-17 at 4 36 45 PM

Screenshot 2024-10-17 at 4 37 37 PM

grischaerbe (approved):
This is awesome! Well documented, good example, well built! 🎉

PR #1139, opened 2024-10-24
fix: correct type for `scale` in `PerfMonitor`

FoHoOV:
Scale type was harcoded to `1` which is changed into a number.


DefinitelyMaybe:
LGTM! thank you

PR #1142, opened 2024-10-26
Fix grammar on Services Page

joshpinto6:
I know this is an incredibly tiny PR change but this issue is definitely relevant when trying to market your services to customers and it stood out to me right away when browsing around the threlte website.

Really awesome project though and I can't wait to use it! 🤘


DefinitelyMaybe (approved):

DefinitelyMaybe:
good catch! thank you 👍

PR #1144, opened 2024-10-27
[next] <CubeCamera>

joshwashywash:
port of [drei's cube camera](https://drei.docs.pmnd.rs/cameras/cube-camera)

## differences

mostly everything is covered in the docs mdx file but here's a summary with some extra stuff.

passing both the render target and the cube camera to the `children` snippet. not sure if the cube camera or the entire render target will be needed in the snippet block since most of the time you'll just use `renderTarget.texture` but it's there if anyone needs it.

because these are exposed, a `restart` function is also available. i'm not sure about the name of this function. it is used to force a render to the target if ever necessary - for example doing something other than updating the props that are passed to `<CubeCamera>`.

I don't know if drei's will rerender if the `frames` prop is not `Infinity` and `background`, `fog`, or any of the other props are updated dynamically. This one will always restart the frame counter when any of the component's props change.

if the user wants to use the current background or fog of the scene, the prop values for these are 'auto'. you're not required to set these to 'auto' it's just nice and explicit to associate them with a value.

## considerations

I think the only thing that may be requested is to be able to listen for when the render task has started or when the renderer has rendered for the last time. something like `onrenderfinish` or `onrenderstart`.


grischaerbe:
Thanks @joshwashywash for this excellent contribution!
I refactored it slightly so that the render target and the camera aren't derived values but are mutated in side effects so that they can be exported. Otherwise, you'd have to export functions that return the current value, which is a bit odd in my opinion. I think exports come in handy when you want the position of the camera to be different from the position you'd want to use the texture at. Let me know what you think!

I like your idea of `onrenderstart` and `onrenderfinish` callbacks!

joshwashywash:
alright i'll work on adding those event callbacks and add them to the docs as well.

i really like this "export functions and props" on the component thing.

```export const restart = () => {}```

that makes it super easy to bind to it and call it from the outside if ever needed.

the only bit i'm not sure of is this:

```typescript
  export const camera = new CubeCamera(near, far, renderTarget)
  observe.pre(
    () => [near, far],
    () => {
      camera.children.forEach((child) => {
        if (isInstanceOf(child, 'PerspectiveCamera')) {
          child.far = far
          child.near = near
          child.updateProjectionMatrix()
        }
      })
    }
  )
```

but i think while typing this up, i just figured it out though. CubeCamera has a PerspectiveCamera as a child and we're updating that.

anyway i'll get those changes in and ping you when ready.

grischaerbe:
Yep, a `CubeCamera` is basically 6 `PerspectiveCamera`s in a particular setup. They're [initialized in the constructor](https://github.com/mrdoob/three.js/blob/master/src/cameras/CubeCamera.js#L20) 👍
Awesome, we're ready to merge as soon as the events are in 🎉

joshwashywash:
@grischaerbe ready to go i believe.

grischaerbe (approved):
LGTM 🚀

PR #1145, opened 2024-10-28
[next] <AsciiRenderer>

joshwashywash:
port of [drei's AsciiRenderer](https://drei.docs.pmnd.rs/abstractions/ascii-renderer) with more options.

edit:

the `options.color` prop is slow but it could be useful for static renders, At the moment there's no way to stop the underlying rendering task. I'll come up with a plan for that tomorrow.


joshwashywash:
to better state the problem above, the `colors` prop is fine and useful for single renders but it's very slow. there's no way to render once and then stop the task or to run the task at will. we need a way to say 'i want to use colors, but i only need to render one frame or i only need to render when i say render'

joshwashywash:
alright, i think i came up with a mediocre solution to the problem. `<AsciiRenderer>` is a lot like a task so currently it accepts an `autoRender` prop. this prop controls whether the underlying task is paused or not.

Pausing can be useful if you only want to render one frame or render on demand. This is especially useful when using `options.color` since that really slows down the renderer.

It might make more sense to call it `paused` and use it like this

```svelte
<AsciiRenderer paused={someVariable} />
```

i kinda like `paused` more but `autoRender` stays inline with the main threlte task - the one that you get back from `useThrelte`.

```typescript
const { autoRender } = useThrelte();
```

joshwashywash:
we could use a similar approach to how `<CubeCamera>` works. it uses a `frames` prop that defaults to Infinity. If you set the `frame` to 1, it will only render once. It's dynamic on `frames`. The downside is that the task is always running. There's no pausing it from the outside.

I got this technique from drei's components. I wonder if we should make it the standard for these kinds of things.

joshwashywash:
ok i think this one is finally ready to go. maybe one more look over for any typos or weird phrasing in the docs but it does everything i'd want/expect it to do now.

apps/docs/src/content/reference/extras/ascii-renderer.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+If `options.color` is set to `true`, `fgColor` and `bgColor` will be ignored and the corresponding color of the scene will be used for each character.
+
+```svelte
+<AsciiRenderer options={{ color: true }} />
+```
+
+<Tip type="info">
+  Setting `options.color` to `true` really slows down performance but it can be used if you have a
+  static scene or only need to render the scene on demand.
```
DefinitelyMaybe, line 131:
```suggestion
  Setting `options.color` to `true` slows down performance. Using a
  static scene or manually rendering can help.
  ```

File: apps/docs/src/content/reference/extras/ascii-renderer.mdx [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+          default: 'true',
+          description: 'whether to automatically run the render task',
+          name: 'autoRender',
+          required: false,
+          type: 'boolean'
+        },
+        {
+          name: 'characters',
+          default: "' .:-+*=%@#'",
+          description: 'character set used by the renderer. should be sorted from least to most opaque',
```
DefinitelyMaybe, line 45:
```suggestion
          description: 'character set used by the renderer. Sorted left-to-right from least to most opaque',
```

DefinitelyMaybe (commented):
Hahaha the feeling when orbiting a TorusKnot but its ascii  ... my brain took some extra cycles to catch up 😅 

pretty good work here. Couple of edits I hope might improve the docs but otherwise LGTM 👍

PR #1149, opened 2024-10-30
[next] fix: import component instead of using <svelte:self>

joshwashywash:


DefinitelyMaybe:
LGTM 👍

DefinitelyMaybe:
[svelte docs for ref](https://svelte.dev/docs/svelte/legacy-svelte-self)

Issue #565, opened 2023-08-30
Porting drei


DefinitelyMaybe:
> We'd like to acknowledge the work r3f/drei has done for the 3D web graphics community as the abstractions they've created have been hugely popular.

This Github issue is about keeping track of our progress and helping contributors who'd like to join in. The place to get in touch with us is on [discord](https://discord.gg/ZN7HtXxWcd) or in the discussion below. From there check out current [proposals](https://discord.com/channels/985983540804091964/1098301342239432784), double check the list below to read relevant info for the part of the port you'd like to tackle and ask questions in the [contributing channel](https://discord.com/channels/985983540804091964/986233334747254854) when more help/info is needed. I'll be updating this post as things change so we're all on the same page.

Brace yourselves, this will take a fair bit of time, energy and effort. The discussion will evolve, we'll pivot around implementations, we'll laugh cry and argue 😃 😢  but in the end we're here because we think doing this is important 💯 

# Controls

- CameraControls
- FlyControls
- MapControls
- DeviceOrientationControls
- TrackballControls
- ArcballControls
- PointerLockControls [docs](https://threlte.xyz/docs/examples/camera/pointer-lock-controls)
- FirstPersonControls (look into discord/proposals)
- ScrollControls
- PresentationControls
- KeyboardControls
- FaceControls

# Gizmos

- GizmoHelper
- PivotControls
- TransformControls [docs](https://threlte.xyz/docs/reference/extras/transform-controls) (#237 #196)
- Grid [docs](https://threlte.xyz/docs/reference/extras/grid) (#213 #210 #209 #207 #203 #200)
- useHelper

# Abstractions

- Image
- Text [docs](https://threlte.xyz/docs/reference/extras/text) (#220)
- Text3D
- PositionalAudio [docs](https://threlte.xyz/docs/reference/extras/positional-audio) (#248)
- Billboard
- ScreenSpace
- Effects
- GradientTexture
- Edges [docs](https://threlte.xyz/docs/reference/extras/edges) (#230)
- Trail
- Sampler
- Computed Attribute
- Clone
- useAnimations
- MarchingCubes
- Decal
- Svg
- Gltf [docs](https://threlte.xyz/docs/reference/extras/gltf)
- AsciiRenderer

# Shaders

- MeshReflectorMaterial
- MeshWobbleMaterial
- MeshDistortMaterial
- MeshRefractionMaterial
- MeshTransmissionMaterial
- MeshDiscardMaterial
- PointMaterial
[...]

PR #1151, opened 2024-10-30
[next] <ScreenSpace>

joshwashywash:
port of drei's [ScreenSpace](https://drei.docs.pmnd.rs/abstractions/screen-space) component.

pretty simple but the example may not be too impressive.

the only difference here is that you can use any camera and the depth is a positive number to put objects in front of the camera. not sure if in front of the camera should be positive or negative. it probably depends on what you call the prop. `depth` is probably fine but `distance` might be preferred? i kinda like depth though cuz it implies that it's along the z-axis.


grischaerbe:
@joshwashywash Thanks for that port!
I don't know why the drei Version doesn't do that, but I adapted it to also work in situations where `<ScreenSpace>` is pre-transformed as part of a child object. Do you happen to know why they would not do that? So currently, this works by calculating the world position/quaternion but we could also just use `attach` (Pseudocode):

```svelte
<script>
  const { camera } = useThrelte()
</script>

<T.Group attach={$camera} position.z={depth}>
  {@render children?.()}
</T.Group>
```

… and at this point I wonder if it should be a component if it's so simple 😅 Let me know what you think!
Of course, the scene graph is different as the object in screen space is now actually attached to the camera, but that's all.

Also: Please [add changesets](https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md#adding-changesets) to your PRs 🙂

joshwashywash:
> @joshwashywash Thanks for that port! I don't know why the drei Version doesn't do that, but I adapted it to also work in situations where `<ScreenSpace>` is pre-transformed as part of a child object. Do you happen to know why they would not do that? So currently, this works by calculating the world position/quaternion but we could also just use `attach` (Pseudocode):
> 
> ```svelte
> <script>
>   const { camera } = useThrelte()
> </script>
> 
> <T.Group attach={$camera} position.z={depth}>
>   {@render children?.()}
> </T.Group>
> ```
> 
> … and at this point I wonder if it should be a component if it's so simple 😅 Let me know what you think! Of course, the scene graph is different as the object in screen space is now actually attached to the camera, but that's all.
> 
> Also: Please [add changesets](https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md#adding-changesets) to your PRs 🙂

i was thinking about that too. it could just be a child of the camera and make use more use of the scene graph capabilities. I'm not sure why drei decided it was necessary to split it into its own component. i think it's more in-line with what we have to do it with `attach`. So i don't mind tossing this component out but i think it should be crossed out on #565 with a reason explaining the decision to not include it. probably just a link to this discussion.

I'll start adding changesets. For new components should it be a minor bump or a patch?

grischaerbe:
Good point, we're tossing this one. As much as I like composability and orchestration using pre-made components, it's also about maintainability of the library. Thanks so much for your efforts though ❤️

> I'll start adding changesets. For new components should it be a minor bump or a patch?

New components are new features user-facing but non-breaking so minor 👍

Issue #1086, opened 2024-08-14
Copying example verbatim from example shows error


peterolson:
I copied the code verbatim from this page:

https://threlte.xyz/docs/reference/extras/meshline-geometry

but I just see an empty grid rotating with no line

![image](https://github.com/user-attachments/assets/45e23f3d-a42c-4e7a-ae98-906dba5ed0a2)

and I see the following error in my console:

```
root.svelte:21 THREE.WebGLProgram: Shader Error 0 - VALIDATE_STATUS false

Material Name: 
Material Type: ShaderMaterial

Program Info Log: Fragment shader is not compiled.


FRAGMENT

ERROR: 0:211: 'LinearTosRGB' : no matching overloaded function found
ERROR: 0:211: '=' : dimension mismatch
ERROR: 0:211: 'assign' : cannot convert from 'const mediump float' to 'out highp 4-component vector of float'


  206: 
  207: 	if( useDash == 1. ){
  208: 			c.a *= ceil(mod(vCounters + dashOffset, dashArray) - (dashArray * dashRatio));
  209: 	}
  210: 
> 211: 	gl_FragColor = LinearTosRGB(c);
  212: }
  213: 

255
WebGL: INVALID_OPERATION: useProgram: program not valid
root.svelte:21 WebGL: INVALID_OPERATION: drawElements: no valid shader program in use
chunk-GMFGCRRE.js?v=676ebfb3:14301 WebGL: too many errors, no more errors will be reported to the console for this context.
﻿
```

## Steps to reproduce:

Clone
https://github.com/peterolson/asl-pose-svelte4

`npm install` and `npm run dev`

---
Alternatively, press the "StackBlitz" button on https://threlte.xyz/docs/reference/extras/meshline-geometry

jkufa:
Hey @peterolson, I ran into a similar issue today trying out the same code example. I was on the latest version of `three: "^0.168.0"`, but changing the version to `"three": "^0.160.1",` (the same version in @threlte/extras's [package.json](https://github.com/threlte/threlte/blob/main/packages/extras/package.json)) solve it for me.

joshwashywash:
this should be closed when #1097 gets merged. see ticket #1095 for details.

grischaerbe:
This should be fixed, thanks all!

PR #1153, opened 2024-10-31
[next] Meshline fixes

edsunman:
A couple of fixes for MeshLineGeometry/MeshLineMaterial.

**First problem** can be seen in console here:
https://next.threlte.xyz/docs/reference/extras/meshline-geometry

Also reported here: https://github.com/threlte/threlte/issues/1086

I'm not sure why this ever worked because the LinearTosRGB function was not defined. Perhaps in older versions of three there was a LinearTosRGB function in a shader chunk that has been removed? Anyway I have added a function to the fragment shader so it should work with any version of three.

**Second problem** is that the uv buffer attribute is currently not being set properly:

meshline

I think the problem was introduced a few months ago when switching from a custom setXYZW function to three's Buffer Attribute .setXYZW method which works slightly differently.

### DIFF
--- a/.changeset/short-toys-smile.md
+++ b/.changeset/short-toys-smile.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
+Fix issues with MeshLine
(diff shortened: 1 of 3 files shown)


grischaerbe (approved):
Awesome, thanks for the fix!

PR #1163, opened 2024-11-02
Extras types migration

joshwashywash:
i think there's one commit per directory. I tried to fix some warnings/errors along the way but nothing too major. I also added documentation where it was lacking and formatted jsdoc style comments in some areas.

Hopefully nothing broke. i'll check each page later today once the preview gets deployed.


joshwashywash:
i went over all the examples in extras and everything works as expected. i had to fix BakeShadows.

joshwashywash:
@grischaerbe do you get this error? i can't figure out what it is but it shows up on anything that overwrites the default snippet from the second type argument of `Props`. for example Align, CubeCamera, Resize. i thought it was cuz properties weren't optional but making everything optional doesn't make the error go away

Screenshot 2024-11-03 at 5 44 02 PM
 
the type is correct if you look at it though.

Screenshot 2024-11-03 at 5 51 07 PM

This even seems to break with a component like this:

```svelte
<!-- Test.svelte -->
<script lang="ts">
  import type { Snippet } from 'svelte'
  type Props = {
    children?: Snippet<[{ foo: string; bar: number }]>
  }

  let { children }: Props = $props()
</script>

{@render children?.({ foo: 'josh', bar: 100 })}
```
Screenshot 2024-11-03 at 6 01 43 PM

but if you provide all args, it seems to work just not with correct types.

Screenshot 2024-11-03 at 6 02 16 PM

joshwashywash:
the only other question i have is why this

```typescript
Omit<Props<ShaderMaterial>, 'fragmentShader' | 'vertexShader'>
```

and not this?

```typescript
Props<Omit<ShaderMaterial, 'fragmentShader' | 'vertexShader'>>

other than those two things, everything else looks sound. 💪🏻

grischaerbe:
@joshwashywash 

Concerning the snippet types, that seems to work fine in VS Code. Not sure why that doesn't work for you, maybe we should raise an issue on the Svelte side of things?

Concerning the types:
Whatever type argument you pass to `Props<T>` will be used for things such as [the `ref` type](https://github.com/threlte/threlte/blob/7ee3bea51a074d1499d2058a7897d0ae3da81e1c/packages/core/src/lib/components/T/types.ts#L90). So `Props<Omit<ShaderMaterial, 'fragmentShader' | 'vertexShader'>>` means that the `ref` that is being passed around (e.g. via the snippet prop `ref`) is a `ShaderMaterial` that is lacking the properties `fragmentShader` and `vertexShader`, which is not the case since it's a regular `ShaderMaterial`. If you turn that around, `ref` still stays a `ShaderMaterial` but the component doesn't allow receiving the props `'fragmentShader' | 'vertexShader'` 👍

grischaerbe (approved):
Thanks @joshwashywash for this excellent work! LGTM 🚀

Issue #1164, opened 2024-11-02
Replace dependency of `useThrelte` in `useTask` hook with `useSheduler` hook instead


phoenisx:
### Description

I've been working on integrating Threlte in my project to support [Multi Views in ThreeJS](https://threejs.org/examples/?q=multiple#webgl_multiple_elements), but while working on it, I've found some issues with `packages/core/src/lib/hooks/useTask.ts` hook which is tightly coupled to `useThrelte` hook, making it hard to create a custom renderer for cases where Renderer is created independent of Threlte context

### Example

For e.g. let's assume we need to create a custom context pipeline for our project:

```js
createCanvasContext(options); // imported from `@threlte/core`
createCacheContext(); // imported from `@threlte/core`
createDisposalContext(); // imported from `@threlte/core`
createSchedulerContext(options); // imported from `@threlte/core`
createScenesContext(); // custom fragment context, which may or may not contain list of scenes
createCameraContext(); // imported from `@threlte/core`

// custom renderer pipeline, which depends on `useTask` -> `useThrelte` -> `useScene`
// But we haven't created any Scene Context yet, so using `useThrelte` throws an error.
createRendererContext(options);

createUserContext(); // custom context
```

### Solution
Use `useScheduler` instead of `useThrelte` inside `useTask` hook.

phoenisx:
I've already worked on a fix and will raise a PR in the next few days.

grischaerbe:
Done :)

PR #1172, opened 2024-11-03
feat(core): #1164 removes usethrelte dependency in usetask hook

phoenisx:
This PR fixes: #1164 

### Description
Removes dependency on `useThrelte` hook in `useTask` hook, and replaces it with `useScheduler` hook.


grischaerbe (approved):
LGTM 🚀

PR #1181, opened 2024-11-05
Update interactivity.mdx

manegame:
When updating a store, we need to return the object to see the changes


grischaerbe (approved):
Good catch!

PR #1192, opened 2024-11-07
[next] restore ability to disable automatic attaching for gradient textures

joshwashywash:
Restores the ability to disable automatic attaching of these textures to parents. They will automatically attach to the `map` property of their parent but you can now override that with `attach={false}` or any of the other attaches that <T> accepts.

other things

- adds docs for disabling auto-attachment.
- uses `observe` instead of the weird looking

```typescript
$effect(() => {
  /* dependencies */
  depends
  on
  this
  and
  that
  /* effect code that doesn't reference any dependencies */
})
```
- no longer sets the colorspace to the colorspace of the renderer. you can still do that but now you must do it yourself.

### DIFF
--- a/.changeset/wet-badgers-end.md
+++ b/.changeset/wet-badgers-end.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": attach
+---
+
+[next] Restore ability to disable *attaching* for gradient textures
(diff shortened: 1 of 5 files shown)


grischaerbe (approved):
LGTM! 🚀

PR #1193, opened 2024-11-07
[next] Cube camera ref assignment tweaks

joshwashywash:
Other components create the binding ref in the script tag and then do

```svelte
<T is={someThreeObjectInstance} bind:ref />
```

so cube camera should do the same.


michealparks (approved):
LGTM! should this have a changeset for extras?

PR #1197, opened 2024-11-09
[next] useShake example

joshwashywash:
a shake hook inspired by drei's [CameraShake](https://drei.docs.pmnd.rs/cameras/cube-camera).

i think i finally landed on semething that i like - a hook.

> [!IMPORTANT]  
> while i was making this, it started to feel like it would make a good example but now after completion, it feels substantial enough to add to `extras` but i want to know what you all think - should we put `useShake` in `extras`?

Please read the corresponding `.mdx` file. i think its feature-heavy enough to warrant putting it into `extras`.

### DIFF
--- a/apps/docs/src/examples/shake/shake.ts
+++ b/apps/docs/src/examples/shake/shake.ts
@@ -1,0 +1,42 @@
+import type { AxisItem, AxisParams, ShakeParams } from './types'
+import type { Object3D } from 'three'
+import { SimplexNoise } from 'three/examples/jsm/Addons.js'
+import { useTask } from '@threlte/core'
+
+const createAxis = ({ frequency, intensity }: AxisParams = {}): AxisItem => {
+  const noise = new SimplexNoise()
+  return {
+    frequency: frequency ?? 0.1,
+    intensity: intensity ?? 0.1,
+    noise
+  }
+}
+
+export const useShake = (params: ShakeParams = {}) => {
+  const pitch = createAxis(params.pitch)
+  const roll = createAxis(params.roll)
+  const yaw = createAxis(params.yaw)
+
+  let time = 0
+
+  const shake = (ref: Object3D, autoStart = true) => {
+    const { x, y, z } = ref.rotation
+    return useTask(
+      (delta) => {
+        time += delta
+        ref.rotation.set(
+          x + pitch.intensity * pitch.noise.noise(pitch.frequency * time, 1),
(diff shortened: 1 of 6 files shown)


DefinitelyMaybe:
+1 for it going into extras

apps/docs/src/examples/shake/shake.ts
```diff
@@ -0,0 +1,52 @@
...
+  const shake = (ref: Object3D, options: ThrelteUseTaskOptions = {}) => {
+    const { x, y, z } = ref.rotation
+    return useTask((delta) => {
+      time += delta
+      ref.rotation.set(
+        x + pitch.intensity * pitch.noise.noise(pitch.frequency * time, 1),
+        y + yaw.intensity * yaw.noise.noise(yaw.frequency * time, 1),
+        z + roll.intensity * roll.noise.noise(roll.frequency * time, 1)
+      )
+    }, options)
```
grischaerbe, line 44: This needs to be hoisted into top level `useShake` since it accesses context. If `ref` of a component that has `oncreate={shake}` declared, the context would be accessed again outside of component initialization. The pattern to follow is this:

```ts
export const useShake = (params: any) => {
  let ref: Object3D | undefined = undefined

  const { start, stop } = useTask((delta) => {
    // shake
  }, {
	 autoStart: false
	})

  const shake = (newRef: Object3D) => {
    ref = newRef
    start()
    return () => {
      stop()
      ref = undefined
    }
  }
  
  return {
    shake,
    // …
  }
}
```

grischaerbe (changes requested):
Nice! This should 100% be in extras!

joshwashywash:
@grischaerbe do you mean something like this for a changing ref?:

```svelte
<script lang="ts">
  import { interactivity } from '@threlte/extras'
  import { useShake } from '../camera/shake'
  import { BoxGeometry, Mesh, MeshStandardMaterial } from 'three'

  const { shake } = useShake()

  const createMesh = () => {
    return new Mesh(new BoxGeometry(), new MeshStandardMaterial({ color: 'orangered' }))
  }

  let ref = $state(createMesh())
  
  interactivity()
</script>

<T.AmbientLight />

<T
  is={ref}
  onclick={() => {
    ref = createMesh()
  }}
  oncreate={(ref) => {
    console.log('starting')
    const { stop } = shake(ref)
    return () => {
      console.log('stopping')
      stop()
    }
  }}
/>
```

PR #1202, opened 2024-11-12
[next] Tsl slice shader example

joshwashywash:
This example showcases 2 important things

1. how to extend three's materials
2. how to use [TSL](https://github.com/mrdoob/three.js/wiki/Three.js-Shading-Language) with threlte

There's two things I'm not sure about

1. Getting the correct types for `<T>` inside `SliceMaterial.svelte`. I can't figure out why it's erroring in `SliceMaterial` but doesn't error in something like `Detailed.svelte` from `extras`.
2. TSL has a lot of types that encapsulate different kinds of data. I think that's by design because various nodes could be a lot of things. I opted for being very generic with the TSL function types inside `tsl.ts`. The only bit I'm not satisfied with is this piece of code

```typescript
$effect(() => {
  uColor.value.set(color);
}
```

I believe `value` is an instance of `Three.Color`. It at least has a `set` method but typescript thinks that `value` is a `ShaderNodeObject<Node>` because that's the return type of the `color` function from `three/tsl` . To get rid of the error you'd have to make a new Color whenever the user updates the `color` prop from the pane. 

```typescript
$effect(() => {
  uColor.value = color(sliceColor)
})
```

That's not necessary since we can just modify what's already in the uniform.

---

In the second mini-example, I used a texture we already had. Feel free to use a different one. I was thinking we could replace it with an earth texture but what's there now is probably good enough.


joshwashywash:
Not sure why prettier check failed. i ran `pnpm format` and it said everything was okay.

DefinitelyMaybe:
> Not sure why prettier check failed. i ran `pnpm format` and it said everything was okay.

possibly needs to merge in next branch changes

joshwashywash:
so running `pnpm format` after pulling next and merging it did a lot to some files in next. just formatting changest though.

apps/docs/src/content/examples/shaders/TSL Slice Material.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+  extend(THREE)
+</script>
+```
+
+The extension makes it possible to use class definitions such as `MeshPhysicalNodeMaterial` on `<T>` components.
+
+```svelte
+<T.MeshPhysicalNodeMaterial />
+```
```
michealparks, line 44: This section is very interesting and likely worthwhile when generally using `three/webgpu`. Do you think it makes sense to perhaps move it to the webgpu renderer example and link a "tips for getting started with webgpu"? Or alternatively copy this section for that example as well?

michealparks: Another random thought that I had when reading this over that is definitely outside the scope of this PR but is a bit of a looming problem: 

Three.js's recent change to bundle `three/webgpu` separately will probably pose big tree shaking issues for us as its adoption increases since the T component does a wildcard import of the `three` package. So here we're actually wildcard importing both `three` and `three/webgpu`, which I bet will result in large bundle sizes.

joshwashywash: > This section is very interesting and likely worthwhile when generally using `three/webgpu`. Do you think it makes sense to perhaps move it to the webgpu renderer example and link a "tips for getting started with webgpu"? Or alternatively copy this section for that example as well?

i agree. i wonder if we should have a webgpu section that has a /examples/webgpu/getting-started index page or something along those lines.

joshwashywash: > Another random thought that I had when reading this over that is definitely outside the scope of this PR but is a bit of a looming problem:
> 
> Three.js's recent change to bundle `three/webgpu` separately will probably pose big tree shaking issues for us as its adoption increases since the T (`T.ts` in core) component does a wildcard import of the `three` package. So here we're actually wildcard importing both `three` and `three/webgpu`, which I bet will result in large bundle sizes.

yeah you can see the console logs out that two instances of three are being used. same with the other webgpu example. i'm not sure how to consolidate that.

michealparks: I'm definitely unsure as well. First thought is that we may have to introduce a `@threlte/core/webgpu` export with a different internal three import. Not sure yet, but we've got time to think!

DefinitelyMaybe: Pretty sure three.js team is still trying to figure out their best route too: https://github.com/mrdoob/three.js/pull/29404

DefinitelyMaybe:
Coming back to this, I think there's a case for going one step further with what @michealparks was saying

> Do you think it makes sense to perhaps move it to the webgpu renderer example and link a "tips for getting started with webgpu"?

I purpose we create two new pages under `learn/advanced`. One about setting-up/using webgpu (and it's experimental status) and the other about getting started with TSL.

joshwashywash:
> I purpose we create two new pages under learn/advanced. One about setting-up/using webgpu (and it's experimental status) and the other about getting started with TSL.

Oh yeah we could do that. I'll open an issue for that and link it here.

Not sure if we should house both webgpu examples under `learn/advanced`. I'll make note of that in the issue.

DefinitelyMaybe:
Sweet! LGTM 👍

DefinitelyMaybe (approved):

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+## WebGPU
+
+To use Three.js's WebGPU renderer, import it and then initialize it within your
+`<Canvas>`'s `createRenderer` prop.
+
+```svelte title="App.svelte" {4}+ {8-14}+
+<script>
+  import Scene from './Scene.svelte'
+  import { Canvas } from '@threlte/core'
+  import * as THREE from 'three/webgpu'
```
michealparks, line 25: Would recommend against wildcard imports in the docs to encourage best practices for tree shaking:
```suggestion
  import { WebGPURenderer } from 'three/webgpu'
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+</script>
+
+<Canvas
+  createRenderer={(canvas) => {
+    return new THREE.WebGPURenderer({
```
michealparks, line 30:
```suggestion
    return new WebGPURenderer({
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+    })
+  }}
+>
+  <Scene />
+</Canvas>
+```
+
+<Tip type="note">
+  WebGPU is still an experimental browser api and at the time of writing has 73% global
+  [support](https://caniuse.com/?search=webgpu). For this reason, Three.js's webgpu renderer
```
michealparks, line 43:
```suggestion
  WebGPU is still an experimental browser api and at the time of writing has [limited availability across major browsers](https://caniuse.com/?search=webgpu). For this reason, Three.js's webgpu renderer
```

Easier to keep updated 😂

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+<script>
+  import { Canvas, T } from '@threlte/core@next'
+  import * as THREE from 'three/webgpu'
+  let renderMode = $state('manual')
+</script>
+
+<Canvas
+  {renderMode}
+  createRenderer={(canvas) => {
+    const renderer = new THREE.WebGPURenderer({
```
michealparks, line 70:
```suggestion
    const renderer = new WebGPURenderer({
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+use a find and replace solution to get this job done. Three.js has identified
+that it's not an ideal solution and recommends using the [Three.js Shading Language](https://github.com/mrdoob/three.js/wiki/Three.js-Shading-Language) or TSL for short.
+
+The example below is an adaptation of [this](https://threejs.org/examples/?q=tsl#webgpu_tsl_angular_slicing)
+Three.js example. There are many more [TSL examples](https://threejs.org/examples/?q=tsl)
+within Three.js that you can use or adapt for your project.
+
+<Example path="shaders/slice/main" />
+
+### Extend the catalogue
```
michealparks, line 121: Even though this is convenient it still has big implications for bundle size. I'd recommend warning about this caveat and [also providing the alternative method that verekia came up with](https://github.com/verekia/three-gpu-ecosystem-tests/blob/main/vite-ts-threlte8/src/App.svelte) (which I'd use - and probably always recommend for stuff that's going into production considering how much unused code is imported with this method)

apps/docs/src/content/learn/more/resources.mdx
```diff
@@ -3,11 +3,14 @@ category: More
 title: Resources
 ---
 
-This is a curated list of invaluable resources that will assist you in mastering Threlte and its underlying technologies. Whether you are a beginner looking to get started or an experienced developer seeking advanced knowledge, these resources offer a wealth of information to level up your skills.
+The three.js space is constantly evolving and expanding. Whether you are a beginner
+looking to get started or an experienced developer seeking advanced knowledge,
+there are resources in the wider communitty which offer a wealth of information
```
michealparks, line 8:
```suggestion
there are resources in the wider community which offer a wealth of information
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,192 @@
...
+  })(arg2, arg2)}
+/>
+```
+
+Node materials give you the ability to modify three's builtin materials.
+In the sliced gear example, two nodes are modified; the `outputNode` and the `shadowNode`.
+The `outputNode` is set up in such a way that it discards any fragments that are
+outside the permitted `startAngle` and `arcAngle`. If a fragment is not discarded and
+it is not front-facing, it is assigned the color in the `color` uniform. The material
+needs its `side` set to `THREE.DoubleSide` otherwise threejs will cull them out if they are facing away
```
michealparks, line 170:
```suggestion
needs its `side` set to `THREE.DoubleSide` otherwise three.js will cull them out if they are facing away
```

apps/docs/src/examples/shaders/slice/main/App.svelte
```diff
@@ -0,0 +1,70 @@
+<script lang="ts">
+  import * as THREE from 'three/webgpu'
```
michealparks, line 2:
```suggestion
  import { MathUtils } from 'three/webgpu'
```

apps/docs/src/examples/shaders/slice/main/App.svelte
```diff
@@ -0,0 +1,70 @@
...
+
+  extend(THREE)
+
+  let arcAngleDegrees = $state(90)
+  let startAngleDegrees = $state(60)
+  let sliceColor = $state('#ff4500')
+  let rotate = $state(true)
+
+  const arcAngle = $derived(THREE.MathUtils.DEG2RAD * arcAngleDegrees)
+  const startAngle = $derived(THREE.MathUtils.DEG2RAD * startAngleDegrees)
```
michealparks, line 15:
```suggestion
  const arcAngle = $derived(MathUtils.DEG2RAD * arcAngleDegrees)
  const startAngle = $derived(MathUtils.DEG2RAD * startAngleDegrees)
```

michealparks (approved):
🚀

apps/docs/src/content/learn/advanced/webgpu.mdx
DefinitelyMaybe: Hmm I think I'd like to talk about both. Maybe start with the easy Laissez-faire option of going wholesale and then talking about caveats in a sentence or two afterwards. It's a good point and grisch said verekia had tested it out too.

michealparks: Yep! Completely agreed. I'd recommend suggesting both approaches and making explicit the consequences of each.

DefinitelyMaybe: I put a lot of thought into structuring and ordering. I hope ya'll like it 😄

DefinitelyMaybe:
@joshwashywash Unfortunately something I've done to the examples has meant that lighting no longer works in our TSL examples. Could I please have your help with it?

joshwashywash:
strangely shadows work locally but not in the deployment? is that happening for you too @DefinitelyMaybe ?

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,245 @@
...
+The example below is an adaptation of [this](https://threejs.org/examples/?q=tsl#webgpu_tsl_angular_slicing)
+Three.js example. There are many more [TSL examples](https://threejs.org/examples/?q=tsl)
+within Three.js that you can use or adapt for your project.
+
+<Example path="shaders/slice/main" />
+
+### Using the new classes
+
+By default `<T>` doesn't know about the new classes and will
+error on things like `<T.MeshPhysicalNodeMaterial />`. You have two options.
```
michealparks, line 124:
```suggestion
`<T>` doesn't know about the new `WebGPURenderer` classes because they are part of a separate and new `three/webgpu` export, and creating components like `<T.MeshPhysicalNodeMaterial />` will throw errors. There are two options to fix this.
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+1. Give `<T>` all the new classes from `three/webgpu` with the [`extend`](/docs/reference/core/t#extending-the-default-component-catalogue)
+   function. Adding all the classes will increase the bundle size of your application
+   but is easy and works.
```
michealparks, line 128:
```suggestion
1. Provide `<T>` with all the new classes from `three/webgpu` with the [`extend`](/docs/reference/core/t#extending-the-default-component-catalogue)
   function. It's important to note that using a wildcard import for adding all the classes will increase the bundle size of your application because both `three` and `three/webgpu` will be imported in a non-tree-shakable way. However, this method is easy and works.
```

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -0,0 +1,245 @@
...
+>
+  <Scene />
+</Canvas>
+```
+
+Bundle size of your application will remain minimal but you'll have to keep things updated as you go.
+
+<Tip type="tip">
+  An easy option for projects is to start with option #1 and then transition to option #2 when
+  bundle size becomes an issue.
```
michealparks, line 200:
```suggestion
  bundle size becomes an issue or when you need to ship to production.
```

michealparks:
Love the improvements! Just added a few last suggestions.

DefinitelyMaybe:
TIL! well done @joshwashywash 

Having a look over things now

DefinitelyMaybe:
I'm good. that's me. I've edited and re-edited this page enough.

@joshwashywash merge when your ready 👍 this look pretty darn good to me 💯

joshwashywash:
i can't figure out why it's hanging on a linting check. i ran the command locally and there's no file changes.


https://github.com/user-attachments/assets/fa030a78-92fa-4d27-aabc-e8ca8e6fa58c

PR #1204, opened 2024-11-12
[next] format

joshwashywash:
just formatting the Hero data.json file so that it's not holding up other pull requests.


grischaerbe (approved):
LGTP 💅

PR #1206, opened 2024-11-15
[next] Use fbo simplifications

joshwashywash:
i was working with trying to make a `RenderTexture` component or an example of an example of a screen quad / mini postprocessing example. Similar to what's described in this [article](https://luruke.medium.com/simple-postprocessing-in-three-js-91936ecadfb7).

i noticed that fbo would be really nice to use here but while reading the docs, i decided that it could use a little spruce up.

The original port was almost a 1:1 with drei's. Most of the changes are to simplify the example and to get rid of things like `as Type` casts and to let the threejs defaults pass through.

More docs about the different ways to call `useFBO` and a small code snippet about how to use the optional depth texture have been added.

because the WebGLRenderTarget defaults are now used and we were using nondefaults in `useFBO`, this is a breaking change. i'm not sure if there should be a minor or major changeset for this. probably minor right because at the time of writing this, the next branch has not been officially released.

### DIFF
--- a/.changeset/wise-dryers-shop.md
+++ b/.changeset/wise-dryers-shop.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": minor
+"@threlte/docs": patch
+---
+
+updates useFBO to use default options of `WebGLRenderTarget`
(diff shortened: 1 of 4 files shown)


joshwashywash:
this pr will be too difficult to test

PR #1207, opened 2024-11-15
[next] useFBO rewrite

joshwashywash:
This is a rewrite of `useFBO` that adds functionality, improves readability of the implementation, and adds to the docs to be more comprehensive.

`useFBO`'s signature is now

```typescript
export type UseFBOOptions = RenderTargetOptions & {
  /**
   * if set, the scene depth will be rendered into buffer.depthTexture.
   */
  depth?: { width?: number; height?: number } | DepthTexture | boolean
  /**
   * if set, the render target size will be set to the corresponding width and height and not use or follow the size of the canvas
   */
  dimensions?: { width?: number; height?: number }
}

export function useFBO({
  depth = false,
  dimensions,
  ...targetOptions
}: UseFBOOptions = {}): WebGLRenderTarget {
```

depth is now allowed to be either a boolean, depth texture, or dimensions object and `useFBO` handles each case according to what's described in the .mdx file.

the implementation of useFBO has been modified to be more readable and traceable. The old implementation was a very close 1:1 with drei's useFBO and i think deviating from there yields a more interesting and useful hook.

The corresponding mdx file has been modified to account for all the various uses of `useFBO`. It now contains many code snippets as examples are usually more explicit than docs.

The example has been rewritten to be more simple while retaining its original intent and goal.

### DIFF
--- a/apps/docs/src/content/reference/extras/use-fbo.mdx
+++ b/apps/docs/src/content/reference/extras/use-fbo.mdx
@@ -14,7 +14,7 @@ Framebuffer objects are useful when you want to render a scene but not display i
 
 ## Options
 
-```ts
+```typescript
 type UseFBOOptions = THREE.WebGLRenderTargetOptions & {
   /**
    * If set, the scene depth will be rendered into buffer.depthTexture.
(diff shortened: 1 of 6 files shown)


packages/extras/src/lib/hooks/useFBO.ts
```diff
@@ -1,67 +1,54 @@
...
-    magFilter: LinearFilter,
-    type: HalfFloatType,
-    ...targetOptions
-  })
-
-  if (depth) {
-    target.depthTexture = new DepthTexture(_width, _height, FloatType)
+export const useFBO = ({
+  depth = false,
+  dimensions,
```
michealparks, line 21: Think it makes sense to call this `size` to match the naming convention of the `useThrelte` hook?

apps/docs/src/content/reference/extras/use-fbo.mdx
```diff
@@ -6,28 +6,129 @@ name: useFBO
 type: 'hook'
 ---
 
-useFBO (Framebuffer Object) is a hook that creates a `THREE.WebGLRenderTarget`. This is a port of [drei's useFBO hook](https://github.com/pmndrs/drei#usefbo).
+`useFBO` (Framebuffer Object) is a port of [drei's useFBO](https://github.com/pmndrs/drei#usefbo) that creates a [THREE.WebGLRenderTarget](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderTarget).
 
-Framebuffer objects are useful for cases where you want to render a scene but not display it. For example, rendering a scene to a framebuffer then passing that to a fragment shader as a uniform. You can read more about `WebGLRenderTarget`'s [here](https://threejs.org/docs/index.html?q=render%20target#api/en/renderers/WebGLRenderTarget).
+Framebuffer objects are useful when you want to render a scene but not display it. For example, rendering a scene to a framebuffer then passing that to a fragment shader as a texture.
```
michealparks, line 11: The wording here is pre-existing but upon re-reading maybe can be improved / simplified? 
```suggestion
Framebuffer objects are useful when you want to render a scene to something other than the canvas, such as a texture.
```

joshwashywash: what do you think about this

> Framebuffer objects are useful when you want to render a scene to something other than the canvas. In ThreeJS, you do this by rendering the scene to a `RenderTarget`. The render target has a `texture` property that can be used for various things such as postprocessing effects or applying the texture to something in the scene.

joshwashywash:
i modified the example to better show off how you might use a render target. it's not much more complex than what was there already but I think it ties in a little better with the updated introduction paragraph.

grischaerbe (approved):
Awesome! I slightly simplified the example in the docs, I hope that works for you. We're good to go 🚀

PR #1208, opened 2024-11-16
[next] View Component w/ some core changes

DefinitelyMaybe:
Changes:

- canvas context became dom context. It holds the actual wrapper div and canvas dom, along with the size. Size type is now a `DOMRect` instead of the custom `{width:number, height:number}`
- `useThrelte` hook now has the `canvas` and `dom` objects
  - The `useThrelte` hook page has been updated accordingly
- updated extras and examples to use the new dom context
- Updates the docs events pages
  - The `learn/basics/handling-events` page has been removed. 
    - It's example moved to the `extras/interactivity` page.
  - The `reference/core/t` page references the `extras/interactivity` plugin.
  - The `reference/core/canvas` page adds a note to attach events to the `useThrelte` `dom` object
- renames the `test` example to `extras/interactivity`

The new Component:

- outer View initializes contexts and creates an `intersectionObserver` for the new dom element
- inner View takes care of renderering 🎉 

huge thanks to @grischaerbe for the initial port of this component as well as a lot of improvements while its been cooking 👨‍🍳


packages/extras/src/lib/components/View/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { Props } from '@threlte/core'
+import type { Scene } from 'three'
+
+/**
+ * [Scissor cut renders](https://threejs.org/examples/?q=mul#webgl_multiple_elements) its contents to the given dom area. The default threlte scene is used if none provided.
+ */
+export type ViewProps = Props<Scene> & {
```
DefinitelyMaybe, line 7: How to type nothing ref here?

apps/docs/src/examples/extras/view/App.svelte
```diff
@@ -0,0 +1,97 @@
...
+    // add one random mesh to each scene
+    const geometry = geometries[(geometries.length * Math.random()) | 0]!
+
+    const material = new THREE.MeshStandardMaterial({
+      color: new THREE.Color().setHSL(Math.random(), 1, 0.75, THREE.SRGBColorSpace),
+      roughness: 0.5,
+      metalness: 0,
+      flatShading: true
+    })
+    // @ts-ignore
```
DefinitelyMaybe, line 25: hopefully sorting the type means I can remove this comment.

packages/core/src/lib/context/fragments/dom.ts
```diff
@@ -0,0 +1,63 @@
...
+  dom: HTMLElement
+}
+
+export const createDOMContext = (options: CreateDOMContextOptions) => {
+  const domRect = options.dom.getBoundingClientRect()
+
+  let lastWidth = domRect.width
+  let lastHeight = domRect.height
+
+  const size = currentWritable({ width: domRect.width, height: domRect.height })
```
phoenisx, line 25: I think it would be nice if we could forward other `Rect` properties to size as well. Cause at points getting more than width/height is helpful.

packages/core/src/lib/context/fragments/dom.ts
phoenisx, line 1: Is there a reason for not using Svelte signals, instead of Readables/Writables?
Are object `Proxy` slow?

phoenisx (commented):
Thanks for the PR.
LGTM! 🎉

packages/core/src/lib/context/fragments/dom.ts
DefinitelyMaybe: 🤷 Maybe just stores for now and we update later if it does become a bottleneck

phoenisx:
@DefinitelyMaybe Are events working for you?

When I try to create a Mesh or a Group, and add events to it like `onpointerdown`, they are not getting triggered for me at all.
I debugged the code, and I see they are getting registered, but they are not getting triggered for sure. (I might be doing something wrong, but I'm not sure)

I tried updating ur example as well, with the following change:

```svelte
// apps/docs/src/examples/extras/view/Scene.svelte
<T.Mesh rotation.y={rotation} onpointerdown={() => {console.log('Foo')} }>
  ...
</T.Mesh>
```

```svelte
// apps/docs/src/examples/extras/view/App.svelte
// Tested both with and without View, both didn't work
<Canvas>
  <Scene {...items[0]} />
  <!-- {#each items as item} -->
    <!-- <View dom={item.dom}> -->
      <!-- <Scene {...item} /> -->
    <!-- </View> -->
  <!-- {/each} -->
</Canvas>
```

DefinitelyMaybe:
I haven't added `interactivity()` into the doc example but it should be working. I did update the [interactivity context](https://github.com/threlte/threlte/blob/08fab4decf8ba850802cbb34f833fadbcd54849b/packages/extras/src/lib/interactivity/context.ts#L58) too so they would be attaching to the new DOM like you say.

maybe there's been recent changes in next I should pull in just to check

phoenisx:
Apologies. I copy-pasted your code, removed mine, and removed `interactivity()` setup code, which broke that functionality in my repo. I completely forgot about setting `interactivity` again.

Enabling `interactivity()` fixed the issue.

grischaerbe (approved):

phoenisx:
@DefinitelyMaybe I was facing issues with interactivity, thus created a repro here: https://github.com/phoenisx/repros/tree/main/threlte-view-repro

### Issue:
Since Views can have different viewport size compared to the parent canvas, we need to modify the `compute()` behaviour of the `interactivity` API.
Interactivity API works if we provide the proper dimensions of the viewport, but `dom` -> `size.current` provided by `useThrelte()` context returns an incorrect current `dom` rect dimension, for the active scene the interactivity is set for.


Also attaching the recording to demo the issue that I am facing:

<video type="video/webm" src="https://github.com/user-attachments/assets/085d4c20-32b2-4d42-b7e6-13be03bd6325"></video>

DefinitelyMaybe:
excellent! good catch 👍 lets jump into the source and try figure it out

DefinitelyMaybe:
Think I figured it out but could I get a double check from you @phoenisx 

Having any of the sizes off whether that be in `<View>` or otherwise is weird because it all comes back to the same `ResizeObserver` in `lib/context/fragments/dom.ts`. Size is usually only ever updated once on mounting so I `console.table()`'d the values.

Capture

It appears that even though we initially set things to the correct value with `dom.getBoundingClientRect()`, when the `ResizeObserver` fired an initial event the `entry.contentRect` DOMRect is not the same but updates anyway to the weird 0 values.

DefinitelyMaybe:
The simplest fix I can think of is to reduce down the function to a minimum i.e.

```typescript
const resizeObserver = new ResizeObserver(() => {
  const newDomRect = dom.getBoundingClientRect()
  size.set(newDomRect)
})
```

phoenisx:
Yeah!
For me calling `getBoundingClientRect` on resize worked.
Though I'm not sure why `entry.contentRect` doesn't work. It should do the same right, not sure. I couldn't find any details on this online.
Maybe I'm missing something here 🤔

PR #1210, opened 2024-11-21
[next] export `Size` type from core

joshwashywash:
```typescript
import type { Size } from '@threlte/core';
```

is used in a few places such as `extras/.../useViewport.ts`.

this pr exports the `Size` type from core to fix that and also so that people can type functions that acceptale a size

```typescript
// doSomethingWithASize.ts
import type { Size } from '@threlte/core';

export const doSomethingWithASize = (size: Size) => {
  // ...
};
```

```svelte
<script lang="ts">
import { doSomethingWithASize } from './doSomethingWithSize';
import { useThrelte, watch } from '@threlte/core';

const { size } = useThrelte();

watch(size, doSomethingWithSize);
</script>
```


grischaerbe (approved):
Thanks, LGTM 🚀

PR #1216, opened 2024-11-28
[next] plugins pages adjustments

DefinitelyMaybe:


grischaerbe (approved):
I slightly reworded it and made "When to use a plugin" a separate section, I hope that works for you!

PR #1218, opened 2024-11-28
[next] InstancedSprite fix for three 170+

jerzakm:
Updated the instanced sprite lib for three v170 and removed PerfMonitor component from examples until it's fixed.


DefinitelyMaybe (approved):
LGTM 👍

PR #1229, opened 2024-12-02
[next] Environment Updates

### LINKED ISSUE #1217: [next] refactor <Environment> and split it into two components - a multifile version and a single file version
## background

so i'm trying to refactor `<Environment>` cuz it's still in svelte 4 but it's super overloaded to handle different image formats. some formats take one file, for example hdr and exr and others are cube maps where you pass in 6 images into the `files` prop. There is slight overlap with the loaders for these formats but not enough because if you try to derive the loader and call `.load` the type of the first argument is `string & string[]`.

Screenshot 2024-11-27 at 7 45 28 PM

## proposal

what do you all think about splitting `<Environment>` into `<CubeEnvironment>` and `<Environment>`. (not sure if i like these names so its open for suggestion) a cube map version and a single file version. someone with more knowledge can tell me if it's at all common to switch between environments using a cube map with 6 separate images and one that uses a single image.

this change would make the component much easier to type and guarantee correctness. right now there's a lot of `as whatever` casting going on and a lot of different things need to be checked to narrow down the cases. part of designing a good api is having systems that handle whatever you throw at them but i think there's a tradeoff between a very overloaded component and type correctness, dev ex, readability, and keeping things simple and tractable.

the component props of both `<Environment>` and `<CubeEnvironment` would essentially be the same except `<CubeEnvironment>` 's `file` prop would be `[string, string, string, string, string, string]` instead of `string`. actually `files` on `<Environment>` would be named `file` emphasizing that it only operates with one file. this change would mean that the `load` function for `loaders` in each respective component is typed in such a way that you can guarantee that the type of `files` is what `loader.load` expects - `string` for `<Environment>` or `string[]` for `<CubeEnvironment>`. you can also guarantee that `files` has 6 entries in it in the case of `<CubeEnvironment>`, in other words that `files` is a six-tuple.

the way `<Environment>` is implemented now is similar to if `PerspectiveCamera` and `OrthographicCamera` were squished all into one `Camera` class. maybe not to that degree but it's a good analogy 😏 .

## drei's \<Environment>

regarding drei, they stick everything into `<Environment>` but this is just like the case i made for splitting up `GradientTexture`  into `<RadialGradientTexture>` and `<LinearGradientTexture>`. i think this is another place where we should deviate from drei.

## additional benefits

this alteration should make it easier to extend the environment functionality. one thing i think would be cool is to dynamically apply the Environment to different scenes.

`<Environment scene={sceneToApplyEnvTo} />`

## downsides of such a change

the functionality we lose with splitting them up in such a way is that a user can't as easily swap a "cube environment" for a single texture environment. right now this is possible

```svelte
<script lang="ts">
  type CubeMapTuple = [string, string, string, string, string, string];
  const multiple: CubeMapTuple = [
    'px.png',
    'nx.png',
    'py.png',
    'ny.png',
    'pz.png',
    'nz.png'
  ];
  const single = 'singleTexture.png';

// imagine useCubeFiles can be toggled on and off by some event
const files = $derive(useCubeFiles ? cubeMap : single);
</script>

<Environment {files} />
```

in the proposed change above this would become

```svelte
<script lang="ts">
  type CubeMapTuple = [string, string, string, string, string, string];
  const files: CubeMapTuple = [
    'px.png',
    'nx.png',
    'py.png',
    'ny.png',
    'pz.png',
    'nz.png'
  ];
  const file = 'singleTexture.png';
</script>

{#if useCubeMap}
  <CubeEnvironment {files} />
{:else}
  <Environment {file} />
{/if}
```

## other goals of the refactor

> [!NOTE]
> this section may be updated from time to time as a pr for this change is in the works

- [ ] make something like [this example](https://codesandbox.io/p/sandbox/aquarium-forked-jw26ln?file=/src/App.js:18,39) possible. drei's \<Environment> has a special case when children are used. if children are used it will portal the children into a different scene and applies a cube camera texture to the "portal" scene.
- [x] load textures asynchronously using three's `loader.loadAsync` and threlte's `suspense`.
  - [x] `loadAsync` allows you to pass in an `onProgress` callback. should this be a prop on the component?
- [x] `onloadercreated` function prop to be able to further customize the loader after it is created such as setting its credentials or its crossorigin property. Essentially all the set methods of the [loader](https://threejs.org/docs/#api/en/loaders/Loader). either this or a `loader` prop to be able to use your own loader.
- [x] provide a way for the loaded texture to be modified. either a `ontextureloaded` or a getter for the loaded texture.
- [x] allow user to pass in a texture if they already have the cube texture loaded.

joshwashywash:
closes #1217

Separates `<Environment>` into two components -> `<Environment>` which is used for equirectangular-mapped textures and `<CubeEnvironment>` which is used for cube maps. The reason for this is to delineate which loaders to use for each case -> `<Environment>` uses loaders that accept a single fil in their `load` function and `<CubeEnvironment>` uses loaders that accept 6 files - one for each face of the cube.

### Changes that apply to both components

- Can now accept a texture instance that will skip loading a texture but still set `scene.environment`, handle disposal, etc.
- rename `file` prop to `resource` for `<Environment>` and `resources` for `<CubeEnvironment`> to clarify that it can either be a file path string or a texture instance
- Allow for modification of texture if it has to be loaded through the `loadOptions` prop
- Allow for more customization of the loader through the `loaderOptions` prop. Before you could only set its `path` but now you can do things like set its fetch credentials, cross origin policy, etc. Everything that you can do if you instantiated the loader yourself. [Loader docs](https://threejs.org/docs/index.html?q=loader#api/en/loaders/Loader). This makes it much easier to load files over the network
- `scene` prop that allows you to choose which scene will have its environment/background affected
- bindable skybox prop which is a reference to a `Three.GroundedSkybox` instance that you can treat as any other mesh

### `<Environment>` specific
- Support for `exr` files

### non-component changes

- updated docs and examples around `Environment` and `CubeEnvironment`
- reorganized texture file locations
  - equirectangular textures now live at `/apps/docs/public/textures/equirectangular`
  - cube textures now live at `/apps/docs/public/textures/cube/
- further separation of equirectangular files into directories based on their type
  - exr files live at `/apps/docs/public/textures/equirectangular/exr
  - hdr files live at `/apps/docs/public/textures/equirectangular/hdr
  - jpg files live at `/apps/docs/public/textures/equirectangular/jpg
- updated all uses of `Environment` in all other examples

---

I will add more changes if i think of any that I haven't documented. The updated .mdx files should be comprehensive


joshwashywash:
the swap scene example is still broken. to reproduce the bug, refresh the page, set the `isBackground` checkbox to `on`, swap to right scene. then check off "use Environment". the right side should be pink, but it is red. i'll work on fixing it tomorrow.

joshwashywash:
all the examples on the environment.mdx page are working. i need to go in and update the props switching `resource` to `url` and its type to `url?: string`. then i will update the docs to match the updated api.

grischaerbe (approved):
Holy cow, that's nice!! 🚀

Issue #1242, opened 2024-12-11
[next] import of `./types.d.ts` from `./dist/index.js` makes Threlte incompatible with svelte.dev/playground


Rich-Harris:
Because of [this line](https://github.com/threlte/threlte/blob/75dc30c43cd76210e87a189956efd5682c5117f5/packages/core/src/lib/index.ts#L1)...

```js
import './types.d.ts'
```

...the packaged `dist/index.js` file _also_ imports the `.d.ts` file. In the context of a typical Svelte app this isn't a problem, because there's a bundler involved. In the context of https://svelte.dev/playground, it _is_ a problem, because the playground doesn't (yet, at least) speak TypeScript.

Accordingly, [this happens](https://svelte.dev/playground/75e834f52c584112b49c9359ef38b7da?version=5.11.0):

<img width="1341" alt="image" src="https://github.com/user-attachments/assets/4ca60e7f-99b8-4281-8f7e-f5f1aeffd73d" />

There's likely a few ways we could tackle this. One would be for the playground to learn to speak TypeScript, though that's likely bit of a longer term thing.

Alternatively, what if the `"types"` export in `package.json` pointed to a file that re-exported the contents of `index` and _also_ imported the ambient module that declares the namespace?

```ts
// src/types.d.ts
import './namespace.d.ts';
export * from './index';
```

```ts
// src/namespace.d.ts
declare namespace Threlte {...}
```

Or even this?

```ts
// src/types.d.ts
declare global {
	namespace Threlte {...}
}

export * from './index';
```

That way, TypeScript would still discover the exports and the namespace, but environments that don't speak TypeScript wouldn't encounter it.

grischaerbe:
Fixed in #1243

PR #1243, opened 2024-12-11
fix: remove .d.ts import from dist/index.js

Rich-Harris:
Implements the suggestion in #1242 (albeit only for `@threlte/core` — not sure if it would also need to happen in other places?)


grischaerbe (approved):
Thanks! Clever solution 👏

grischaerbe:
Also as far as I know the only place we use this pattern, so should be good 👍

PR #1247, opened 2024-12-12
[next] replace useCursor with Hovering class

joshwashywash:
Replaces `useCursor` with a runes-based class, `Hovering`.

## Why the rename?

the functionality that `useCursor` provides is primarily about hovering over an object and then adjusting the cursor style. This is evident in its usage. `Hovering` has a `current` prop that when `true` adjusts the style of the cursor to the provided `hoverStyle` and sets it to `unhoverStyle` when `current` is false.

The logic is similar to the following

```typescript
if (current) {
  element.cursor.style = hoverStyle 
} else {
  element.cursor.style = unhoverStyle 
}
```

Actually it might make more sense to call this `Cursor` or `HoverStyle`. Either way we cant call this `UseCursor`.

## Why the change to a class?

Svelte seems to be favoring classes in their implementations. For example there's now class versions of [Spring](https://svelte.dev/docs/svelte/svelte-motion#Spring) and [Tweened](https://svelte.dev/docs/svelte/svelte-motion#Tween), there's [MediaQuery](https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery) which is a class and there's others. I think we should implement with classes where we can for hooks and utility things.

## Getter Parameters

both `hoverStyle` and `unhoverStyle` can now be getter functions which means we can get rid of passing in writables since now you can do something like this

```svelte
<script>
  const style = $state('move') // cursor style
  const hovering = new Hovering(() => style)
</script>
```

in the snippet above, when `hovering` is `true`, the style inside the state is accessed, so the cursor style is set to `move`. because its a getter, anytime the style updates, the style updates with it. There's an example of this in the new mdx page.

in the case where someone is using a store they can do this

```svelte
<script>
  import {} from 'svelte/store'
  
  const style = writable('move')
  const hovering = new Hovering(() => $style)
  
</script>
```

If they're using `Hovering` with stores outside of a svelte file, they can use the `get` function from `svelte/store'. 

I renamed the parameters to indicate better what the do. Before they were `onPointerOver` and `onPointerOut` which felt like they were event callbacks when in reality they are strings or, now, getters of strings.

## Removing `onPointerEnter` and `onPointerLeave`

One big difference is the removal of the `onPointerEnter` and `onPointerLeave` methods. both of these were just convenience methods that essentially did the following

```typescript
const onPointerEnter = () => {
  hovering = true
}

const onPointerLeave = () => {
  hovering = false
}
```

these are so simple that i don't think there's anything to be gained from putting them on the `Hovering` class. The user can create their own `onPointerEnter` and `onPointerLeave` callbacks if they need reusability. Something like the following:

```svelte
<script>
  import { Hovering } from '@threlte/extras'

  const hovering = new Hovering()
  const onPointerEnter = () => {
    hovering.current = true
  }

  const onPointerLeave = () => {
    hovering.current = false
  }
</script>

<T.Mesh onpointerenter={onPointerEnter} onpointerleave={onPointerLeave} />
```

This means that everywhere there was a destructure and usage of onPointerEnter/onPointerLeave is now replaced with inline functions that do the same thing.

All the examples have been updated.

## Other Changes

I made a couple of changes to some examples to modernize them. I tried not to do too much, just cleaned up some things that were easy to clean up.

The last example uses the <Text> component. for some reason it would always stay one style behind when updated so i moved the `text` prop out `...props` and passed it directly applying it like so seemed to fix things

```svelte
<Text
  {text}
  {...props}
/>
```


joshwashywash:
oops forgot to pull next before branching. will fix that.

joshwashywash:
gonna close this one and make a new one with better ideas

PR #1252, opened 2024-12-13
[next](docs) replace all mentions of `BoxBufferGeometry` with `BoxGeometry`

joshwashywash:


joshwashywash:
Okay, I guess it was late last night and I wasn’t paying attention but all those BufferGeometry should say BoxGeometry. I’ll fix that later today.

joshwashywash:
ready to go

apps/docs/src/content/learn/basics/disposing-objects.mdx
```diff
@@ -42,7 +42,7 @@ components. This disables disposal for the entire subtree.
 ```svelte
 <script>
   import { T, useTexture } from '@threlte/core'
-  import { BoxBufferGeometry, MeshStandardMaterial } from 'three'
+  import { BoxGeometry, MeshStandardMaterial } from 'three'
```
michealparks, line 45: Looks like we can remove this altogether? (it's just used in a T component)

apps/docs/src/content/reference/extras/use-cursor.mdx
```diff
@@ -32,7 +32,7 @@ Provide arguments to determine the cursor style. The defaults are`'pointer'` for
 
 <T.Mesh
   onpointerenter={onPointerEnter}
-  geometry={new BoxBufferGeometry(1, 1, 1)}
+  geometry={new BoxGeometry(1, 1, 1)}
```
michealparks, line 35:
```suggestion
  geometry={new BoxGeometry()}
```

michealparks (approved):

apps/docs/src/content/learn/basics/disposing-objects.mdx
joshwashywash: yeah we could replace with T.BoxGeometry in each case. same for the material too.

michealparks: I mean in this case this import is completely unused

joshwashywash: oh i see okay. i'll get that outta there.

Issue #1237, opened 2024-12-06
[next] allow OrbitControls to be standalone


joshwashywash:
I was thinking why isn't something like this possible?

```svelte
<script>
  import { OrbitControls } from '@threlte/extras'
  import { useThrelte } from '@threlte/core'
  
  const { camera } = useThrelte()
</script>

<OrbitControls camera={camera.current}>
```

or automatically grab the camera from `useThrelte()`. (See below)

Right now `OrbitControls` is coded such that the parent must be a Camera instance but it should not be too difficult to allow an optional `camera` prop. pseudocode for this feature might look like the following

```svelte
<!-- OrbitControls.svelte -->
<script lang="ts">
  import type { PerspectiveCamera, OrthographicCamera } from 'three'

  type Props = {
    camera: PerspectiveCamera | OrthographicCamera 
  }
  let { camera } = $props()

  const parent = useParent()

  // this won't work since parent is a store but you get the idea
  const _camera = $derived(camera ?? parent)

  // attach orbit controls to _camera

</script>
```

since `camera` would be reactive, this means that the controls would need to update to follow the updated camera. i think this should be possible with the following.

```svelte
<script>

  // same

  $effect(() => {
    controls.object = _camera
  })
</script>
```

i also wonder if we can default to the scene's default camera. More often than not i find myself creating a camera that is essentially the same as the default camera in order to use `<OrbitControls>`.

```svelte
<T.PerspectiveCamera makeDefault position.z={5}>
  <OrbitControls />
</T.PerspectiveCamera>
```

The camera above is the same as the default camera from `useCamera()`. Why not just default to the current camera?

```svelte
<script>
  // same

  import { useCamera } from '@threlte/core'
  import { fromStore } from 'svelte'

  const defaultCamera = fromStore(useCamera())

  const _camera = $derived(camera ?? defaultCamera.current)

  // attach controls to _camera
</script>
```

this way you can add `<OrbitControls>` to any scene without having to create a camera.

```svelte
<OrbitControls />
```

joshwashywash:
hmm so after trying it out we have to do something like this if we want to see changes that OrbitControls makes:

```svelte
<T.PerspectiveCamera makeDefault/>
<OrbitControls />
```

passing `camera` as a prop is still possible but you'd be updating a camera that the renderer is not using. so the above works or you can do the following

```svelte
<script>
  const { camera } = useCamera()
  camera.current.position.set(5, 5, 5)
</script>

<OrbitControls />
```

[...]

PR #1254, opened 2024-12-13
[next] Extras - new Gizmo using Three Viewport Gizmo library

michealparks:
### Overview

This PR overhauls and simplifies our Gizmo by letting `three-viewport-gizmo` do a lot of the heavy work. Replacing it solves a few issues and adds many new features:
- Adds `camera-controls` support
- Fixes a few few rendering bugs with the last implementation
- Adds alternate coordinate system support.
- Adds an alternative box gizmo.
- [A bunch of new options](https://fennec-hub.github.io/three-viewport-gizmo/api.html) from the library itself.

I've updated the docs accordingly.


packages/extras/src/lib/components/controls/OrbitControls/OrbitControls.svelte
```diff
@@ -31,17 +31,25 @@
...
+      props.onchange?.(event)
+    }
+
+    orbitControls.set(controls)
+    controls.addEventListener('change', handleChange)
+    return () => {
+      orbitControls.set(undefined)
+      controls.removeEventListener('change', handleChange)
+    }
+  })
```
michealparks, line 47: This fixes a bug in which OrbitControls was overwriting `onchange` listeners in some circumstances (like the gizmo example)

joshwashywash: i think i have the same sort of fix in my local experimenting branch for #1237

packages/extras/src/lib/components/Gizmo/Gizmo.svelte
```diff
@@ -1,467 +1,143 @@
...
-      clickTarget.style.bottom = ''
-      clickTarget.style.top = `${boundingRect.bottom - size - paddingY}px`
-    } else {
-      clickTarget.style.bottom = ''
-      clickTarget.style.top = `${paddingY + boundingRect.top}px`
-    }
-
-    clickTarget.style.height = `${size}px`
-    clickTarget.style.width = `${size}px`
+    ref.camera = $camera as PerspectiveCamera | OrthographicCamera
```
joshwashywash, line 63: so what i've been doing in some components is allow an optional camera prop that if not present will default to the camera in `useThrelte` something like this

```svelte
<script lang="ts">
  type Props = { camera: PerspectiveCamera | OrthographicCamera }
  let { camera, ref } = props()
  
  const { camera: defaultCamera } = useThrelte()
  
  // can't remember if defaultCamera is a store or has a `current` prop
  const _camera = $derived(camera ?? $defaultCamera)
  ref = new ViewportGizmo(camera.current)
  $effect(() => {
    if (_camera !== ref.camera) {
      ref.camera = _camera
    }
  })
</script>
```

should we do the same here? i don't think you'd ever want to control a camera that wasn't the one used for rendering but maybe???

packages/extras/src/lib/components/Gizmo/Gizmo.svelte
```diff
@@ -1,467 +1,143 @@
...
+    if (!_controls) return
 
-    if (intersects.length > 0) {
-      const alreadyReached = handleIntersection(intersects[0])
-      if (alreadyReached) {
-        // get the second closest intersection
-        if (intersects.length > 1) {
-          handleIntersection(intersects[1])
-        }
+    const handleStart = () => (_controls.enabled = false)
```
joshwashywash, line 69: it looks like attaching the gizmo will do this already i think https://github.com/Fennec-hub/three-viewport-gizmo/blob/291992c3a03648acf9ac7318a9c38ad7e4c797d6/lib/ViewportGizmo.ts#L377-L381

michealparks: good catch, fixed up!

packages/extras/src/lib/components/Gizmo/Gizmo.svelte
michealparks: camera arg added!

PR #1259, opened 2024-12-16
[next] update to three 171

### LINKED ISSUE #1257: [next] upgrade to three v171
all packages and the docs

joshwashywash:
gonna let it deploy and then check the examples. i think `gltf` may need special testing

closes #1257


grischaerbe:
Fixed in #1279

PR #1263, opened 2024-12-16
[next] provide way to start/stop html render task and a way to render on demand

### LINKED ISSUE #1262: [next] start and stop render task inside HTML
brought up at https://discord.com/channels/985983540804091964/1151118183785316352

i think we can do something like what `<AsciiRenderer>` does in that if you only need to render a frame, you can turn off the render task. here's the code for reference https://github.com/threlte/threlte/blob/425c70569be84a86082949d30c1bad61a010692e/packages/extras/src/lib/components/AsciiRenderer/AsciiRenderer.svelte#L70-L96

joshwashywash:
closes #1262 

~~i want to add an example for rendering on demand so i will undraft when i have that working.~~

i couldn't think of a good example

---

includes minor changes to `<AsciiRenderer>` - just using an $effect instead of an observe.

### DIFF
--- a/apps/docs/src/content/reference/extras/html.mdx
+++ b/apps/docs/src/content/reference/extras/html.mdx
@@ -51,7 +51,7 @@ componentSignature:
         { name: 'as', type: 'keyof HTMLElementTagNameMap', default: "'div'", required: false },
         { name: 'portal', type: 'HTMLElement', default: 'undefined', required: false },
         { name: 'castShadow', type: 'boolean', default: 'undefined', required: false },
-        { name: 'recieveShadow', type: 'boolean', default: 'undefined', required: false },
+        { name: 'receiveShadow', type: 'boolean', default: 'undefined', required: false },
         { name: 'material', type: 'THREE.Material', default: 'undefined', required: false },
         { name: 'geometry', type: 'THREE.BufferGeoemtry', default: 'undefined', required: false },
         {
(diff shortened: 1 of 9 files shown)


apps/docs/src/content/reference/extras/html.mdx
```diff
@@ -54,8 +71,75 @@ This component is a port of [drei's `<Html>` component](https://github.com/pmndr
 The container of your `<Canvas>` component needs to be set to `position: relative | absolute | sticky | fixed`. This is because the DOM element will be mounted as a sibling to the `<canvas>` element.
 </Tip>
 
+## Stopping and Starting the Task
+
+\<HTML> has an `autoRender` prop that you can use to turn off and on its render task. If at some point in your application, you no longer need to update the hmtl, you can set `autoRender` to `false`. If you need to resume the task, set `autoRender` back to `true`.
+
 <Example path="extras/html/basic" />
```
DefinitelyMaybe, line 78: Would be cool to see the example first imo

```suggestion
<Example path="extras/html/basic" />

## Stopping and Starting the Task

<HTML> has an `autoRender` prop that you can use to turn off and on its render task. If at some point in your application, you no longer need to update the hmtl, you can set `autoRender` to `false`. If you need to resume the task, set `autoRender` back to `true`.
```

apps/docs/src/content/reference/extras/html.mdx
```diff
@@ -54,8 +71,75 @@ This component is a port of [drei's `<Html>` component](https://github.com/pmndr
...
+</HTML>
+```
+
+In both cases you should set `autonRender` to false so that the render task doesn't automatically begin.
+
+Lastly, you can access these functions from the `<HTML>`'s children snippet.
+
+```svelte
+<HTML autoRender={false}>
+  {#snippet children({ render, startRendering, stopRendering })}
```
DefinitelyMaybe, line 135: I'm still learning svelte 5 snippets... in the Component Signature we've put `start`, `stop`, `render`, I take it order and spelling aren't an issue here?

apps/docs/src/examples/extras/html/phone/RoundedPlaneGeometry.ts
```diff
@@ -17,8 +17,6 @@ export class RoundedPlaneGeometry extends BufferGeometry {
       segments
     }
 
-    console.log(width, height)
```
DefinitelyMaybe, line 20: good catch!

packages/extras/src/lib/components/HTML/HTML.svelte
```diff
@@ -124,7 +125,7 @@
     }
   })
 
-  useTask(() => {
+  export const render = () => {
```
DefinitelyMaybe, line 128: hold up, why `export`? ... wait nvm, found the svelte docs for it: https://svelte.dev/docs/svelte/legacy-export-let#Component-exports. makes sense.

apps/docs/src/content/reference/extras/html.mdx
```diff
@@ -54,8 +71,75 @@ This component is a port of [drei's `<Html>` component](https://github.com/pmndr
...
 
+## Stopping and Starting the Task
+
+\<HTML> has an `autoRender` prop that you can use to turn off and on its render task. If at some point in your application, you no longer need to update the hmtl, you can set `autoRender` to `false`. If you need to resume the task, set `autoRender` back to `true`.
+
 <Example path="extras/html/basic" />
 
+\<HTML> also exports it's internal render task and the `startRendering`, `stopRendering`, and `render` functions so you can either render on demand or start and stop the internal task at your will.
+
+```svelte title="Render-On-Demand"
```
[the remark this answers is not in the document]
joshwashywash: maybe "manual rendering" would be a better term. essentially just not autostarting the task and calling `render` at keyframes or when necessary. I'll change the code snippet title and the phrasing on that.

apps/docs/src/content/reference/extras/html.mdx
joshwashywash: oh yeah gotta update that. ✅ i went with `startRendering` over just `start` that way you know what you're starting. it would be more evident if the component were named `HTMLRenderer` but i think this is okay.

packages/extras/src/lib/components/HTML/HTML.svelte
joshwashywash: oh yeah just a component export. it's interesting `export let` means two different things in svelte 4 and 5.

DefinitelyMaybe:
Great stuff with the headers 👍 I think thats a good change.

LGTM 👍

PR #1266, opened 2024-12-17
CI/CD update

DefinitelyMaybe:
PR actions are working for us now however Release PRs require the validations job. This PR simply renames that release job to validations.

@michealparks


grischaerbe:
I finally got the validations in version packages PR fixed. The problem is that by default the changeset PR is created by a github actor and PRs created by github do not trigger workflows to prevent infinite loops. I created a PAT and use that to create the changesets PR. This fixed the issue 👍

PR #1283, opened 2024-12-20
convert useCursor into a reactive class

let-josh:
introduces a `Cursor` class with reactive properties that aims to replace the store-based `useCursor`.

rather than update all the examples to use this class, this pr simply introduces the class. if this gets merged, then we can update each example and usage one by one to eventually deprecate `useCursor`.

usage of the class is shown below

```svelte
<script>
  import { Cursor } from '@threlte/extras'

  const cursor = new Cursor()
  const scale = $derived(cursor.hovering ? 2 : 1)
</script>


<!-- scale up the mesh when it is hovered -->
<T.Mesh
  {scale}
  onpointerenter={() => {
    cursor.hovering = true
  }}
  onpointerleave={() => {
    cursor.hovering = false
  }}
/>
```

you can see that there's no "hover" or "unhover" function. the `hovering` state can just be set from anywhere. if a user wants these functions they can define them outside the class like so:

```typescript
const cursor = new Cursor()
const hover = () => {
  cursor.hovering = true
}

const unhover = () => {
  cursor.hovering = false
}
```

## updating props at anytime

`Cursor` has a `.hoveringStyle` and `.unhoveringStyle` state that can be set at anytime to update the cursor's styles. These can even be set by an event or programmatically. the underlying effect will just get reran.

`Cursor` also has an `element` property which can be updated at anytime.

## considerations

its not clear that `.hoveringStyle`, `.unhoveringStyle`, `.hovering`, nor `.element` are all reactive. there doesn't seem to be a consensus in the svelte community on how to communicate that a variable is reactive. Svelte just says in the docs "hey, by the way, this is reactive" - for example in their `svelte/reactivity` and `svelte/motion` packages.

i'm against adding `.hover()` and `.unhover()` methods but they could easily be added. i don't want to confuse people with "should I be calling `.unhover()` or should i be setting `.hovering = false` myself?"

i think if we do add these methods than `.hovering` should be made a private variable and a getter for the value should be made. that way you can not set `.hovering` yourself. instead you must always use `.hover()` and `.unhover()`. implementation might look like the following:

```typescript
class Cursor {
  this.#hovering = $state(false)
  constructor() {
    // same but with `this.#hovering` instead of `this.hovering`
  }

  hover() {
    this.#hovering = true
  }
  unhover() {
    this.#hovering = false
  }
  get hovering() {
    return this.#hovering
  }
}
```

with this in place, the following will error cuz you're only allowed to modify `.hovering` through the provided methods and `.hovering` isn't exposed to the outside.

```typescript
const cursor = new Cursor()
cursor.hovering = false
```

## added example and updated example

an example has been added that cycles through all the cursor styles. the other example on the `useCursor` page has also been updated. the docs have not been updated yet and will probably come after this pr is merged.


joshwashywash:
i think this could be made a little simpler. i don't think the states in the class really need to be private since if they're changed programmatically, the effect should just rerun. for example if someone hovers over something and then some event fires that changes the `hoveringStyle` to something else, it should just take effect. same idea goes for the `element` and `unhoverStyle` as well.

this should get rid of a lot of weirdness in the constructor remove the `.of` constructor entirely.

gonna update this and make a test to be sure.

packages/extras/src/lib/hooks/useCursor.svelte.ts
```diff
@@ -42,9 +42,26 @@ type Cursor = LiteralUnion<
   string
 >
 
+export class Cursor {
+  hovering = $state(false)
+  element = $state(document.body)
```
DefinitelyMaybe, line 47: I'm unfamiliar cursor classes elsewhere, is a element state to the document body a usual thing?

DefinitelyMaybe (commented):
I'm down with having a cursor class 👍

packages/extras/src/lib/hooks/useCursor.svelte.ts
joshwashywash: hmm good point. i think what would happen is the element would get its cursor style set, but you'd have to have the cursor over that element. the important thing to note is that `Cursor` doesn't have to be used with objects in the scene. if someone has some overlayed html or is using the <HTML> component. they can toggle the `.hovering` state from there. But in order to see the actual cursor style, the cursor has to be over `.element`. i don't think it hurts anything to have the element default to the document body nor be updatable but i also can't think of a case where you would want it to be something other than the document body. maybe there's some 1% of cases where you'd want something different but i don't think we're going out of our way at all to account for that 1%.

DefinitelyMaybe: ok all good 👍

joshwashywash:
@DefinitelyMaybe i think we should take the approach here that `useCursor` is marked deprecated but still available for now so that it doesn't break any existing usages. we should be able to add `@deprecated use the new Cursor class instead` in jsdoc comments. then we can gracefully make it obsolete. that way too the examples that use `useCursor` are still functional until they've been updated. i think once all the examples have been updated, then we can safely remove `useCursor`. somewhere along the way we can put the migration guide in the threlte 8 
blog post as well. what do you think?

michealparks:
Closing since useCursor after recent fixes fully fits the definition of a hook: It shares state among a single Threlte app instance - and in this case it's to prevent hover state race conditions.

PR #1292, opened 2024-12-22
[next]Update use-progress example to svelte 5

joshwashywash:


DefinitelyMaybe (commented):
LGTM

I don't think docs need any changesets but that's just a nit pick

PR #1294, opened 2024-12-23
[next] useTask page examples update

DefinitelyMaybe:


apps/docs/src/content/reference/core/use-task.mdx
```diff
@@ -263,8 +263,12 @@ effects using the library
   }
 
   // We need to set up the passes according to the camera in use
-  $: setupEffectComposer($camera)
```
DefinitelyMaybe, line 266: @joshwashywash would this become an `$effect()` or should it be a `$derived.by()`?

joshwashywash: that one would be an effect. i think we should just inline it so that the name of the function doesn't unintentionally imply that it could be derived. I suppose we could break out the individual passes into their own `derived`s

```typescript
const composer = new EffectComposer(renderer)

const renderPass = $derived(new RenderPass(scene, $camera))

// the bloom effect doesn't change or have any dependencies so we can keep that around
const bloomEffect = new BloomEffect({
  intensity: 1,
  luminanceThreshold: 0.15,
  height: 512,
  width: 512,
  luminanceSmoothing: 0.08,
  mipmapBlur: true,
  kernelSize: KernelSize.MEDIUM
})
const bloomPass = $derived(new EffectPass($camera, bloomEffect))

// smaaEffect never updates / doesn't have dependencies so can keep it around too
const smaaEffect = new SMAAEffect({ preset: SMAAPreset.LOW })
const smaaPass = $derived(new EffectPass($camera, smaaEffect))

$effect(() => {
  composer.addPass(renderPass)
  composer.addPass(effectPass)
  composer.addPass(smaaPass)
  return () => {
    composer.removeAllPasses() // always remove all passes before the effect ever reruns.
  }
})
```

grischaerbe (approved):

Issue #1161, opened 2024-11-02
[next] Migrate from handwritten svelte.d.ts files to generated type definitions


grischaerbe:
Currently, a lot of components are typed by hand with accompanying Comp.svelte.d.ts files. With Svelte 5 we can finally skip this unnecessary step 🎉 

## Process

Currently, these svelte.d.ts files hold all info about component props and exports:

```ts
import type { Box3, Group } from 'three'
import type { Props, Stage } from '@threlte/core'
import type { Component, Snippet } from 'svelte'

export type ResizeProps = Omit<Props<Group>, 'children'> & {
  /**
   * when true, resizes the group when objects are added or removed
   *
   * @default false
   */
  auto?: boolean
  /**
   * axis to constrain by. if not provided, the max of all 3 axes is used.
   */
  axis?: 'x' | 'y' | 'z'
  /**
   * optional Box3 to use in bounding box calculation
   */
  box?: Box3
  /**
   * whether to use precise bounding box calculation
   * https://threejs.org/docs/index.html#api/en/math/Box3.setFromObject
   *
   * @default true
   */
  precise?: boolean
  /**
   * called after the group has been scaled
   */
  onresize?: () => void
  resize?: () => void
  /**
   * The stage used to resize
   */
  stage?: Stage
  children?: Snippet<[{ ref: Group; resize(): void }]>
}

export type ResizeExports = {
  /**
   * Manually trigger a resize
   */
  resize: () => void
}

declare const Resize: Component<ResizeProps, ResizeExports>

export default Resize
```

We want to migrate to using a separate "types.ts" file where currently the svelte.d.ts file is and have the prop definitions in there. In the example above, that looks like this:

```ts
import type { Props, Stage } from '@threlte/core'
import type { Box3, Group } from 'three'

export type ResizeProps = Props<Group, [{ ref: Group; resize(): void }]> & {
  /**
   * If true, will automatically resize when children or added or removed
   * @default false
   */
  auto?: boolean
  /**
   * Specifies which axis to constrain. If not provided, the maximum axis is
   * used
   */
  axis?: 'x' | 'y' | 'z'
  /**
   * Bring your own box to capture the bounding box.
   */
  box?: Box3
  /**
   * If true, use precise bounding box calculation.
   */
  precise?: boolean
  /**
   * Bring your own stage to control when resizing occurs. If not provided,
   * resizing will occur before the main render stage.
   */
  stage?: Stage
  /**
   * A callback function to run whenever resizing is done
[...]

PR #1298, opened 2024-12-23
[next] mirgrate doc types

DefinitelyMaybe:
#1161


apps/docs/src/examples/extras/positional-audio/Disc.svelte
```diff
@@ -15,10 +16,11 @@
       autoStart: false
     }
   )
-  $: {
+
+  $effect(() => {
```
joshwashywash, line 20: i think we can write this $effect as

```typescript
$effect(() => {
  if (discSpeed > 0) {
    start()
    return () => {
      stop()
    }
})
```


then we don't have to reference `$started`

so the task is started if discSpeed > 0. and it's stopped before the effect reruns. but it never needs to be stopped if it was never started.

could go even further and break it out into a derived to prevent additional work.

```typescript
const discSpeedGreaterThanZero = $derived(discSpeed > 0)

$effect(() => {
  if (discSpeedGreaterThanZero) {
    // ...
  }
})
```

if `discSpeed` goes from say 1 to 2 then `discSpeedGreaterThanZero` goes from `true` to `true` and the effect cleanup won't run because its dependencies haven't changed.

apps/docs/src/examples/extras/positional-audio/Scene.svelte
```diff
@@ -6,16 +6,17 @@
...
-  $: smoothVolume.set(volume)
+  $effect(() => {
+    smoothVolume.set(volume)
+  })
 
   const { size } = useThrelte()
 
-  let zoom = $size.width / 18
-  $: zoom = $size.width / 18
+  let zoom = $derived($size.width / 18)
```
joshwashywash, line 19:
```suggestion
  const zoom = $derived($size.width / 18)
```

apps/docs/src/examples/extras/positional-audio/Turntable.svelte
```diff
@@ -77,14 +79,16 @@
...
   }>('/models/turntable/cover.glb')
-  let coverGeometry: BufferGeometry | undefined
-  $: if ($gltf) {
-    const coverMesh = $gltf.nodes.Cover as Mesh
-    coverGeometry = coverMesh.geometry
-  }
+  let coverGeometry: BufferGeometry | undefined = $state()
+  $effect(() => {
+    if ($gltf) {
+      const coverMesh = $gltf.nodes.Cover as Mesh
```
joshwashywash, line 85:
```suggestion
      const coverMesh = $gltf.nodes.Cover
```

it looks like `gltf` gets its type further up in the file with the `useGLTF` function.

apps/docs/src/examples/misc/gaussian-splatting/LumaSplatsThree/LumaSplatsThree.svelte
```diff
@@ -1,17 +1,22 @@
...
+    mode = 'object',
+    loadingAnimationEnabled = false,
+    particleRevealEnabled = false,
+    enableThreeShaderIntegration = true,
+    children,
+    ...rest
+  }: LumaSplatsThreeProps = $props()
 
-  const { invalidate, renderer, scene } = useThrelte()
+  const { invalidate, renderer, scene } = $state(useThrelte())
```
joshwashywash, line 19:
```suggestion
  const { invalidate, renderer, scene } = useThrelte()
```

apps/docs/src/examples/misc/gaussian-splatting/LumaSplatsThree/LumaSplatsThree.svelte
```diff
@@ -57,18 +62,21 @@
...
+    if ($splats && $splats[1]) {
+      previousEnvironment = scene.environment
+      previousBackground = scene.background
+      previousBackgroundBluriness = scene.backgroundBlurriness
+      scene.environment = $splats[1]
+      scene.background = $splats[1]
+      scene.backgroundBlurriness = 0.5
+      invalidate()
+    }
+  })
```
joshwashywash, line 79: this should probably be broken up into multiple `$derived`s and `$effect`s but we can do that in a later pr somewhere else.

joshwashywash (commented):
i think there's still improvement to be had with some of these examples but what this pr addresses with the types and moving things into `$state`s is definitely moving in the right direction.

👍🏻 just some things to look over and it looks good to me otherwise.

apps/docs/src/examples/extras/positional-audio/Scene.svelte
DefinitelyMaybe: why `const`? wouldn't a `$derived` always be `let`?

apps/docs/src/examples/misc/gaussian-splatting/LumaSplatsThree/LumaSplatsThree.svelte
DefinitelyMaybe: I must've been so tired. Good catch.

DefinitelyMaybe:
> i think there's still improvement to be had with some of these examples but we can do tackle that in future prs.

Agreed 👍 merging and will continue to improve

PR #1311, opened 2024-12-29
[next]Xr teleport example svelte 5

joshwashywash:
- convert to svelte 5
- uses more `$derived`s and effects to keep the use task callback very minimal

this is the last example that used the simplex-noise package so i've removed that from `package.json`.


DefinitelyMaybe:
good stuff. LGTM 👍

PR #1319, opened 2025-01-03
[next] Update `useGltfAnimations` examples and docs to svelte 5

joshwashywash:
- updates to svelte 5; both the example and the .mdx page
- updates the example to show how to export the mixer and actions and control them from outside


apps/docs/src/content/reference/extras/use-gltf-animations.mdx
```diff
@@ -99,16 +103,16 @@ This can also be done after the animations have been loaded. In this case, you c
...
-{/if}
+    <T is={scene} />
+  </T.Group>
+{/await}
 ```
 
 ## Types
 
-```ts
+```typescript
```
grischaerbe, line 115: We use `ts` here all the time, what's wrong with it?

apps/docs/src/examples/extras/use-gltf-animations/Scene.svelte
```diff
@@ -1,26 +1,22 @@
...
 
-  const { scene } = useThrelte()
-  scene.background = new Color(0xeae8e2)
+  const dracoLoader = useDraco()
+  const gltf = useGltf('/models/LittlestTokyo.glb', { dracoLoader })
+  const { actions, mixer } = useGltfAnimations<'Take 001'>(gltf)
 
-  let showTokyo = true
+  const s = fromStore(actions)
+  const getActions = () => s.current
```
grischaerbe, line 11: Binding the scene component and creating reactive state from the store seems overly complex to me. You can create the tweakpane UI right in this component, works too 👍

grischaerbe (changes requested):
Thanks for porting that over! There's some unnecessary complexity that we can remove :)

apps/docs/src/content/reference/extras/use-gltf-animations.mdx
joshwashywash: sorry, that one is out of habit. i'll switch it back.

apps/docs/src/examples/extras/use-gltf-animations/Scene.svelte
joshwashywash: i simplified that a little. now the store is exported and the action is derived.

apps/docs/src/examples/extras/use-gltf-animations/Scene.svelte
```diff
@@ -31,9 +22,11 @@
   />
 </T.PerspectiveCamera>
 
-<T.AmbientLight />
-<T.DirectionalLight position={[5, 10, 5]} />
+<Environment
+  url="/textures/equirectangular/hdr/industrial_sunset_puresky_1k.hdr"
+  isBackground
+/>
```
DefinitelyMaybe, line 28: love the added environment ❤️ very nice touch

PR #1325, opened 2025-01-08
[next] <Backdrop> component

joshwashywash:
a port of drei's backdrop component https://drei.docs.pmnd.rs/staging/backdrop

Two new approches

## Export Component Prop Types

i export the Backdrop component props type so that they would be available to the user. this is something that i think we should do for new components and probably go back and do for existing components. my reasoning is that if i want to forward props to a component it's very easy to do the following:

(Backdrop is used in the examples below but it could be any component from @threlte/extras)

```svelte
<Backdrop {...backdropProps} />
```

but what type should `backdropProps` have? exporting the type from extras makes it very easy to give it the correct type.

```svelte
<!-- Scene.svelte -->
<script lang="ts">
import type { BackdropProps } from '@threlte/extras'

type SceneProps = BackdropProps & {
// ... additional scene Props
}

let {sceneProp1, sceneProp2, ...backdropProps}: SceneProps = $props() 
</script>

<Backdrop {...backdropProps} />
```

or if i just want to create an object this is spreadable onto `<Backdrop>`

```svelte
<script lang="ts">
import type { BackdropProps } from '@threlte/extras'

const props: BackdropProps = {
// ...
}
</script>

<Backdrop {...props} />
```

I don't think there's any downside to exporting the component props type like this, it just makes it easier to provide the correct types in some cases. i usually advocate for following in the footsteps of svelte and they make most of their types available to the user - for example `TweenedOptions` for Tween classes and stores from the `svelte/motion` library.

## index.ts file for the component

`/Backdrop/index.ts` re-exports the component and the prop types to make it easier to export from the main `lib/index.ts` file

doing so makes the export from the `lib/index.ts` file become

```typescript
export { default as Backdrop, type BackdropProps } from './components/Backdrop';
```

instead of

```typescript
export { default as Backdrop } from './components/Backdrop/Backdrop.svelte'
export type { BackdropProps } from './components/Backdrop/types'
```

i think this is a more suitable approach if we plan to export component prop types.

## other things about this pr

- there's some unused imports in `MeshDiscardMaterial` that were removed
- feel free to tweak the docs or the example to look or feel better or to show off the component in a better way.

### DIFF
--- a/apps/docs/src/content/reference/extras/backdrop.mdx
+++ b/apps/docs/src/content/reference/extras/backdrop.mdx
@@ -1,0 +1,60 @@
+---
+order: 7.32
+category: '@threlte/extras'
+name: '<Backdrop>'
+sourcePath: 'packages/extras/src/lib/components/Backdrop/Backdrop.svelte'
+type: 'component'
+componentSignature:
+  {
+    extends: { type: 'Group', url: 'https://threejs.org/docs/index.html#api/en/objects/Group' },
+    props:
+      [
+        {
+          name: 'floor',
+          type: 'number',
+          default: '0.25',
+          description: 'length of the floor segment',
+          required: false
+        },
+        {
+          name: 'receiveShadow',
+          type: 'boolean',
+          default: 'false',
+          description: 'whether the backdrop is to receive shadows',
+          required: false
+        },
+        {
+          name: 'segments',
+          type: 'number',
+          default: '20',
+          description: "resolution of the backdrop's geometry",
+          required: false
+        }
+      ]
+  }
+---
+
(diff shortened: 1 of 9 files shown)


michealparks:
@joshwashywash I'd suggest that this instead be `<BackdropGeometry>` for a little more flexibility and to match the approach of similar components like `RoundedBoxGeometry`, `Text3DGeometry`, etc.

If you agree, mind if I take a crack at it?

let-josh:
yes. i agree. it's primarily a geometry so `<BackdropGeometry />` makes the most sense. this is a pretty old pr that may be outdated at this point so take from it what you want. feel free to close this one in favor of a new "BackdropGeometry" pr. i'm not super attached to this one.

PR #1332, opened 2025-01-11
[next] update transitions example

DefinitelyMaybe:


apps/docs/src/examples/animation/transitions/Scene.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { T } from '@threlte/core'
   import Character from './Character.svelte'
+  let { action } = $props()
```
joshwashywash, line 4: should we add the type here?

joshwashywash, line 4: maybeWe should call this `actionKey` as well so that we can do this

```svelte
<Character {actionKey} />
```

that would mean that the `App.svelte` file could also have its `action` state renamed to `actionKey`.

not a big deal but they would all be `actionKey` at that point.

joshwashywash (commented):
looks good to me. mostly just small suggestions.

apps/docs/src/examples/animation/transitions/Scene.svelte
DefinitelyMaybe: yeah, thought about it. Going for good enough

PR #1333, opened 2025-01-11
[next] more 3D-Terrain example improvement and docs

joshwashywash:
biggest difference is that the "flatness" of the surface can be controlled - shows the fact that you may need to do `attribute.needsUpdate = true` and/or recompute the vertex normals after changing the vertex positions. `geometry.computeVertexNormals()`.

adds more informative docs and fixes some typos. the usual stuff.


DefinitelyMaybe:
ship it! LGTM

PR #1343, opened 2025-01-12
[next] Postprocessing outlines example updates

joshwashywash:
- simplifies the effects and signals a little - instead of deriving the passes and outline effect from the camera, an effect is used to set the `mainCamera` property where needed. this means these objects can be normal objects instead of `$derived`s. one effect can handle setting the `mainCamera` for all passes and effects 
- adds a disposal `$effect` for effects, passes, and the composer. all of these objects need to be disposed when the component unmounts
-  simplifies the geometry for the walls by using an extrude geometry. now there's no longer duplicated vertices when two walls touch. now there is a left shape and a right shape which are extruded to look like walls.
- change to ortho camera cuz i think it looks a little cooler and nicer.

### DIFF
--- a/apps/docs/src/examples/postprocessing/outline/Scene.svelte
+++ b/apps/docs/src/examples/postprocessing/outline/Scene.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { Mesh, Vector3Tuple } from 'three'
+  import type { ExtrudeGeometryOptions, Mesh, Vector3Tuple } from 'three'
   import type { Wall } from './types'
   import { DoubleSide } from 'three'
   import { Environment, OrbitControls } from '@threlte/extras'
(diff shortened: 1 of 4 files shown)


apps/docs/src/examples/postprocessing/outline/Scene.svelte
```diff
@@ -17,76 +17,79 @@
...
-    {rotation}
-  >
-    <T.BoxGeometry args={size} />
-    <T.MeshStandardMaterial color="silver" />
-  </T.Mesh>
-{/each}
+  const extrudeOptions: ExtrudeGeometryOptions = { bevelEnabled: false }
+</script>
 
 <Environment url="/textures/equirectangular/hdr/shanghai_riverside_1k.hdr" />
```
DefinitelyMaybe, line 57:
```suggestion
```

DefinitelyMaybe:
looks good!

PR #1346, opened 2025-01-13
[next]Mask example svelte 5

joshwashywash:
the updated example is slightly different than the one currently on `next`, particularly when using the `inverse` prop. i'm not sure which is correct. to see what i mean, you'll probably have to put next.threlte and this pr's preview side by side.

Screenshot 2025-01-12 at 7 43 48 PM

in the picture above `next` is on the left

i think the difference is that in this PR the returned value of a `useMask` call is always spread onto a material whereas on `next`'s version an empty object would sometimes get spread. but i'm not sure if spreading an empty object would cause the stencil props of the mask's material to get reset to what they were before.

might need some more investigation.

i don't think either way is *wrong* but i'm not sure which is to be preferred.

---

i removed the rotation `useTask` cuz i didn't think it was necessary. simplifies things just a tad.

### DIFF
--- a/apps/docs/src/examples/extras/mask/Scene.svelte
+++ b/apps/docs/src/examples/extras/mask/Scene.svelte
@@ -1,38 +1,15 @@
 <script lang="ts">
-  import { T, useTask } from '@threlte/core'
+  import { T } from '@threlte/core'
   import { OrbitControls, Grid, Float, TransformControls, Mask, useMask } from '@threlte/extras'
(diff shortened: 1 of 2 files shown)


DefinitelyMaybe:
shoot... and which one was right 😅

grischaerbe (approved):

grischaerbe:
I think the Svelte 5 example is right, thanks! ❤️

Issue #1341, opened 2025-01-12
Rendering HTML inside XR


HannesGitH:
I have issues with rendering the HTML component inside an XR (VR) session.

While normal meshes are rendered fine, HTML isnt. Its just not visible at all when entering the VR session.
To reproduce this issue an actual VR headset is needed as e.g. the meta WebXR Emulator does not reproduce this issue (seems to be working fine there).
This is because all transformations etc seem to be applied to the canvas, so the preview moves, when the headset moves, but as i said its not displayed in the actual session at all.

Also see the attached video, where you can see that the 'test' html actually is transformed when I move my head, but only in the browser window, as soon as I continue the session in fullscreen it vanishes..

---

Steps to reproduce:
1. `npm create threlte@next` (select xr and extras)

2. replace App.svelte with:
```svelte
<script lang="ts">
  import { Canvas, T } from "@threlte/core";
  import { HTML } from "@threlte/extras";
  import { XR, VRButton } from "@threlte/xr";
</script>

<Canvas>
  <XR></XR>
  <T.Mesh position={[0, 0, -2]}>
    <HTML transform>
      <div>
        <p style="color:white;">test</p>
      </div>
    </HTML>
  </T.Mesh>
</Canvas>
<VRButton></VRButton>
```
3. enter VR session

https://github.com/user-attachments/assets/8eb3bde1-1e22-442f-9879-ba7215fed920

HannesGitH:
discord post: https://discord.com/channels/985983540804091964/1327970808441868370

HannesGitH:
related:  https://github.com/pmndrs/react-xr/issues/21

HannesGitH:
This is not at all threlte's fault, as the WebXR renderer does not show DOM nodes as of its spec. A proposal to add that functionality to WebXR can be found here: https://www.w3.org/TR/webxr-dom-overlays-1/

DefinitelyMaybe:
@michealparks would the [threlte-uikit](https://github.com/michealparks/threlte-uikit) package help in this case? I could add a note about it in the docs if so

michealparks:
HTML cannot be rendered inside an XR environment, this is just a limitation of the WebXR API. @DefinitelyMaybe is correct, threlte-uikit, and UIKit in general, are solutions to help with this problem, and good call! It's not broadly known that HTML just cannot be rendered in an immersive XR session, so adding docs would help.

PR #1349, opened 2025-01-13
[next] mentioning threlte-uikit within the docs

DefinitelyMaybe:
updating both XR getting started page and the HTML component page.

bit of git diff cause I formatted the mdx too but both changes are at the bottom.

related: #1341


michealparks (approved):
❤️

PR #1356, opened 2025-01-14
[next] categories organization

DefinitelyMaybe:
- Audio section moved further down
- Misc moved to bottom
- Performance moved further down
- suspense pages merged with loading category
- `transitions` page moved to bottom of staging category, hoping its easier to pickup on when scrolling the list.

### DIFF
--- a/apps/docs/src/content/reference/extras/align.mdx
+++ b/apps/docs/src/content/reference/extras/align.mdx
@@ -1,5 +1,5 @@
 ---
-order: 7.11
+order: 4.11
 category: '@threlte/extras'
 name: '<Align>'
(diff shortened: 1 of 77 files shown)


joshwashywash:
i like it. i suppose we'd have the most popular pages at the top. it'd be really cool if we could track which categories are the most visited and have them automatically sorted somehow. but that sounds like a lot of work. 🙃

DefinitelyMaybe:
thats a good point. I was going on feel and my understanding of "common" but there's probably a way to track page visits and be more accurate/grounded.

ah yeah, and for the moment I'm not putting my hand up for making that a thing 😅

PR #1362, opened 2025-01-16
Export `SceneGraph` type from extras

let-josh:
i wanted to create a type like the following

```typescript
// types.ts
export type GltfProps<T extends SceneGraph> = {
  children?: Snippet<[{ ref: T }]>;
  gltf: ThrelteGltf<T>;
};
```

then i could use it in a component like so:

```svelte
<script lang="ts">
  import type { GltfProps } from './types'

  type Graph: SceneGraph = {nodes: {}, materials: {}}

  let {
    children,
    gltf = bindable()
  }: GltfProps<Graph> = $props()

  const _gltf = useGltf<Graph>(/* ... */)
  _gltf.then((_gltf) => {
    gltf = _gltf
  })
</script>

{#await _gltf then ref}
  {@render children?.({ref})}
{/await}
```

but the `SceneGraph` type is not exported from extras. you could get the type like the following but it's a lot of work

```typescript
type SceneGraph = ReturnType<typeof useGltf> extends Writable<undefined | ThrelteGltf<infer R>> ? R : never;
```


joshwashywash:
a simpler use case would be having the `Graph` type in a file somewhere but typing it such that it extends SceneGraph

```typescript
// types.ts
import type { SceneGraph } from '@threlte/extras'
export type Graph: SceneGraph = { /* */ }
```

```svelte
<script>
import type { Graph } from './types'
const gltf = useGltf<Graph>()
</script>
```

DefinitelyMaybe:
this looks fine. shall we sort the conflicts and go from there

packages/extras/src/lib/index.ts
```diff
@@ -104,7 +104,7 @@ export {
   type DomEvent,
   type EventMap,
   type Intersection,
-  type IntersectionEvent,
+  type IntersectionEvent,,
```
DefinitelyMaybe:
```suggestion
  type IntersectionEvent,
```

michealparks:
A few concerns with this one: 

- `SceneGraph` is mostly composed of `Record<string, any>` - I'm struggling to see the advantage of exporting this when the type is already very permissive and the user can fill it in flexibly already

- `SceneGraph` is an internal type currently and I don't think has been vetted naming wise. I don't think it's a good name if it's going to be exported for external use. It doesn't really describe the interals (`nodes` and `materials`) well at all

Why not use, or improve, the `GLTF` or `ThrelteGLTF` types here?

let-josh:
you know it's been quite a while. so long that i don't even remember why i wanted this type to be exposed. 🙃

DefinitelyMaybe:
> you know it's been quite a while. so long that i don't even remember why i wanted this type to be exposed. 🙃

:thinking: reopen when remembered :sweat_smile:

PR #1367, opened 2025-01-17
update camera-controls example/docs to describe vite externalization …

joshwashywash:
- updates the docs for the camera controls example to highlight vite externalization may be necessary
- refactors the example to show how to make a component out of camera-controls. this means that the update task has been moved into the component along with the disposal of the controls. i think this is a better idea since it's more clear that `useTask` needs to be called in a component that the class doesn't do anything threlte related - all the threlte-y stuff stays in the component and outside the class.


DefinitelyMaybe:
I'm unsure on this one. I kinda liked the all in one cameraControls class that was a weird mix threlte and svelte in a ts class but then also maybe there's something to not giving an example of mixing the two because ... I dunno, would it lead to weird edge cases?

joshwashywash:
> I dunno, would it lead to weird edge cases?

i dunno. if i remember correctly the example before these changes was a component. i think there's value in that the class doesn't have any threltey things in it. it's not even a svelte.ts file since it doesn't use any runes. someone could copy it and use it for something that's not svelte nor threlte if they wanted to. i think in this case we should keep things like `useTask` and `onDestroy()/$effect` in the component's script tag.

it also looks like something that could be added to extras if it wasn't for the extra dependency that it brings along.

apps/docs/src/examples/camera/camera-controls/Scene.svelte
```diff
@@ -1,45 +1,45 @@
...
-  const camera = new PerspectiveCamera()
-  controls = new CameraControls(camera, dom)
-  controls.setPosition(5, 5, 5)
+  } = $props()
+
+  // only set its initial position
+  $effect(() => {
+    untrack(() => {
+      controls.setPosition(5, 5, 5, false)
+    })
```
michealparks, line 23: Just curious, why's untrack needed here?

michealparks (approved):

apps/docs/src/examples/camera/camera-controls/Scene.svelte
joshwashywash: oops. gotta fix that. the idea was that once the controls is set once, then we can stop watching it. but i don't think that's the case here.

apps/docs/src/examples/camera/camera-controls/CameraControls.svelte
```diff
@@ -0,0 +1,32 @@
...
+    return () => {
+      ref.dispose()
+    }
+  })
+</script>
+
+<T
+  is={ref.camera}
+  makeDefault
+/>
```
DefinitelyMaybe, line 32: I feel uncomfortable about a new default perspective camera created by the ts controls class. Would a camera prop be better or nah?

joshwashywash: yeah it might make more sense to do something like:

```svelte
<script lang="ts">
  const camera = new PerspectiveCamera()
</script>

<CameraControls {camera} />
```

but i think that'd we have to handle the case where the camera updates, say if someone had a toggle button for switching between ortho and perspective cameras. i think that should be possible and easy to add into the existing example. i'll update the code to match and add a "switch-between-cameras" button. it looks like it should be possible given that the `.camera` prop on the controls class is a setter but i want to see it in action.

joshwashywash:
the example has been updated so that the `<CameraControls>` component can now take the camera has a prop. There's been some clean up to the types and some warnings/errors that my editor wasn't reporting for some reason. i think this one is good to go now after checking the docs are all good.

DefinitelyMaybe:
lemme jump into some commits here tomorrow, today's brain is done.

joshwashywash:
@DefinitelyMaybe i think you were right with keeping the `CameraControls` as a class and not making a component out of it. for this particular example it makes for more work cuz you have to pass state up two levels - from CameraControls -> Scene -> App. it's kinda tedious cuz all along the way the controls need to be bindable. without a separate component you can just instantiate the camera controls instance in the Scene and bind to it from the App component. it's a much nicer and simple example that way.

most of the time when it comes to camera stuff, i think it's best to have it happen high in the tree so that you don't have to pass around camera state and functions that control the camera up and down the component tree.

i've edited the example to be like this.

DefinitelyMaybe:
yes and no 😅

frequently enough we come back to wanting camera-controls in our scenes so it'll be super handy to have something in extras.

We do have extras components that require additional installs, [mesh-refraction-material](https://threlte.xyz/docs/reference/extras/mesh-refraction-material) for example

apps/docs/src/examples/camera/camera-controls/CameraControls.ts
```diff
@@ -0,0 +1,38 @@
...
+  Sphere,
+  Spherical,
+  Vector2,
+  Vector3,
+  Vector4
+} from 'three'
+
+export default class CameraControls extends CC {
+  static #installed = false
+  constructor(element: HTMLElement, camera: Camera = new PerspectiveCamera()) {
```
DefinitelyMaybe, line 18:
```suggestion
  constructor(element: HTMLElement, camera: Camera) {
```

DefinitelyMaybe (approved):

Issue #1371, opened 2025-01-20
[next] extras Text - not refresh after change when on demand is on


zadinvit:
Hi, I found a bug in the Text extras component. 

When on-demand is on, and there is only a text change, the value of text is not refreshed. 

I change the little example to not change rotation of text.  So there is no change in the 3d scene when you change the text. 

example: https://stackblitz.com/edit/vitejs-vite-vbbfekzu?file=src%2Fexample%2FApp.svelte

https://github.com/user-attachments/assets/b6b09752-15e3-4e48-b2e2-32b404de360a

zadinvit:
I found the problem with naming (there is twice text), I will create PR with a possible fix.

DefinitelyMaybe:
interesting the docs page example for `<Text/>` is using `<Text3DGeometry/>`

joshwashywash:
fix is in #1373 i think this one is okay to close.

PR #1372, opened 2025-01-20
[next] fix problem with  text update 

zadinvit:
#1371 

Name for text change when the text was migrated to Svelte5. But it seems there is a collision in T props values, which causes the problem mentioned at issue. 

Changing troika text name should fix the bug.


grischaerbe:
Hey @zadinvit, the fix is to listen to all text props that need syncronization, the fix is in #1373, thanks for reporting and your effort! ❤️

PR #1378, opened 2025-01-21
(docs) move `Migration-Guides` page to "More" section

joshwashywash:
i think it makes more sense to put it into the "More" section. "advanced" is for advanced topics in threlte or threejs. migration guide fits more with external slightly unrelated things. not sure if that's a good sell but it feels "off-topic" in a way... if that makes sense.


michealparks (approved):
I like it! Makes it a little easier to hunt down.

PR #1380, opened 2025-01-21
Update three-perf dependency in extras

AlexWarnes:
Updating the `three-perf` dependency to point at the latest version instead of @jerzakm forked repo because his PR into the main package was [merged last month](https://github.com/TheoTheDev/three-perf/pull/3).

Edit: I forgot to mention, StackBlitz cannot install a package that points to a `git` repo, so @threlte/extras will always fail to install in StackBlitz projects while three-perf is pointed at a github repo (so any version after `9.0.0-next.31`)

### DIFF
--- a/.changeset/calm-rockets-search.md
+++ b/.changeset/calm-rockets-search.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
+Point three-perf dependency to latest version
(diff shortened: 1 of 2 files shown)


AlexWarnes:
I'm not sure how to resolve the errors in the install/deployment. Let me know if it's something I can fix on my end!

jerzakm:
Unfortunately theo didn't release three-perf with the fixes, so the npm version is still the old one and breaks threlte. I'll poke him again :|

AlexWarnes:
@jerzakm oh nooo, sorry. I saw the merge and thought it was all good, my mistake!

PR #1382, opened 2025-01-21
use offsetWidth + offsetHeight for dom size calculation

### LINKED ISSUE #1381: Canvas size "sticks" to container transition scale in Threlte 8
Threlte 8 changed the behavior of the canvas to automatically resize to its parent container (https://threlte.xyz/docs/learn/advanced/migration-guides#canvas-component-size-prop)

If a canvas is being "scaled in" using a transition, it's calculating the Canvas size based on the size of the div during the start of transition. After the transition ends, it ends up larger that its container.

Minimal reproducible example:
https://svelte.dev/playground/hello-world?version=5.19.1#H4sIAAAAAAAACm1SS2srIRT-K-faC01giG2hcPGacB-LrrrLrtOFmTmTkTo66EmaIPPfi0qahBYX4uf3OmJkVg3IBFv3Hg0h_IJG2b0K0DhLSlv0QF7ZoEk7CzqEHbKKddpgYOIlMjqOSZ4AVp3M_o7jIuyTH6vYRgX8Dk8BaCkwwWRovB5pVdua9DA6TxDhf-5RwRom6Lwb4PYPlZK8cR5vr8ihUQY_iSWDn4tnskGC0Lt3WMLPQIpw1ikTcF5byc8FrNzsiJwFZxujm7dlnM1huUoZRfsj79O0SjvvdYuSF0mWxxvdZeqUMmWr9xcPKHLPZYwQSHkScL94TFaJCiDLyPmUllwvnjH0n-eM_HOHJ3QDkj8CP1P5BRdA8gsryVu9X9U2ct1NecJAR4P5LtWLxaRHve1JwMOdx-F3wd51S_01tHG-RS_gfjxAcEa3sPWINt9m95rE1riNMrPyk-angK9Kj-1JJ3kpxSpGeCAmyO9wep0-AAWJhMOhAgAA

After the transition finished
- Green border is the container div size.
- Red border is the cavas size.

Image

I'm guessing this part does the resizing:

https://github.com/threlte/threlte/blob/c7e33c4d498701b8557b9b9a3ecdbbe108184cfe/packages/core/src/lib/context/fragments/dom.ts#L18-L22

If it would be using `dom.offsetWidth` and `dom.offesetHeight` for size instead of `dom.getBoundingClientRect()` it should get the dimensions without the transforms. I assume the `x` or `y` values are not needed here.

TimoWilhelm:
This PR changes how the container size is determined. It should be slightly more performant and account for any scale transformations applied to the container.

fixes #1381

### DIFF
--- a/.changeset/happy-ants-flow.md
+++ b/.changeset/happy-ants-flow.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": major
+---
+
+Threlte Canvas size defaults to the offsetWidth and offsetHeight. If you'd like the DOM rect properties you must call dom.getBoundingClientRect()
--- a/.changeset/happy-ants-flow.md
+++ b/.changeset/happy-ants-flow.md
@@ -1,5 +1,5 @@
 ---
-"@threlte/core": major
+"@threlte/core": patch
 ---
 
 Threlte Canvas size defaults to the offsetWidth and offsetHeight. If you'd like the DOM rect properties you must call dom.getBoundingClientRect()
--- a/apps/docs/src/content/reference/core/canvas.mdx
+++ b/apps/docs/src/content/reference/core/canvas.mdx
@@ -65,13 +65,6 @@ componentSignature:
           required: false,
           description: 'To set up a custom renderer, pass a function that returns a new renderer instance.'
         }
-      ],
-    bindings:
-      [
-        {
-          name: 'ctx',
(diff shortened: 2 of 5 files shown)


grischaerbe (approved):

grischaerbe:
I made some documentation changes, apart from that I'd label that as a patch change. Technically it's wrong since the API surface changed and it's not *only* a bugifx, though I highly doubt we have usages of the dom rect. I consider this a bugfix in the end.

PR #1383, opened 2025-01-22
(docs) update the first-scene docs and files to use Spring class

joshwashywash:
- updates both the `step-*/Scene.svelte` files and the corresponding `.mdx` file all to use the Spring class from `svelte/motion` instead of the deprecated `spring` store.
- updates the mdx docs as well.
- removes prettier-ignore directives around code blocks in the `.mdx` file. i don't think they're necessary. the editor can detect and format code blocks for the language that it specifies.


joshwashywash:
i think this will be important to merge in soon as more people see that threlte 8 has been released and may be going through the tutorial. don't want anyone to get the idea that using the spring store means that threlte 8 isn't svelte 5 compatible.

DefinitelyMaybe (approved):

DefinitelyMaybe:
LGTM

PR #1394, opened 2025-01-23
AnimatedSpriteMaterial remix into SpriteSheetMaterial

DefinitelyMaybe:


DefinitelyMaybe:
@michealparks ok this one is ready for an early review if you'd like. I'll ping again after working through my todo's

DefinitelyMaybe:
Ah, and I realize to make this not a breaking change we just leave animatedspritematerial in there but remove it on next extras major. I'm gonna make an issue to list things like this.

DefinitelyMaybe:
Thinking about this, would it be a better idea to have a static spritesheetmaterial component and then have the animatedspritematerial component use the static component

DefinitelyMaybe:
need to update docs here. Was thinking enums where a nice tool to use here but [actually no](https://bsky.app/profile/robpalmer.bsky.social/post/3lgvwb44um22r), doing something like `const fruit = Object.freeze({apple:0,banana:1,grapes:2})` gives you the same thing in a standard way. If TC39 decides to make it simplier then we'll change it then.

PR #1395, opened 2025-01-23
Fix `<Text>` published types

jirihon:
Fixes error reported by TS 5.7: `Cannot access ambient const enums when 'verbatimModuleSyntax' is enabled.` which is caused by wrong import in published `Text.svelte.d.ts` file.

It is enough to rename the troika import to `TroikaText`. I suspect this is some edge case for the Svelte package builder which gets confused when the imported symbol has the same name as the component itself. I could also create an issue in Svelte repo, but I'm not sure if Svelte is even the right place to report it as it might come from some thirparty tool used to build the package. Would be great to get some advice here.

```ts
// Published Text.svelte.d.ts
import { Text } from 'troika-three-text'; // <- This is the wrong import
import type { TextProps } from './types';
declare const Text: import("svelte").Component<TextProps, {}, "ref">;
export default Text;
```

### DIFF
--- a/.changeset/wise-comics-wonder.md
+++ b/.changeset/wise-comics-wonder.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": patch
+---
+
+Fix `<Text>` published types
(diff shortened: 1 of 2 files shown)


grischaerbe (approved):
Oh, I remember we had that problem before with several components, good catch! ❤️ 
It's probably not bad to file an issue in the Svelte repo as well 👍

PR #1418, opened 2025-01-28
Remove lib/classes and migrate some doc components to svelte 5

DefinitelyMaybe:


DefinitelyMaybe:
updated since #1464

apps/docs/src/components/ComponentSignature/ComponentSignature.astro
```diff
@@ -99,10 +98,10 @@ const id = Astro.props.idOverride ?? 'component-signature'
                 <h3 class="mt-0">Props</h3>
 
                 <div
-                  class={c(
+                  class:list={[
```
michealparks, line 101: should this just be `class={[`?

DefinitelyMaybe: my understanding is that astro does `class:list={[...]}` where as svelte does `class={[...]}` but they might've changed since last i looked 🤔 

edit: yeah, no, astro is [`class:list`](https://docs.astro.build/en/reference/directives-reference/#classlist)

michealparks: ah! thanks for the docs link

michealparks (approved):

PR #1419, opened 2025-01-28
docs: convert instanced mesh example using tweened stores to s5 Tween class

joshwashywash:
updated the example to be a little simpler and maybe more exciting. example now shows how to interpolate colors with a svelte Tween.

other things that don't really matter such as no double for loop but using 1D to 2D coordinate conversion.

To see the new example, go to the instanced-mesh docs page and mouse over the balls in the second example.


DefinitelyMaybe:
eeeeeeehh, didn't the old one have color+position interpolation. Is it possible to do that one with the new Tween class?

joshwashywash:
i could probably come up with something. now exactly the same we can get some movement.

joshwashywash:
the difficult thing is that you have to move the ball so that the pointer is still considered over the mesh. maybe scaling it and repositioning it at the same time can work. i'll give that a shot.

joshwashywash:
i tied the y position and scale of the sphere's color. this avoids having another Tween instance for the scale which would actually mean one extra instance per ball.

Using some fancy vector comparisons and stuff - feels a lot like what you'd do in a shader the ball's scale is linearly interpolated from 1 to 2 based on how close the ball's color is to the "end color". the ball's position is just a translation of the ball's scale.

i think it's a lot more interesting now.

@DefinitelyMaybe let me know what you think.

apps/docs/src/examples/extras/instanced-mesh/colors/BallInstance.svelte.ts
```diff
@@ -0,0 +1,20 @@
...
+  y = new Tween(0, { easing: cubicOut, duration: 250 })
+  scale = $derived(this.y.current + 1)
+  startColor = new Color()
+  endColor = new Color()
+  color = $derived(this.startColor.clone().lerpHSL(this.endColor, this.y.current))
+  constructor(
+    startColor: Color,
+    endColor: Color,
+    public x: number,
+    public z: number
```
DefinitelyMaybe, line 15: Is this a shorthand for setting the x and z class props?

DefinitelyMaybe (approved):

DefinitelyMaybe:
much better! 👍 good stuff

apps/docs/src/examples/extras/instanced-mesh/colors/BallInstance.svelte.ts
joshwashywash: yeah

```typescript
class Person {
    constructor(public name: string) {}
}
```

becomes

```javascript
class Person {
    constructor(name) {
        this.name = name;
    }
}
```

https://www.typescriptlang.org/play/?ssl=3&ssc=2&pln=1&pc=1#code/MYGwhgzhAEAKCmAnCB7AdtA3gKGn6w6EALogK7DEqIAUADmQEYgCWw0aYAtvAFzQlELNAHMAlFgC+2SUA

PR #1426, opened 2025-02-04
Fix `<HTML>` auto invalidation

jirihon:
Fixes broken on-demand rendering when using the `<HTML>` component.

When `<HTML>` was present in the scene, on-demand rendering stopped working because it invalidated the rendered frame on every tick.

The `autoInvalidate` option of the `<HTML>` render task was at the default `true` value which I think does not make sense. This PR sets the option to `false`.

There is probably room for further optimization of the `<HTML>` component. When `autoRender` is set to `true`, its `render` function runs on every tick, whether it is needed or not. I tried returning early from the `render` function using `shouldRender` from the threlte context, but without success. It still runs on almost every tick.


grischaerbe:
Hey @jirihon, thanks for your contribution! ❤️ 
To push this PR a bit further, we can make some adjustments. I would argue that the function passed to `useTask` should be part of the `renderStage`, right? In that function, we can use the `shouldRender()` flag provided by `useThrelte` to skip computions when there's no need. Do you agree on this approach? This would make this component a lot more computationally inexpensive. Basically like this:

```ts
  const { renderStage, shouldRender } = useThrelte()

  export const { start: startRendering, stop: stopRendering } = useTask(() => {
    if (!shouldRender()) return
    render()
  }, {
    autoStart: false,
    autoInvalidate: false,
    stage: renderStage
  })
 ```

grischaerbe (approved):
Sorry @jirihon for the late reply. Yes you're right, when using the `renderStage`, it will only ever be re-evaluated when the frame is dirty 👍

PR #1428, opened 2025-02-05
Studio quick fixes

DefinitelyMaybe:


grischaerbe:
I pushed some more Svelte 5 migrations, I think I catched them all :)

DefinitelyMaybe:
@grischaerbe would you like a changeset for this PR or can we merge?

PR #1429, opened 2025-02-06
docs: update sky component example to svelte 5

joshwashywash:


DefinitelyMaybe (approved):
LGTM, there's a modified `apps/docs/src/examples/studio/getting-started/Scene.svelte` file in here too. Keep it or revert it?

joshwashywash:
oh woops. i'm not sure if that file is even used anywhere but i'll remove it

joshwashywash:
actually the spring that was in that file wasn't used anywhere so i think it's a safe change to include.

PR #1432, opened 2025-02-09
docs: add bvh plugin example

joshwashywash:
i'm pretty sure this plugin and example work as intended but it doesn't necessarily show the full power of the plugin and the three-bvh-mesh plugin. not sure if we should keep it as is or try to come up with something that pushes the raycasting a little harder and demonstrates the performance a little better. 🤷🏻

### DIFF
--- a/apps/docs/src/content/examples/plugins/bvh.mdx
+++ b/apps/docs/src/content/examples/plugins/bvh.mdx
@@ -1,0 +1,39 @@
+This example demonstrates how to write and use a threlte plugin to apply the necessary functionality from [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) for faster spatial queries and raycasting.
+
+<Example path="plugins/bvh" />
+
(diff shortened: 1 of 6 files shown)


DefinitelyMaybe:
idea to add to the mix - a carbon copy of the [gif at the top of the bvh readme](https://github.com/gkjohnson/three-mesh-bvh).

Perhaps setup with toggles and params to switch between normal vs bvh; and simple polygon vs complex bunny model.

joshwashywash:
yeah i think an example that shows the performance difference between using bvh raycasting and native raycasting a little better is what will really make this example worth having. this pr will probably have to sit around until i have some extra time to work on it.

DefinitelyMaybe:
hey josh, micheal did some recent work on [bvh](https://threlte.xyz/docs/reference/extras/bvh) and made it into the extras package, could we close this PR and focus improvements there?

let-josh:
oh yeah definitely

PR #1435, opened 2025-02-17
Update secure context check getXRSupportState.ts

msub2:
Solely checking for `https` causes an address like `http://localhost:5173` to be treated as 'insecure', despite the fact that localhost is indeed a sufficient secure context for WebXR to run in.

### DIFF
--- a/packages/xr/src/lib/lib/getXRSupportState.ts
+++ b/packages/xr/src/lib/lib/getXRSupportState.ts
@@ -10,7 +10,7 @@ export const getXRSupportState = async (
     return 'unsupported'
   }
 
-  if (location.protocol !== 'https:') {
+  if (!window.isSecureContext) {
     return 'insecure'
   }
 


DefinitelyMaybe:
oh cool! TIL 👍

michealparks:
Thanks @msub2! Good catch. Could you [add a changeset](https://github.com/threlte/threlte/blob/main/CONTRIBUTING.md#using-changesets) when you get a second? That'll allow us to ship to NPM.

msub2:
> Thanks @msub2! Good catch. Could you [add a changeset](https://github.com/threlte/threlte/blob/main/CONTRIBUTING.md#using-changesets) when you get a second? That'll allow us to ship to NPM.

Sure thing! I'll try and get that in tonight.

DefinitelyMaybe (approved):

PR #1437, opened 2025-02-19
update ImageMaterial example

### LINKED ISSUE #1436: fix(docs): ImageMaterial example uses deprecated `WebGLMultipleRenderTargets` and needs to be converted to svelte 5
will have a fix for this coming tomorrow or in the near future.

joshwashywash:
fixes #1436

## changes

- `WebGLMultipleRenderTargets` is deprecated so i swapped it out for `WebGLRenderTarget` with a `count` parameter.
- updated components to use svelte 5 syntax
- moved everything into `<Scene/>` to not have to deal with "global" state.
- moved `BentPlaneGeometry` into a svelte class
- moved card state into a Card class
- moves the "texture renderer" code into a task that stops and start if the texture override is enabled - makes things a little more performant
- move "textures" scene into a hud so that it follows the viewport more accurately

## other

the saturation prop is a valid prop on `<ImageMaterial>` but i can't get it to do anything different. it's the same way on the [live version](https://threlte.xyz/docs/reference/extras/image-material) so i'm not sure if that's intended or if something is missing.


joshwashywash:
it's hard to tell. `saturation` is used in only one place the `ImageMaterial` component https://github.com/threlte/threlte/blob/8d5b45e0b3290085b93749d3200d5090ee63b486/packages/extras/src/lib/components/ImageMaterial/ImageMaterial.svelte#L138 but it does effect the shader. i'm gonna leave it alone for now. i might make a difference in a different example but 🤷🏻 .

PR #1446, opened 2025-03-11
Fix portal components not being removed from `scene`

michealparks:
### Overview

In a recent project I discovered that a bug was introduced to Threlte's `<Portal>` component at some point causing portaled objects to not be removed from the scene on unmount. This PR fixes this bug and refactors portals to be a bit simpler and use runes.


DefinitelyMaybe (approved):
svelte 5 ftw 👍

Issue #1443, opened 2025-03-08
[feat] Custom fade point of origin in Grid


Mystler:
It would be great if the distance fading of the Grid would not exclusively scale from the projected point of the camera onto the grid plane but also allow using other (projected) points. This could be an arbitrary point passed in as a property, even though in my particular case I would specifically be interested in supplying the target vector of my OrbitControls.

I did have a quick peek at the code and it seems like it would only require an adjustment to use a property instead of the default camera position inside the component task. I don't know if there are other reasons or complications for or against a change like this.

DefinitelyMaybe:
on a side note, does tsl seem like a thing that'd help here?

Mystler:
I've not really used TSL so far but I also don't think it's relevant for this particular kind of change? If I'm not mistaken, it's just about a way to affect a uniform that is passed to the shader in the existing setup.

PR #1449, opened 2025-03-12
feat: Add fadeOrigin parameter to Grid

Mystler:
This is a simple attempt at #1443 and allows overriding the fading point of origin to not just be the camera positon. Custom things like fading around an OrbitControl's lookAt vector or offsets thereof can be achieved more easily with this.

Let me know if there are any problems with this or additional steps are required to get this in.

### DIFF
--- a/.changeset/rotten-forks-refuse.md
+++ b/.changeset/rotten-forks-refuse.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": minor
+---
+
+Add fadeOrigin parameter to Grid
--- a/apps/docs/src/content/reference/extras/grid.mdx
+++ b/apps/docs/src/content/reference/extras/grid.mdx
@@ -32,6 +32,13 @@ componentSignature:
         { name: 'fadeDistance', type: 'number', default: '100', required: false },
         { name: 'fadeStrength', type: 'number', default: '1', required: false },
         {
+          name: 'fadeOrigin',
+          type: 'Vector3',
+          default: 'undefined',
+          required: false,
+          description: 'Set a custom fading point of origin. The current camera position will be used if unset.'
+        },
+        {
           name: type,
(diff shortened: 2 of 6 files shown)


michealparks:
Great addition to grid 💯 I could use this as well and I don't think anyone would have issue with adding this.

Three small requests and I think this should be good to go:
* Can you [add a changeset](https://github.com/threlte/threlte/blob/main/CONTRIBUTING.md#using-changesets)? That'll allow us to ship to NPM.
* Can you add `fadeOrigin` to the configurable tweakpane options in the grid example ([here](https://github.com/threlte/threlte/blob/main/apps/docs/src/examples/extras/grid/App.svelte))? It would be helpful to be able to play with this option in the docs
* Can you run `pnpm format` on the `packages/extras` repo? That'll fix lint not passing here in CI

Mystler:
I wasn't sure how to best do the fade origin in the demo. I decided to add a Transform Control to move it around when a "use fade origin" checkbox is set. This should work fine, however, I do need to note that there seem to be bugs with TransformControl (not sure if Threlte or upstream) that come into play here:
- Putting TransformControl into a svelte `{#if useFadeOrigin}` does not appear to destroy the component properly, so that was not an option.
- It appears that the control still somehow reacts to drag movements around the 0,0,0 origin even when disabled, causing the position to have NaN or Infinity values and the whole control breaking down. I tried various approaches including explicitly enabling/disabling it but could not find a workaround.
I decided to push this as is anyway, since it's not a problem coming from my changes, but if you think these bugs are too bad for the demo, let me know.

Hopefully the prettier formatting is fixed now. I also added a simple changeset for threlte/extras.

DefinitelyMaybe (approved):
nice, happy to hear you did a bunch to try get it working nicer. In this case I'm happy enough with the current state 👍

michealparks:
Same! Thanks for this! Merging.

PR #1452, opened 2025-03-18
fix: Better TransformControl for Grid fadeOrigin demo

Mystler:
When implementing the demo for this the other day I ran into problems with the disposal of the TransformControl. However, after going at it fresh now, it seems to work as intended just fine. This fixes an issue with the control potentially causing infinity or NaN values when hidden around 0,0,0 while the user rotates the OrbitControls.


michealparks (approved):
Thank you!

PR #1458, opened 2025-04-09
Bump troika-three-text in extras. 

shot-codes:
Addressing getter/setter changes of customDepthMaterial in threejs 0.175.0. See https://github.com/protectwise/troika/commit/78e00b512eb9618623d248070ba75ea0c55803ee


grischaerbe:
Thanks @shot-codes, I decided to bump a lot of dependencies in #1464, so this is one of them! ❤️

PR #1473, opened 2025-04-24
fix: interactivity pointer state not updating for touch movement

wmurphyrd:
I was looking into an issue where the pointer state from  `useInteractivity` would not update while performing a touch-and-drag type interaction and I found [this SO answer](https://stackoverflow.com/a/48254578) referencing this passage from MDN:

>The pointermove event is fired when a pointer changes coordinates, **and the pointer has not been canceled by a browser touch-action**.

Tried out `touch-action: none` on the canvas style and the issue was indeed resolved. 

I'm not 100% on this exact solution - it would change behavior for non-fullscreen canvases in that touch drag over them would no longer scroll the page. It might also be nice to only apply this when interactivity plugin is installed, but I didn't see any existing mechanism for affecting that sort of change.


grischaerbe:
Thanks @wmurphyrd, but I think this change is too intrusive. What we can do however is make automatically adding touch-action: none an option for the interactivity plugin?

```ts
interactivity({ manageTouchAction: true })
```

Something like that?

In the setup of the interactivity plugin, you would get the canvas element and set that to `touch-action: none` and reset it `onDestroy`. Is that a suitable solution for you?

wmurphyrd:
@grischaerbe ah yes I hadn't thought of modifying the canvas DOM element directly. I can do that.

I've been trying to think of other approaches so that interactivity would just work across platforms by default without interfering with page scrolling. Looks like another option would be to have interactivity plugin subscribe to `"touchmove"` events (in addition to `"pointermove"`). These will fire even when the browser is also using the touch for page scrolling

grischaerbe:
@wmurphyrd I'm going to close this one since it's not a viable solution, I hope you're okay with this :) If you're going for any of the mentioned solutions, feel free to open another PR 🚀

wmurphyrd:
:/ I was going to update it after you let me know if you're interested in the pointermove option

grischaerbe:
Oh sorry, I was a bit too fast with this one! Yes, please feel free to make a suggestion :)

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -217,12 +237,14 @@ export const setupInteractivity = (context: InteractivityContext) => {
...
       target.removeEventListener(eventName, getEventHandler(eventName))
     })
+    target.removeEventListener('touchmove', transformTouchMove)
   }
 
   const connect = (target: HTMLElement) => {
     DOM_EVENTS.forEach(([eventName, passive]) => {
       target.addEventListener(eventName, getEventHandler(eventName), { passive })
     })
+    target.addEventListener('touchmove', transformTouchMove, { passive: true })
```
wmurphyrd, line 247: Why don't I just add touchmove to DOM_EVENTS? TouchEvent has a very different interface from PointerEvent, so adding it to the `type DomEvent = PointerEvent | MouseEvent | WheelEvent` union would require a lot of downstream changes including a breaking change to users since the native event is passed along

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -7,6 +7,26 @@ function getIntersectionId(event: Intersection) {
...
+    return
+  }
+  const syntheticPointerEvent = new PointerEvent('pointermove', {
+    clientX: touch.clientX,
+    clientY: touch.clientY,
+    bubbles: event.bubbles,
+    cancelable: event.cancelable,
+    pointerType: 'touch'
+  })
+  event.target.dispatchEvent(syntheticPointerEvent)
```
wmurphyrd, line 27: Why dispatch again instead of just passing the synthetic pointermove event along in place of the native touch event inside the event handler? The interactivity code depends on offsetX/offsetY properties, but these cannot be set via JS; the event has to be dispatched to cause them to be calculated and set

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -217,12 +237,14 @@ export const setupInteractivity = (context: InteractivityContext) => {
     DOM_EVENTS.forEach(([eventName]) => {
       target.removeEventListener(eventName, getEventHandler(eventName))
```
wmurphyrd, line 238: FYI there's a minor leak here - removeEventListener removes handlers by reference, but getEventHandler returns a new function every time it is called, so it will never match the previously added handler. 

Probably inconsequential as the canvas is not often unmounted, but I could send a fix if you're interested

wmurphyrd (commented):
@grischaerbe ok check this out - will work by default on touch devices without interfering with browser touch actions

wmurphyrd:
Rebased on current main and updated PR description for new approach. Let me know what you think

packages/extras/src/lib/interactivity/setupInteractivity.ts
```diff
@@ -7,6 +7,26 @@ function getIntersectionId(event: Intersection) {
...
+ * Dispatches a synthetic pointermove event for touchmove interactions
+ * because the browser may not send pointermove events for touch, depending
+ * on the canvas touch-action css property.
+ */
+function transformTouchMove(event: TouchEvent) {
+  const touch = event.touches[0]
+  if (!touch || !event.target) {
+    return
+  }
+  const syntheticPointerEvent = new PointerEvent('pointermove', {
```
michealparks, line 20: Would this in some cases - depending on browser and device - cause `pointermove` events to fire twice due to `pointermove`'s presence in `DOM_EVENTS`?

wmurphyrd: hmm, so let's say you set `pointer-events: none;` on the canvas and then used interactivity with this patch, yes I think you would get double pointermove events. I don't think there's any other way to cause it since anything that can send touchmove would also be cancelling the pointer event by default.

One option would be to listen for [pointercancel](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event), so we know specifically when the pointermove events will cease and our synthetic substitutes are needed. Would have to track that over the lifecycle of the interaction, perhaps through InteractivityContext. My instincts tell me this would be overly complicated.

Maybe the solution to this is in documentation rather than code, i.e. a prominent note that if you want touch-and-drag you have to disable the browser touch action on the canvas.

packages/extras/src/lib/interactivity/setupInteractivity.ts
wmurphyrd: Will do :+1:

michealparks: Fixed in https://github.com/threlte/threlte/pull/1526

michealparks:
@wmurphyrd I've added a section detailing this to the docs: https://threlte.xyz/docs/reference/extras/interactivity/#touch-interactions

Let me know if you still think this is insufficient for helping folks with this problem and I'll reopen!

PR #1482, opened 2025-05-07
gltf asset pipeline

### LINKED ISSUE #1477: Model Pipeline Example in Docs
With Threlte 8 we deprecated `npm create threlte`. The one thing this template got quite right was the model pipeline script. We should probably expose that in the Docs for folks to get an idea of how to set something like that up.

grischaerbe:
Fixes #1477

### DIFF
--- a/apps/docs/src/content/reference/gltf/getting-started.mdx
+++ b/apps/docs/src/content/reference/gltf/getting-started.mdx
@@ -56,17 +56,20 @@ npx @threlte/gltf@latest /path/to/Model.glb [options]
 ## Example
 
 <Tip type="note">
+  {' '}
   This example assumes you have your model set up and exported from an application like
   [Blender](https://www.blender.org/) as a GLTF file.
 </Tip>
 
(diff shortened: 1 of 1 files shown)


apps/docs/src/content/reference/gltf/getting-started.mdx
```diff
@@ -310,4 +330,163 @@ If you want to use the component [`<Suspense>`](/docs/reference/extras/suspense)
...
+  // remove the file from /static/models
+  try {
+    unlinkSync(path)
+  } catch (error) {
+    console.error(`Error removing file: ${error}`)
+  }
+})
+```
+
+Place this script in `scripts/transform-models.ts` and run it with `npx tsx scripts/model-pipeline.ts`.
```
DefinitelyMaybe, line 492: could possibly provide a link to tsx, I didn't know about this package until now 😅

DefinitelyMaybe (approved):
Good enough 👍 add it

PR #1500, opened 2025-05-09
End XR session when XR component unmounted

### LINKED ISSUE #1499: XR session left in broken state when scene unmounted
If a user navigates away from the Threlte scene via Sveltekit/SPA navigation while XR is presenting, the XR session is left active but no longer updating so the last frame is glued to your face. 

https://github.com/user-attachments/assets/d2fdcb6c-4da8-480e-be27-ef2003286393

wmurphyrd:
Fix #1499 

Tested on emulator and Quest 3

### DIFF
--- a/.changeset/nice-icons-bow.md
+++ b/.changeset/nice-icons-bow.md
@@ -1,0 +1,5 @@
+---
+"@threlte/xr": patch
+---
+
+Fix an issue where the XR session was left presenting after navigating away from the Threlte scene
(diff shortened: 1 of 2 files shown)


grischaerbe (approved):
Thanks @wmurphyrd for your contribution! It looks good to me, @michealparks wdyt?

michealparks (approved):
Looks good, thank you for fixing this!

PR #1506, opened 2025-05-12
[docs] add astro check and fix astro check issues

DefinitelyMaybe:


apps/docs/src/examples/core/three-arcade-game/game/hooks/useArenaCollider.ts
```diff
@@ -8,11 +8,8 @@ export const useArenaCollisionEnterEvent = () => {
     easing: cubicOut
   })
 
-  const onCollision = (event: CollisionEnterEvent) => {
-    if (
-      !event.detail.targetRigidBody ||
-      event.detail.targetRigidBody?.handle !== game.ballRigidBody?.handle
-    ) {
+  const onCollision: CollisionEnterEvent = (event) => {
```
DefinitelyMaybe, line 11: When we fix the rapier issue I'll come back and check this fix

DefinitelyMaybe: sorted

PR #1515, opened 2025-05-31
Fix transition reference

emmanuel-ferdman:
## PR Summary
This small PR fixes the transition reference. Relevant page: https://threlte.xyz/docs/reference/extras/transitions#global-transitions


DefinitelyMaybe (approved):
Good catch! thank you 👍

PR #1516, opened 2025-06-08
Pnpm catalogs for easier dependency version management

jerzakm:
Pnpm catalogs are a really awesome feature that should make monorepo DX a lot nicer when updating dependencies.

I made an example with just the "three" package here. The global list of versions of packages is stored in `pnpm-workspace.yaml`:

```yaml
packages:
  - 'packages/*'
  - 'apps/*'

catalog:
  three: ^0.175.0
  ```
  
  Then, instead of typing exact version in each package, we can use "package:" directive - similar to something we do with the "workspace:*" already.
  
  ```diff
  + "three": "catalog:"
  - "three": "^0.175.0"
  ```

https://pnpm.io/catalogs


michealparks (approved):
Recently heard about this, love it and would be great to move a lot more stuff to this!

DefinitelyMaybe:
cool! there have been quite a few times of project wide dep updates. Some streamlining there sounds great 👍

jerzakm:
How should I changeset it (if at all?). This is at most a "patch" for all affected packages, right? Ready to press the merge button once you help me figure this out 🚀

michealparks:
The only changes are to devDependencies so no changeset necessary 🚀

PR #1521, opened 2025-06-14
[core] Don't recreate useThrelte objects on each call

michealparks:
### Overview

This PR is a slight performance improvement for `useThrelte`. Context objects are now only created once per `createThrelteContext` call instead of for every `useThrelte` call.

### DIFF
--- a/.changeset/popular-spoons-hide.md
+++ b/.changeset/popular-spoons-hide.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": patch
+"@threlte/core": patch
+---
+
+Don't recreate useThrelte objects on each call


packages/core/src/lib/context/compounds/useThrelte.ts
```diff
@@ -86,39 +83,5 @@ export type ThrelteContext<T extends Renderer> = {
...
-    get scene() {
-      return sceneCtx.scene
-    },
-    set scene(scene) {
-      sceneCtx.scene = scene
-    }
-  }
-
-  return context
+  return getContext<ThrelteContext<T>>('threlte-context')
```
DefinitelyMaybe, line 86: its giving clear and simple, and I like both 👍

grischaerbe:
Good call @michealparks, no need to recreate every time 👍 
I see that the visual output of `<View>` in the example is slightly different, possibly an aspect ratio problem?

DefinitelyMaybe:
looks like the view issue is css related. If we add `w-full` to canvas it might work as intended

DefinitelyMaybe:
ooooo actually yeah, no, not a css issue. I added `w-full` css and still got this:

https://github.com/user-attachments/assets/717d9679-c210-42f0-9075-fb293b6df685

PR #1522, opened 2025-06-15
[extras] Add `<CameraControls>` component

michealparks:
### Overview

Contributors have had mixed feelings about whether to add a CameraControls component for a while, but I've found myself copy-pasting the component added here from project to project for a long time now without any alteration. That to me signals that this is a good candidate for an extra. I think that this would be pretty valuable to myself and others so I'm bringing back up the proposal!


kitschpatrol:
Nice, just wanted to express support for this!

Definitely find myself schlepping around a similar component from project to project. 😁

DefinitelyMaybe (approved):
love it ❤️ It fits with the melting pot of extras really well 😄

PR #1525, opened 2025-06-17
[extras] Add `<SVG>` and `<Decal>` components

michealparks:
🎉


DefinitelyMaybe (approved):
great stuff 👍 

I updated one of the variable names, initially my brain was like `$svelte`? we're waiting on svelte here... oh the icon.

PR #1530, opened 2025-06-20
Creating UI page

### LINKED ISSUE #1527: Add Creating UI page
Identified in #1519

The new page should discuss shared state and provide an simple UI structure setup. I like the idea of putting the new page into the Learn section under the advanced category.

Any thoughts, queries, concerns?

DefinitelyMaybe:
closes #1527

### DIFF
--- a/apps/docs/src/content/config.ts
+++ b/apps/docs/src/content/config.ts
@@ -96,7 +96,7 @@ export const referenceCollection = defineCollection({
 export const learnCollection = defineCollection({
   schema: z.object({
     schemaType: z.string().default('learn'),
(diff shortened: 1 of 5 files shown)


grischaerbe:
Thanks @DefinitelyMaybe although I'm not sure this page is needed. It links to a lot of Svelte pages which tells me that this info is related to general Svelte knowledge rather than being Threlte-specific (when it comes to sharing state between components).

DefinitelyMaybe:
this is true. It's creation was born out of a user issue earlier but they could have needed to brush up on their svelte skills too. Happy to close for now

PR #1536, opened 2025-07-04
Change renderer effects to effects.pre

### LINKED ISSUE #1535: ToneMapping.set used in onMount is overwrite with renderer effect
Hi, 
I have a component, and onMount on this component, I want to change toneMapping (in some cases, it's mounting at the same time as Canvas). 

```
      onMount(() => {
		scene.environmentIntensity = 0.8;
		toneMapping.set(NeutralToneMapping);

		return () => {
			scene.environmentIntensity = 1;
			toneMapping.set(AgXToneMapping);
		};
	});
```

In this case toneMapping.set value is not applied, because it's called before the effect in renderer.svelte.ts is called (the effect is called after the onMount function)

`$effect(() => {
        context.toneMapping.set(options.toneMapping ?? AgXToneMapping);
    });`

The solution is to make all these effects in renderer.svelte.ts as effect.pre (I rewrote it in node_modules and it worked). I can create a PR with this change. 

Workaround:

- renderer.toneMapping = NeutralToneMapping  - this works good,
-  I can, in that case, set different tonemapping in Canvas options.

zadinvit:
fix #1535 

Apply the value from options or the default value in effect.pre to avoid overwriting values in the onMount function of the children components.

### DIFF
--- a/.changeset/ninety-planets-own.md
+++ b/.changeset/ninety-planets-own.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": patch
+---
+
+Change renderer.svelte.ts effects to effects.pre
--- a/packages/core/src/lib/context/fragments/renderer.svelte.ts
+++ b/packages/core/src/lib/context/fragments/renderer.svelte.ts
@@ -175,19 +175,19 @@ export const createRendererContext = <T extends Renderer>(
     }
   })
 
-  $effect(() => {
+  $effect.pre(() => {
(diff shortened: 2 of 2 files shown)


michealparks:
Looks good, can you add a [changeset](https://github.com/threlte/threlte/blob/main/CONTRIBUTING.md#using-changesets)?

zadinvit:
@michealparks I think I added it before? Is there any mistake in my change set?

PR #1542, opened 2025-07-06
[extras] add <ShadowMaterial> component

joshwashywash:
a port of [drei's Shadow component](https://drei.pmnd.rs/?path=/story/misc-shadow--shadow-st)

not sure what's the best position for the docs page is in the side panel so the `order` frontmatter prop may need to be changed.

another idea i had was to add a "create shadow" section to the RadialGradientMaterial docs page since that's essentially what the \<Shadow> is. RadialGradientMaterial is a material instead of a mesh (\<Shadow> is a mesh) which means you potentially wouldn't have to deal with z-fighting is much and could apply it as a material. in the end it would be a lighter more versatile implementation but i don't think it'd be as discoverable as if it were its own component.


michealparks:
Awesome! Think that this should be a material, like ImageMaterial, FakeGlowMaterial, etc? then we could have the advantage of it being able to be added to an instanced or batched mesh. I'm not seeing anything in the component specific to the mesh or geometry

let-josh:
i agree with you. i think it would be more suitable as a material than as a mesh. i'll work on getting it modified later today.

let-josh:
So after working at it for a bit, i think <Shadow> should really just be a section in the `RadialGradientTexture` docs cuz the component as a material would reduce all the way down to this:

```svelte
<!-- ShadowMaterial.svelte -->

<script>
  let {
    color,
    offset,
    ...restProps,
  } = $props();
  
  const colorString = $derived(new Color(color).getStyle());
  
  const stops = $derived([
    {
      color: colorString,
      offset,
    },
    {
      color: 'rgba(0,0,0,0)', // the transparent part
      offset: 1,
    }
  ]);
</script>

<T.MeshBasicMaterial {...restProps}>
  <RadialGradientTexture width={128} height={128} {stops} />
</T.MeshBasicMaterial>
```

the only thing that the user would have to do is set the depthWrite, opacity, transparent etc on the component rather than relying on it being set during inialization.

in practice it would look like this:

```svelte
<!-- Scene.svelte -->

<script>
  const stops = [
    {
      offset: 0,
      color: 'black',
    },
    {
      offset: 1,
      color: 'rgba(0,0,0,0)',
    },
  ]
  </script>
  
<T.Mesh rotation.x={-1 * 0.5 * Math.PI}>
  <T.PlaneGeometry />
  <T.MeshBasicMaterial depthWrite={false} transparent opacity={0.5} side={DoubleSide} />
</T.Mesh>
```

i think this is too simple to turn into its own component but let me know what you think.

michealparks:
I think it still makes sense to be its own component, even if it uses RadialGradientTexture under the hood! Especially if people are coming from drei, and that would make it more discoverable than a section in RadialGradientTexture. But either would be useful IMO.

DefinitelyMaybe:
i second the include it anyway. simple things made readily available is solid start for everyone. Everyone revisiting adds to the re-learning curve.

let-josh:
i made some changes to RadialGradientTexture and LinearGradientTexture so I think this should now be considered a breaking change. The breaking change being that color stops now only accept a color string instead of a ColorRepresentation for their `color` property.

In the previous code, before the color would get added to the gradient, it would get "Colorified" by `color.set(stop.color)`. Where `color` was a reused file-level Three.Color instance. Then it would get stringified through `color.getStyle()`. Doing this would cause the alpha information to be lost since `Three.Color` only holds onto the RGB values. In the updated code the color is sent directly into the gradient's `addColorStop` function which only accepts css-style color strings.

in other words, this no longer works

```svelte
<script>
  const stops = [
    {
      color: new Color(),
      offset: 0
    }
  ];
</script>
<RadialGradientTexture {stops} />
```

instead the color property must be a css color string

```diff
- color: new Color()
+ color: 'black'
```

doing this means that both gradient textures can now have transparency in their color stops.

---

so i'm not sure if this should be a major or a minor changeset now.

michealparks:
Probably a minor changeset? Seems like a small breaking change and TS will alert that the object signature has changed.

let-josh:
changeset added. i think it's ready to be merged. i changed a little bit of stuff in the derivations that was constant, also changed the example to use positioning rather than polygon offset - to be more inline with what's used in the [threejs article about shadows](https://threejs.org/manual/#en/shadows)

PR #1552, opened 2025-07-18
[extras] Snippet-based portals

michealparks:
### Overview

Replaces our current portal implementation with a snippet-based one, which I've found to be more reliable during complex scenarios.

This PR also removes the `object` prop and updates the docs to recommend users to use `attach` for simple portal needs.

### DIFF
--- a/.changeset/gorgeous-goats-complain.md
+++ b/.changeset/gorgeous-goats-complain.md
@@ -1,0 +1,6 @@
+---
+"@threlte/extras": patch
+"@threlte/studio": patch
+---
+
+Use snippet based Portal component


packages/extras/src/lib/components/portals/Portal.svelte
```diff
@@ -1,66 +1,32 @@
...
 </script>
-
-{#if target}
-  <T
-    is={proxy}
-    attach={false}
-  >
-    {@render children?.()}
-  </T>
-{/if}
```
DefinitelyMaybe, line 66: oooooh, I think I see the appeal of snippet based. it's less `<T>` objects around right?

PR #1571, opened 2025-08-11
[chore][docs] update common deps and remove types import from theatre/core

DefinitelyMaybe:
~~So this import is actually a commonjs module and I'm doing this so we can move towards updating vite~~

why go halfway to updating vite and different deps when you could go all the way 🤷‍♂️ ~~WIP~~

some notes from along the way:

- updated tsconfig (https://docs.astro.build/en/guides/upgrade-to/v5/#changed-typescript-configuration)
- removed svelte-preprocess import (https://github.com/sveltejs/svelte-preprocess?tab=readme-ov-file#when-to-use-it)
- `"three": "0.122.0"` dep in gltf package
- "@kitschpatrol/tweakpane-image-plugin" uses 11 deprecated subdependencies - threlte/studio (is there an different way we could do this?)
- `rimraf` can be replaced https://e18e.dev/guide/replacement-guides/rimraf.html. there might be others in a similar position https://e18e.dev/guide/replacements.html


DefinitelyMaybe:
I have touched a couple of things things are are slightly outside of scope of this PR.

- some markdown files changed because of the output being mangled somehow
- an updated HTML example scene file

The remaining question I have is whether I remove the patch changeset across all packages. They all use vite and I updated that dep so I felt I should but I'm open to suggestion there

michealparks (approved):
I'd probably leave out the changeset, Vite is only a devDependency and doesn't make it to what's shipped to NPM, so therefore there's no real change

PR #1574, opened 2025-08-21
[extras] Add wireframe component

michealparks:
### Overview

Adds a shader based wireframe component:

Screenshot 2025-08-21 at 17 28 30

I also added shadows to a few docs examples for fun.

### DIFF
--- a/packages/extras/src/lib/components/Wireframe/material.ts
+++ b/packages/extras/src/lib/components/Wireframe/material.ts
@@ -1,0 +1,146 @@
+// Based on https://github.com/mattdesl/webgl-wireframes
+
+import { type Uniform, type Material, DoubleSide } from 'three'
+
+export const WireframeMaterialShaders = {
+  vertex: `
+attribute vec3 barycentric;
+
+varying vec3 v_edges_Barycentric;
+varying vec3 v_edges_Position;
+
+void initWireframe() {
+  v_edges_Barycentric = barycentric;
+  v_edges_Position = position.xyz;
+}`,
+  fragment: `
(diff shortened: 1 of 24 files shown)


packages/extras/src/lib/components/Wireframe/material.ts
```diff
@@ -0,0 +1,146 @@
...
+}`
+}
+
+export const setWireframeOverride = (
+  material: Material,
+  uniforms: {
+    [key: string]: Uniform
+  }
+) => {
+  material.onBeforeCompile = (shader) => {
```
DefinitelyMaybe, line 109: nit pick / boarder question, tsl? .....hmmm actually, probs too early. wait for boarder adoption/use before defaulting to it.

DefinitelyMaybe:
I like the shadows, tho somehow I'm getting a haunted forest vibe 😅 
Screenshot 2025-08-22 100242

DefinitelyMaybe:
docs LGTM, it makes it clear early and several times that this is a shader.

packages/extras/src/lib/components/Wireframe/material.ts
michealparks: It's a good question. I've been playing around with tsl and the webgpu renderer a bit in my free time. I think that it would be a good idea in the near future to match what three is doing and add two exports:

```ts
import { T } from "@threlte/core/webgpu" // <- uses three/webgpu export instead
import { Wireframe } from "@threlte/extras/webgpu" // <- has been converted to TSL
```

PR #1577, opened 2025-08-26
[docs][extras] Type fixes

michealparks:
### Overview

This PR combs the docs for type issues. I ended up finding a few in core and extras, which I've fixed here. The remaining work is docs cleanup.

### DIFF
--- a/apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
+++ b/apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { Mesh } from 'three'
+  import { Mesh } from 'three'
   import { BakeShadows } from '@threlte/extras'
   import { T, useTask } from '@threlte/core'
 
@@ -9,12 +9,10 @@
 
   let { bake }: Props = $props()
 
-  let mesh = $state.raw<Mesh>()
+  const mesh = new Mesh()
 
   useTask((delta) => {
-    if (mesh) {
-      mesh.rotation.y += delta
-    }
+    mesh.rotation.y += delta
   })
 </script>
 
@@ -31,14 +29,14 @@
   castShadow
 />
 
-<T.Mesh
-  bind:ref={mesh}
+<T
+  is={mesh}
   castShadow
   position.y={1}
 >
   <T.BoxGeometry args={[1, 2, 1]} />
   <T.MeshStandardMaterial color="orangered" />
-</T.Mesh>
+</T>
 
 <T.Mesh
   receiveShadow
--- a/apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
+++ b/apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
@@ -9,11 +9,12 @@
 
   let { bake }: Props = $props()
 
-  let rotation = 0
-  let mesh = $state<Mesh>()
+  let mesh = $state.raw<Mesh>()
+
   useTask((delta) => {
-    rotation += delta
-    mesh?.rotation.set(0, rotation, 0)
+    if (mesh) {
+      mesh.rotation.y += delta
+    }
   })
 </script>
 
@@ -31,9 +32,7 @@
 />
 
 <T.Mesh
-  oncreate={(ref) => {
-    mesh = ref
-  }}
+  bind:ref={mesh}
   castShadow
   position.y={1}
 >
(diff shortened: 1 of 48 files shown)


apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
```diff
@@ -9,11 +9,12 @@
 
   let { bake }: Props = $props()
 
-  let rotation = 0
-  let mesh = $state<Mesh>()
+  let mesh = $state.raw<Mesh>()
```
let-josh: i think we should make the mesh in the script tag and use `<T is={mesh} />` that way we wouldn't have to check if the mesh exists in the task below - but those changes can all come in a later PR.

packages/extras/src/lib/components/BakeShadows/BakeShadows.svelte
```diff
@@ -7,13 +7,18 @@
...
-      renderer.shadowMap.needsUpdate = true
+  $effect.pre(() => {
+    if (suspended.current) {
+      return
     }
+
+    const { autoUpdate } = renderer.shadowMap
+
+    renderer.shadowMap.autoUpdate = false
+    renderer.shadowMap.needsUpdate = true
```
let-josh, line 18: this line is also in the effect's returned function but it's the same line. i wonder if it can be removed. this was just a copy and paste from drei and they do the same thing https://github.com/pmndrs/drei/blob/master/src/core/BakeShadows.tsx#L6 probably no harm to leave it the way it is.

joshwashywash (approved):

apps/docs/src/examples/extras/gizmo/Scene.svelte
```diff
@@ -51,13 +52,10 @@
...
-      attach={({ parent, ref }) => {
-        parent.setAttribute('color', ref)
-      }}
-    />
-  </T.BoxGeometry>
+  <T.BoxGeometry
+    oncreate={(ref) => {
+      ref.setAttribute('color', new BufferAttribute(colors, 3))
+    }}
+  />
```
DefinitelyMaybe, line 59: As a framework we've grown well 🙌 
why say many thing when few do trick 😄

apps/docs/src/examples/extras/gradient-texture/linear/Scene.svelte
```diff
@@ -47,7 +47,7 @@
     textureWrapT
   }: SceneProps = $props()
 
-  let stops: ColorStop[] = $derived([
+  let stops = $derived<ColorStop[]>([
```
DefinitelyMaybe, line 50: Hey can we have a quick talk about type declarations. I asked AI about why use the `$state<type>` and it talked about giving the compiler extra info, thought ok but checked in with svelte docs, [state types page](https://svelte.dev/docs/svelte/typescript#Typing-$state), umm didn't mention `$state<type>` way at all ... is this a preference thing?

packages/core/src/lib/components/T/T.ts
```diff
@@ -74,4 +66,8 @@ export const T = new Proxy(TComp, {
 
     return TComp
   }
-}) as typeof TComp & TComponentProxy & Extensions
+}) as typeof TComp & {
+  [Key in keyof typeof THREE]: Component<Props<(typeof THREE)[Key]>, {}, 'ref'>
+} & {
+  [Key in keyof Threlte.UserCatalogue]: Component<Props<Threlte.UserCatalogue[Key]>, {}, 'ref'>
+} & Extensions
```
DefinitelyMaybe: nice! this lets users add things to the ambient types file right?
```typescript
namespace Threlte {
    UserCatalogue {
        mytype: string
    }
}
```
or something right?

DefinitelyMaybe (approved):
cool cool 👍 

Clean-up isle docs, lets go 😃

apps/docs/src/examples/extras/gradient-texture/linear/Scene.svelte
michealparks: Not just a preference, this is more typesafe since you're passing the type to the type argument rather than asserting that this type is correct. With this method you get automatic `undefined` inference if the rune arguments are empty, etc.

packages/core/src/lib/components/T/T.ts
michealparks: This is trying to fix some core type issues, but I'm going to leave this for a later PR.

apps/docs/src/examples/extras/gradient-texture/linear/Scene.svelte
DefinitelyMaybe: that's handy. thank you 👍

packages/core/src/lib/components/T/T.ts
michealparks: Fixed here: https://github.com/threlte/threlte/pull/1580

PR #1533, opened 2025-06-24
Simplify `<T>` and improve performance

michealparks:
### Changelog

* Simplify the T proxy, removing an unnecessary apply call
* Simplify the `useAttach` hook control flow
* Simplify `useCamera`, `useDispose`, `useEvents`
* Replace `useCreateEvent` with a simple `$effect`
* Enable fine-grained reactivity for individual props within `useProps`
* Enable fine-grained reactivity for individual props within `useEvents`

### Benchmarking

I ran multiple lighthouse tests as a rough estimate for performance gain. The following before and after screens are pretty representative of the average.

Before:
Screenshot 2025-06-24 at 15 44 47

After:
Screenshot 2025-06-24 at 15 47 17

The benchmark looked like:
```
<script lang="ts">
  import Stats from 'three/examples/jsm/libs/stats.module'
  import { T, useTask } from '@threlte/core'

  const stats = new Stats({
    trackGPU: false,
    trackHz: false,
    trackCPT: false,
    logsPerSecond: 4,
    graphsPerSecond: 30,
    samplesLog: 40,
    samplesGraph: 10,
    precision: 2,
    horizontal: true,
    minimal: false,
    mode: 0
  })

  document.body.append(stats.dom)

  const count = 1000

  let rotationX = $state(0)
  let rotationY = $state(0)

  useTask((delta) => {
    stats.end()
    stats.begin()
    rotationX += delta
    rotationY += delta
  })

  const arr = new Array(count).fill(0).map((_, index) => index)
</script>

<T.DirectionalLight />

<T.PerspectiveCamera
  makeDefault
  position={[15, 15, 15]}
  oncreate={(ref) => ref.lookAt(0, 0, 0)}
/>

{#each arr as index (index)}
  <T.Mesh
    position={[index % 10, (index / 10) % 10, Math.random() - 0.5]}
    rotation.x={rotationX}
    rotation.y={rotationY}
  >
    <T.BoxGeometry args={[0.1, 0.1, 0.1]} />
    <T.MeshStandardMaterial color="red" />

    <T.Mesh
      rotation.x={rotationX}
      rotation.y={rotationY}
      scale={0.1}
      position={[1, 1, 1]}
    >
      <T.SphereGeometry />
      <T.MeshStandardMaterial color="green" />

      <T.Mesh
        rotation.x={rotationX}
        rotation.y={rotationY}
      >
        <T.TorusKnotGeometry />
        <T.MeshStandardMaterial color="blue" />
      </T.Mesh>
[...]

PR #1580, opened 2025-08-28
[core] T component type fixes

michealparks:
### Overview

Fixes a few type issues with the T component
* Adding `{...rest}` params to a T component was failing type checking due to a `MaybeInstance<Type>` mismatch in the `attach` prop and the `oncreate` prop.
Screenshot 2025-08-28 at 16 12 16

* User catalog extensions to T were unintentionally removed in https://github.com/threlte/threlte/pull/1533

### DIFF
--- a/packages/core/src/lib/components/T/utils/useAttach.svelte.ts
+++ b/packages/core/src/lib/components/T/utils/useAttach.svelte.ts
@@ -79,7 +79,7 @@ export const useAttach = <T extends MaybeInstance<any>>(
     // Custom attach function
     if (typeof attach === 'function') {
       const cleanup = attach({
-        ref: current as T,
+        ref: current as MaybeInstance<T>,
         parent: parent.current,
         parentObject3D: parentObject3D.current
       })


packages/core/src/lib/components/T/T.ts
```diff
@@ -74,4 +74,7 @@ export const T = new Proxy(TComp, {
 
     return TComp
   }
-}) as typeof TComp & TComponentProxy & Extensions
+}) as typeof TComp &
```
DefinitelyMaybe, line 77: iirc we were trying to move away from the `unknown` type tho mentioning this as a difference between the two.
`as unknown as typeof TComp &` was the [original](https://github.com/michealparks/threlte/blob/acf6602b06ece6fcb0d0d661d813c7031bdaee9d/packages/core/src/lib/components/T/T.ts#L72)
As for what this means, I have no idea, types be a black box to me sometimes 🤷‍♂️

michealparks: Yep we're not using unknown here anymore, we still have to use it with extensions though since that could be anything

michealparks: I was able to get rid of unknown here because we're directly proxying TComp, when before we were proxying an anonymous function

PR #1583, opened 2025-08-31
[fix/docs] adds color input to the ShadowMaterial example.

joshwashywash:
adds a color input to the ShadowMaterial example

fixes a bug where updating the color prop didn't fully clear the texture's canvas. now, when `stops` updates which causes a new gradient to be created, the canvas is cleared using `context.clearRect` before the gradient is filled in.

also removes the save/restore method calls since updates to the canvas are only ever in response to prop updates and happens internally. the user can't really interact with the texture's canvas (it's possible but they really shouldn't). saving the context is a lot of extra work when most of the canvas's context isn't accessed and all draws to the canvas are full wipes and redraws.

updates RadialGradientTexture docs to correctly show which types of colors strings are acceptable in the color stops

### DIFF
--- a/.changeset/blue-birds-tickle.md
+++ b/.changeset/blue-birds-tickle.md
@@ -1,5 +1,0 @@
----
-"@threlte/extras": patch
----
-
-fix RadialGradientTexture to clear its canvas whenever the `stops` prop is updated
--- a/.changeset/blue-birds-tickle.md
+++ b/.changeset/blue-birds-tickle.md
@@ -1,5 +1,5 @@
 ---
-"@threlte/extras": major
+"@threlte/extras": patch
 ---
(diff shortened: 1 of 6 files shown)


.changeset/blue-birds-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@threlte/extras": major
```
DefinitelyMaybe: imo a minor changeset 🤔

let-josh:
Oh oops you’re totally right. 😵‍💫I’ll fix that tomorrow.

let-josh:
changeset is updated. i went with a patch cuz it's more of a bug fix than a new feature.

PR #1587, opened 2025-09-01
[docs] improve / simplify language

michealparks:
### Overview

Simplifies a few introductory pages. The intent is to make language more concise and clear, and reduce ambiguity. Secondary intent is to try to reduce the word count while saying the same thing.


apps/docs/src/content/learn/basics/handling-events.mdx
```diff
@@ -1,32 +0,0 @@
----
```
DefinitelyMaybe, line 1: I'm not sure about removing this page. I'm worried its removal will lead to both more questions about events and a worse beginner experience. What was the thought process here?

michealparks: I found this page a bit hard to understand / parse, but you're right that a page on events should be part of the getting started docs. I'll instead see if I can modify it a bit to make it less confusing for myself (and hopefully others 😄 )

DefinitelyMaybe:
would it be alrite if I add the events page back in for now and then merge. I like these changes.

michealparks:
@DefinitelyMaybe go for it!

apps/docs/src/content/learn/basics/handling-events.mdx
```diff
@@ -4,15 +4,15 @@ category: Basics
 order: -700
 ---
 
-Events come in three different flavours in Threlte: The DOM events you know from
```
DefinitelyMaybe, line 7: Did a small pass on the wording in the events page 👍

DefinitelyMaybe (approved):
LGTM

PR #1588, opened 2025-09-03
[extras] Add bvh plugin

michealparks:
Adds a BVH raycast plugin compatible with mesh, batchedmesh, and points.

Docs page: https://threlte-git-fork-michealparks-bvh-threlte.vercel.app/docs/reference/extras/bvh

Screenshot 2025-09-03 at 17 46 09
Screenshot 2025-09-03 at 17 46 03


apps/docs/src/examples/extras/bvh/points/Scene.svelte
```diff
@@ -0,0 +1,125 @@
...
+    onpointerenter={() => {
+      visible = true
+    }}
+    onpointerleave={() => {
+      visible = false
+    }}
+    onpointermove={(event) => {
+      point = event.point.toArray()
+      if (event.index) {
+        points.geometry.getAttribute('color').setXYZ(event.index, 1, 0, 0)
```
mattmacf98, line 95: this is pretty much imperceivable since the tris are so small

apps/docs/src/examples/extras/bvh/points/Scene.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+    return results
+  })
+
+  useTask(() => {
+    if (!points) return
+
+    const attribute = points.geometry.getAttribute('color')
+
+    const indices = points.userData.indices as Set<number>
```
mattmacf98, line 44: is there a reason we don't just change the actual point cloud colors in the glb instead of doing it on load?

apps/docs/src/examples/extras/bvh/points/Scene.svelte
michealparks: This is just for the docs demo :) check out the mouse interaction here: http://localhost:4321/docs/reference/extras/bvh#points

apps/docs/src/examples/extras/bvh/points/Scene.svelte
michealparks: Laziness :P this is just for docs, so I didn't focus on optimization, and the focus here is just animation to show raycasting

packages/extras/src/lib/bvh/raycast.ts
```diff
@@ -0,0 +1,262 @@
...
+      const helper = objectToHelper.get(mesh)
+      helper?.removeFromParent()
+      objectToHelper.delete(mesh)
+    }
+  }
+}
+
+export const createPointsBVH = (points: Points<any>, opts: BVHOptions) => {
+  const geometry = new BufferGeometry()
+  const position = points.geometry.getAttribute('position')
```
mattmacf98: [nit] positions

michealparks: Updated!

packages/extras/src/lib/bvh/raycast.ts
```diff
@@ -0,0 +1,262 @@
...
+  return () => {
+    mesh.geometry.disposeBoundsTree()
+    points.raycast = Points.prototype.raycast
+
+    pointsToMesh.delete(points)
+
+    if (opts.helper) {
+      const helper = objectToHelper.get(points)
+      helper?.removeFromParent()
+      mesh.removeFromParent()
```
mattmacf98, line 142: shouldn't we be deleting this mesh regardless of if the helper is on or not? since it is a temp mesh to just hold tris for the points.

packages/extras/src/lib/bvh/raycast.ts
```diff
@@ -0,0 +1,262 @@
...
+  sphere.copy(this.geometry.boundingSphere!).applyMatrix4(this.matrixWorld)
+  sphere.radius += threshold
+
+  if (!raycaster.ray.intersectsSphere(sphere)) {
+    return
+  }
+
+  const mesh = pointsToMesh.get(this)
+
+  if (!mesh) {
```
mattmacf98: should we put this first (above the bounding sphere compute and intersect check), to fast fail and not do expensive geometry computations if the map doesn't contain the Points anyways?

packages/extras/src/lib/bvh/raycast.ts
michealparks: Good question, the mesh is only added to a parent if there's a helper present (for drawing the bvh) since we need to give it the same world transform. Otherwise there's no need to remove it since it was never added

packages/extras/src/lib/bvh/raycast.ts
michealparks: Good point, updated!

packages/extras/src/lib/bvh/raycast.ts
mattmacf98: Would this Mesh still exist in the Scene if the pointsBVH dismounts? or is that cleaned up somewhere else?

mattmacf98 (approved):

packages/extras/src/lib/bvh/raycast.ts
michealparks: nope, it's never added unless there's a helper, so this pretty much covers removal.

I was on the fence as to whether to call `mesh.geometry.dispose()`, since that's the norm for freeing GPU resources, but here I don't think it's a good idea since (for performance) the geometry reuses the points attributes and a dispose call here would likely cause performance issues.

PR #1590, opened 2025-09-04
[docs] Update tailwindcss

DefinitelyMaybe:


DefinitelyMaybe:
Did a more detailed check over and I've missed some things. back to draft status for now.

DefinitelyMaybe:
awesome! that didn't take as long as I thought it would.

apps/docs/.prettierignore
```diff
@@ -1,4 +1,7 @@
-README.md
+node_modules/**
+dist
+.astro
```
michealparks: Why not format `.astro` files?

apps/docs/astro.config.ts
```diff
@@ -6,38 +6,35 @@ import rehypeSlug from 'rehype-slug'
...
-        '$components/Example/Example.astro',
-        '$components/Tip/Tip.astro',
-        '$components/Card/Card.astro'
-      ]
+      imports: ['$components/Example/Example.astro', '$components/Tip/Tip.astro']
     }),
-    tailwind(),
     svelte(),
     mdx({
+      gfm: false,
```
michealparks, line 37: why not use github flavored markdown?

apps/docs/src/components/Button/Button.svelte
```diff
@@ -18,14 +28,18 @@
...
+  }
+
+  const extras = `${paddings[size]} ${textSizes[size]} ${bgColors[color]} ${_class}`
 </script>
 
 <a
   {href}
   class={`flex w-fit flex-row gap-3 rounded-md text-center text-white ${extras}`}
 >
-  <div class="empty:hidden"><slot name="icon-before" /></div>
```
michealparks, line 28: Were we not using these slots anywhere?

DefinitelyMaybe: nope. Think there's another slot thats not used in mobilenav somewhere but i decided I wouldn't add it to this

DefinitelyMaybe: wait no. maybe the landing page had one instance but removing it doesn't change the look

apps/docs/src/components/Menu/LeftSidebar/LeftSidebarCategory.svelte
```diff
@@ -36,16 +28,15 @@
...
-              isEqual(activeUrlPathName, `${category.urlPrefix}/${item.slug}`)
-                ? 'border-orange font-bold text-orange'
-                : 'border-white/20 hover:border-white'
-            ]}
-            href={`${category.urlPrefix}/${item.slug}`}
-          >
+        <li
+          class={[
+            'sidebar-list-item',
+            activeUrlPathName === `${category.urlPrefix}/${item.slug}`
```
michealparks: We don't need to account for starting & trailing slashes here?

apps/docs/astro.config.ts
DefinitelyMaybe: So gfm is markdown with css styling. I created a `markdown.css` file with this so that styling source would be clear. Also there was a rollup issue around links that this fixes.

apps/docs/src/components/Menu/LeftSidebar/LeftSidebarCategory.svelte
DefinitelyMaybe: I wondered about that but I didn't notice any behavior changes by removing it

apps/docs/.prettierignore
DefinitelyMaybe: oh oops, one moment.

apps/docs/astro.config.ts
michealparks: Got it!

apps/docs/src/components/Menu/LeftSidebar/LeftSidebarCategory.svelte
michealparks: I'd leave it in, it's probably safer if someone navigates from an external source that adds a slash somewhere

michealparks (approved):
Awesome, thanks for the house cleaning!

apps/docs/src/components/Example/Example.astro
```diff
@@ -24,7 +24,7 @@ const allModules = import.meta.glob('../../examples/**/*', {
 }) as Record<string, string>
 
 for (const path in allModules) {
-  if (!path.replace('../../examples/', '').startsWith(Astro.props.path)) {
+  if (!path.replace('../../examples/', '').startsWith(Astro.props.path + '/')) {
```
DefinitelyMaybe, line 27: may have sorted the double scrollbar bug on examples with this

PR #1596, opened 2025-09-10
add useAsciiEffect

joshwashywash:
a hook for using the `AsciiEffect` addon from three/addons

this should be preferred over `AsciiRenderer` as it is much more flexible and gives more control to the user.

i've removed the docs pertaining to AsciiRenderer and added an `@deprecated` comment to the AsciiRenderer component

the only thing i'm not sure of is if we should leave in the AsciiRenderer docs and add a deprecation notice there too.


packages/extras/src/lib/components/AsciiRenderer/AsciiRenderer.svelte
```diff
@@ -1,3 +1,7 @@
+<!--
+@component
+@deprecated use `useAsciiEffect` instead
```
michealparks: I'd add a "will remove by date" notice as well, Three.js does a year which I think is a good timeline

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  $effect(() => {
+    const last = asciiEffect.domElement
+    dom.appendChild(last)
+    return () => {
+      dom.removeChild(last)
+    }
+  })
+
+  return setContext(KEY, {
```
michealparks: If this hook creates a context then I think that the better name is `createAsciiEffect`, that's more standard provider language. Then the function below can be called `useAsciiEffect`, signaling context usage

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    effect.domElement.style.position = 'absolute'
+    effect.domElement.style.top = '0px'
+    effect.domElement.style.left = '0px'
+    effect.domElement.style.pointerEvents = 'none'
+
+    return (ref = effect)
+  })
+
+  const size$ = fromStore(size)
```
michealparks: `size$` is a bit confusing since it's not a svelte convention, I'd just go with
```ts
  const size = fromStore(sizeStore) // Renamed in useThrelte call
```

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const useAsciiEffect = (
+  getters: Partial<{
+    getRenderer: () => WebGLRenderer
+    getCharSet: () => string
+    getOptions: () => AsciiEffectOptions
+    getForegroundColor: () => string
+    getBackgroundColor: () => string
+  }> = {}
+) => {
```
michealparks: I'd remove the `get` prefix, it's adding verbosity with not much gained. It's sort of the problem that existed in C# when all interfaces began with `I` (like `IAsciiEffectOptions`). I also think that the word `get` in a setter introduces a little cognitive dissonance. IDE type hints + people knowing Svelte is enough, I'd recommend instead just using the names:
```suggestion
export const useAsciiEffect = (charset: () => string, options: () => AsciiEffectOptions) => {
```

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+  const { camera: threlteCamera, canvas, dom, renderer, scene: threlteScene, size } = useThrelte()
+
+  const {
+    getRenderer = () => renderer,
```
michealparks: Should we instead just get the renderer from `useThrelte` so we can remove this arg? This param [should only ever be a webgl renderer](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/effects/AsciiEffect.js#L13) and I can't imagine why someone would want to use two.

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+    getCharSet = () => '',
+    getOptions = () => ({}),
+    getForegroundColor = () => 'white',
+    getBackgroundColor = () => 'black'
```
michealparks: Should we instead instruct users in the docs to set the dom properties and set a reasonable default (black / white) so that we can simplify?

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
+import { useThrelte } from '@threlte/core'
+import { AsciiEffect } from 'three/examples/jsm/effects/AsciiEffect.js'
+import type { AsciiEffectOptions } from 'three/examples/jsm/effects/AsciiEffect.js'
+import type { Camera, Scene, WebGLRenderer } from 'three'
+import { fromStore } from 'svelte/store'
+import { getContext, setContext } from 'svelte'
+
+const KEY = {}
```
michealparks:
```suggestion
const KEY = Symbol('ascii-effect-context')
```

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+  })
+
+  return setContext(KEY, {
+    get current() {
+      return asciiEffect
+    },
+    render: ({ scene, camera }: Partial<{ scene: Scene; camera: Camera }> = {}) => {
+      ref.render(scene ?? threlteScene, camera ?? threlteCamera.current)
+    }
+  })
```
michealparks: With this current setup both effect.render (which calls renderer.render) and the default renderer.render will be called (2 passes when we want 1).

I'm wondering, should we handle rendering internally or just export `current` and in the docs instruct people to disable default rendering and call `effect.render` [using this method](https://threlte-git-fork-joshwashywash-useasciieffect-threlte.vercel.app/docs/learn/basics/render-modes#render-modes-and-custom-rendering)?

packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export const useAsciiEffectContext = () => {
```
michealparks: What's the purpose of including the hook in the library? I imagine most cases people will configure and use the above function and not need to access asciiEffect far below in the tree, and if they do they could create their own hook.

apps/docs/src/content/reference/extras/use-ascii-effect.mdx
```diff
@@ -0,0 +1,68 @@
+---
+order: 5.10
+category: '@threlte/extras'
+name: useAsciiEffect
+sourcePath: 'packages/extras/src/lib/hooks/useAsciiEffect.svelte.ts'
+type: 'hook'
+---
+
+A hook that can be used to give a scene an asciiEffect. It is essentially a wrapper around the `AsciiEffect` addon from three/addons that hooks into some of threlte's context for ease of use.
```
DefinitelyMaybe:
```suggestion
A hook that can be used to give a scene an asciiEffect. It is a thin wrapper around the `AsciiEffect` [addon](https://threejs.org/examples/?q=as#webgl_effects_ascii).
```

apps/docs/src/content/reference/extras/use-ascii-effect.mdx
```diff
@@ -0,0 +1,68 @@
...
+```ts
+const asciiEffect = useAsciiEffect()
+
+const camera = new PespectiveCamera()
+const scene = new Sceneiee()
+
+asciiEffect.current.render(scene, camera)
+```
+
+If you want to render with threlte's main scene and main camera, you can pull those from the `useThrelte` hoox
```
DefinitelyMaybe:
```suggestion

Start by disabling the [autorendering](/docs/reference/core/canvas#component-signature) on your canvas i.e. `<Canvas autoRender={false}>`, then call this hooks render function within your `useTask`.
```

Start with the useThrelte hook here, im sure folks could figure out their custom scenes and cameras from there 😅

DefinitelyMaybe (commented):
So it's definitely more focused 👍 

I think in terms of customizing renderers, I could see making myself a custom component and then within it orchestrating one or two hooks. I don't see favoring one over the other tho. hooks and components have their place in this project.

PR #1597, opened 2025-09-14
Fix circular reference with TransformControls.

m3ftwz:
# Threlte v7 + Three.js r180 Circular Reference Fix

## Issue Description
After updating Three.js to version 0.180.0 (actually earlier versions results in the same bug as well), applications using Threlte v7 with TransformControls experience a "Maximum call stack size exceeded" error during object disposal.

### Error Details
```
RangeError: Maximum call stack size exceeded
    at TransformControls.get [as object] (http://localhost:3000/modules.a2902d1be5d16b3ccc2a.bundle.js:180718:19)
    at Object.entries (<anonymous>)
    at Object.collectDisposableObjects (http://localhost:3000/modules.a2902d1be5d16b3ccc2a.bundle.js:69393:20)
```

### Environment
- **Three.js**: 0.180.0
- **Threlte Core**: 7.3.1
- **Threlte Extras**: 8.12.2
- **Svelte**: 4.2.19

## Proposed Solution

The fix adds circular reference detection using a **WeakSet for Visited Objects**: it tracks already-processed objects to prevent infinite recursion

### DIFF
--- a/packages/core/src/lib/lib/contexts.ts
+++ b/packages/core/src/lib/lib/contexts.ts
@@ -104,7 +104,8 @@ export type ThrelteInternalContext = {
    */
   collectDisposableObjects: (
     object?: DisposableThreeObject,
-    arr?: DisposableThreeObject[]
+    arr?: DisposableThreeObject[],
+    visited?: WeakSet<any>
   ) => DisposableThreeObject[]
 
   /**
@@ -166,9 +167,12 @@ export const createThrelteContext = (options: {
       })
       internalCtx.shouldDispose = false
     },
-    collectDisposableObjects: (object, objects) => {
+    collectDisposableObjects: (object, objects, visited) => {
       const disposables: DisposableThreeObject[] = objects ?? []
+      const visitedObjects = visited ?? new WeakSet()
       if (!object) return disposables
+      if (visitedObjects.has(object)) return disposables
+      visitedObjects.add(object)
       // Scenes can't be disposed
       if (object?.dispose && typeof object.dispose === 'function' && object.type !== 'Scene') {
         disposables.push(object)
@@ -179,7 +183,7 @@ export const createThrelteContext = (options: {
         if (propKey === 'parent' || propKey === 'children' || typeof propValue !== 'object') return
         const value = propValue as any
         if (value?.dispose) {
-          internalCtx.collectDisposableObjects(value, disposables)
+          internalCtx.collectDisposableObjects(value, disposables, visitedObjects)
         }
       })
       return disposables


m3ftwz:
It would be great to have a new v7 version published if this gets merged since it is causing major crashes on latest three versions :)

m3ftwz:
After some testing, I found that while this solution resolves the maximum stack issue, it causes the Transform Controls to stop displaying. I’ll keep looking for a better fix, but any help would be greatly appreciated.

DefinitelyMaybe:
I take it upgrading to svelte 5 isn't an option?

cmdemetrio:
> I take it upgrading to svelte 5 isn't an option?

Unfortunately not. The codebase I current work with is really big and includes different services so it's not something that is going to happen anytime soon.

m3ftwz:
> I take it upgrading to svelte 5 isn't an option?

Unfortunately no. The codebase I currently work with is very big and includes different services so the svelte upgrade will not happen anytime soon.

grischaerbe:
@m3ftwz Fixed in @threlte/core@7.3.2, please feel free to report if this resolves your issues :)

m3ftwz:
@grischaerbe Thank you, the issue is solved, just one small problem left as much as I can see: the TransformControls visible property is not respected anymore and the controls are always showing.

This is the patch I'm using locally if it helps (I didn't think much about it so it might be not suitable for prod)

```diff
diff --git a/node_modules/@threlte/extras/dist/components/controls/TransformControls/TransformControls.svelte b/node_modules/@threlte/extras/dist/components/controls/TransformControls/TransformControls.svelte
index 19670fa..9fbdc46 100644
--- a/node_modules/@threlte/extras/dist/components/controls/TransformControls/TransformControls.svelte
+++ b/node_modules/@threlte/extras/dist/components/controls/TransformControls/TransformControls.svelte
@@ -47,20 +47,26 @@ const controlsStore = derived(camera, (camera) => {
 export let controls = $controlsStore;
 $: controls = $controlsStore;
 const attachTo = writable(object ?? group);
+let currentHelper = undefined;
 watch([controlsStore, attachTo], ([controls, attachTo]) => {
     controls.attach(attachTo);
     if ('getHelper' in controls) {
-        const helper = controls.getHelper();
-        scene.add(helper);
+        currentHelper = controls.getHelper();
+        scene.add(currentHelper);
     }
     return () => {
         controls.detach();
         if ('getHelper' in controls) {
             const helper = controls.getHelper();
             scene.remove(helper);
+            currentHelper = undefined;
         }
     };
 });
+// Watch for visibility changes and sync to helper
+$: if (currentHelper && $$restProps.visible !== undefined) {
+    currentHelper.visible = $$restProps.visible;
+}
 // This component is receiving the props for the controls as well as the props
 // for the group, so we need to split them up
 const transformOnlyPropNames = [
```

grischaerbe:
Good point, I'll make a follow up PR from that 👍

m3ftwz:
@grischaerbe any update? do you prefer if I open an additional PR?

DefinitelyMaybe:
> any update? do you prefer if I open an additional PR?

Imo go ahead and make the PR

Issue #1598, opened 2025-09-15
Cleanup issues when using svelte imperative api


GuillaumeQuenneville:
Hi everyone,

I came accros this issue when making a python notebook widget. When calling the svelte api `unmount` on a threlte component that has a buffer geometry, threlte tries to dispose of an undefined object and throws an error. This would occur in the python notebook context when you rerun a cell.

When unmounting the threlte component it throws the error

```javascript
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'isInterleavedBufferAttribute')
    at Object.remove (three.module.js:13650:18)
    at _BufferGeometry.onGeometryDispose (three.module.js:17416:15)
    at _BufferGeometry.dispatchEvent (three.module.js:294:16)
    at _BufferGeometry.dispose (three.module.js:11539:8)
    at disposal.js:33:29
    at Map.forEach (<anonymous>)
    at Object.dispose (disposal.js:31:39)
```

I have made a minimal reproduction [here](https://github.com/GuillaumeQuenneville/threltebug). just pnpm i && pnpm dev like usual.

I haven't been able to reproduce an issue specific to the python notebooks, but there seems to be leftover listeners based on the number of times i've unmounted. In my minimal reproduction you only get the error showing up once, in the notebooks it runs "n" times based on the times i've rerun the cell. I'm still trying to figure out what's causing that.

DefinitelyMaybe:
Hey! interesting find.

I see that @threlte/core will install three@0.166.1 if none is specified.

Could you please confirm that this errors still occurs in recent versions of three.js. threejs versions can have changes that don't work well vs each other and perhaps we need to update our minimum version

DefinitelyMaybe:
Other thing of note, it's asking for [isInterleavedBufferAttribute](https://threejs.org/docs/?q=buff#api/en/core/InterleavedBuffer) which could mean that three.js has changed the object from what threlte expected, or perhaps three.js required a different setup?

GuillaumeQuenneville:
Thank you for the feedback. I can confirm that the issue occurs with     "three": "^0.180.0", . I've fixed the version to that in my minimal reproduction.

PR #1600, opened 2025-09-21
sanitize BufferGeometry attributes before dispose

GuillaumeQuenneville:
Hi everyone,

This fixes https://github.com/threlte/threlte/issues/1598

When attaching attributes to a BufferGeometry (for example "attributes.position"), Threlte’s cleanup restores the valueBeforeAttach only at the leaf of the property chain. This leaves geometry.attributes.position = undefined.

Later, Three.js’ WebGLGeometries cleanup does:

```js
for (const name in geometry.attributes) {
attributes.remove(geometry.attributes\[name]);
}
```

So it ends up calling attributes.remove(undefined), which throws the error reported in the issue.

This PR applies a minimal defensive fix in Threlte’s disposal cycle. If a BufferGeometry has stray undefined entries in geometry.attributes, they are deleted before calling dispose(). This prevents the crash while remaining non-breaking.

A more robust fix would be to adjust useAttach cleanup: if the attach path is "attributes.position", the cleanup should call geometry.deleteAttribute("position") instead of restoring undefined. That would avoid the orphaned attributes in the first place. I’m happy to follow up or adjust if maintainers prefer that direction or some other logic.

I’ve updated the minimal reproduction repo to confirm the fix.

Relevant code sections

Threlte attach
https://github.com/threlte/threlte/blob/6dd18daa938ce286cebaee6c54c751fb6276df09/packages/core/src/lib/components/T/utils/useAttach.svelte.ts#L92-L101
    
 Threejs geometry cleanup https://github.com/mrdoob/three.js/blob/1abd71ef8a2c612be83500450ed246b18395a3ec/src/renderers/webgl/WebGLGeometries.js#L9-L25    

I hit this while using Threlte in a Python notebook widget, where disposal runs every time you re-execute a cell. The crash during cleanup caused strange behavior (canvas not attaching, blank output). With this patch, the cleanup runs correctly and rendering resumes as expected.


GuillaumeQuenneville:
I was wondering if we leak GPU memory by not passing through deleteAttribute but it look like all it does is call delete on the attribute so we should be fine with regards to that.

https://github.com/mrdoob/three.js/blob/c9addf30b55ae91747738685998baa9ad9701638/src/core/BufferGeometry.js#L282-L288

grischaerbe:
Thanks @GuillaumeQuenneville for bringing that to our attention and sorry it's taken so long. I needed to wrap my head around it at first since it only happens in very particular situations. I first rebased your branch and then (as you already suggested) adjusted the attach logic to only restore attribute values if the attribute was present before attachment. This should be pretty robust for most cases. In any other situation there's still the option to attach manually with a pair of functions to attach/detach.

PR #1602, opened 2025-09-26
Update stackblitz main.ts

Mystler:
When using the StackBlitz button in the docs to create a playground for testing or sharing reproducable demos, the main entry still uses an old way of creating the Svelte app that I always had to fix manually, so far. This should update it baseline so that manual fixing is not necessary.


DefinitelyMaybe (approved):

DefinitelyMaybe:
good stuff 👍 LGTM

PR #1603, opened 2025-09-28
[rfc] InstancedMesh2 prototype

jerzakm:


DefinitelyMaybe:
throwing a name into the ring: ezInstancedMesh 🧑‍🍳

Mystler:
This seems to be working great. It also fixes the problem that i've been having with many instances taking unusably long to load since Chromium-based browser version 140 that i've been struggling to find a root cause for.

DefinitelyMaybe:
ok, can add some information to the interactivity side of this because i dug around a little after finding a `// todo review` comment ❤️

right first off is the flame graphs for this before any events:
meshbefore
and then during:
meshduring

so going from a 1ms total to ~3ms

It's not an apples to apples comparison but plain interactivity is the following:
before:
Screenshot 2025-10-10 104911
during:
Screenshot 2025-10-10 104724
Screenshot 2025-10-10 104757

going from a 0.7ms total to ~1.2-1.6ms 

So kinda not far off current interactivity overhead multiplier (~2x)

Of course it's actually smooth as butter if you take out interactivity.

I'm leaning towards if someone is hitting this interactivity issue that they should rework their structure first i.e. chunk their interactive Instance's into multi InstanceMeshes

PR #1605, opened 2025-10-01
[docs] Example cleanup + misc

DefinitelyMaybe:
started from noticing the `import { createEventDispatcher } from 'svelte'` in the Example component, turned into using the context API when I noticed the propdrilling. Furthermore the double scroll bar issue in examples is no more 🎉 

Also added some css variable usage instead of hardcoded numbers and cleaned up some slots for clarity

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@threlte/docs",
   "type": "module",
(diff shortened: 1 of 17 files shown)


DefinitelyMaybe:
Still the double sidebar scroll issue. will have a look around for a fix.

PR #1606, opened 2025-10-02
update screen-quad example improvements

joshwashywash:
- updates some of the example mdx content with more correct information
- simplifies some of the shader code and component code, notably using `useFBO` which does everything we want it to do instead of creating a RenderTarget and doing it all ourselves.

### DIFF
--- a/apps/docs/src/content/examples/postprocessing/ScreenQuad.mdx
+++ b/apps/docs/src/content/examples/postprocessing/ScreenQuad.mdx
@@ -4,25 +4,25 @@ This example demonstrates a well-known technique for doing simple postprocessing
 
 ## Overview
 
-The basic idea of postprocessing is to draw the scene to a frame-buffer that can then be sent into another shader. The output of this shader is used as the texture or material for a mesh that covers the screen. To do this we need two things:
+The basic idea of postprocessing is to draw the scene to a frame-buffer ortexture that can then be sent into another shader. The output of this shader is used as the texture or material for a mesh that covers the screen. To do this we need two things:
 
 1. a WebGL render target.
 2. a separate scene from the one that Threlte provides.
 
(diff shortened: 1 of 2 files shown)


apps/docs/src/content/examples/postprocessing/ScreenQuad.mdx
```diff
@@ -4,36 +4,66 @@ This example demonstrates a well-known technique for doing simple postprocessing
 
 ## Overview
 
-The basic idea of postprocessing is to draw the scene to a frame-buffer that can then be sent into another shader. The output of this shader is used as the texture or material for a mesh that covers the screen. To do this we need two things:
+The basic idea of postprocessing is to draw the scene to a frame-buffer ortexture that can then be sent into another shader. The output of this shader is used as the texture or material for a mesh that covers the screen. To do this we need two things:
```
DefinitelyMaybe:
```suggestion
The basic idea of postprocessing is to draw the scene to a frame-buffer or texture that can then be sent into another shader. The output of this shader is used as the texture or material for a mesh that covers the screen. To do this we need two things:
```

apps/docs/src/content/examples/postprocessing/ScreenQuad.mdx
```diff
@@ -4,36 +4,66 @@ This example demonstrates a well-known technique for doing simple postprocessing
...
+
+const uSize = new Uniform(new Vector2())
+
+const { size } = useThrelte()
+
+const sizeAsState = fromStore(size)
+
+$effect(() => {
+  uSize.value.set(sizeAsState.current.width, sizeAsState.current.height)
+})
```
DefinitelyMaybe, line 60: The `size` variable from useThrelte is a [currentReadable](https://github.com/threlte/threlte/blob/780b3c518fd6aac000ea79323cf4e01041a53095/packages/core/src/lib/context/fragments/dom.ts#L8C2-L8C59) so could use as is

DefinitelyMaybe (commented):
nice, when preview deploys come back online I'll do a visual check but looking good so far 👍

let-josh:
closing in favor of #1624

PR #1607, opened 2025-10-02
[docs] Add SVG render order example

DefinitelyMaybe:
So I was stupid and tried to remove the `renderOrder` variable within the SVG component 😅 

The variable is needed and is not just a old quirk lying around in the three.js [example code](https://github.com/mrdoob/three.js/blob/0af9729d0c143a86a1d725d6e2c3ad83301f3f34/examples/webgl_loader_svg.html#L170).

To resolve the warning I've added a svelte ignore comment `// svelte-ignore non_reactive_update`

We could add an effect like so:

```ts
$effect.pre(() => {
    if (src) {
      renderOrder = 0
    }
  })
```

for completeness's sake if someone updates `src`  reactively a bunch of times .... but even then the variable will just keep going up so not worth worrying about imo


DefinitelyMaybe:
umm we could but I don't think we do that in too many places (like including pngs n stuff). happy to change if you'd prefer

grischaerbe:
In this case it might make sense because the renderOrder is directly correlated with the order of the SVG content, right?

DefinitelyMaybe:
yeah, all good. I'll move it into the example directory

PR #1632, opened 2025-11-02
[docs] migrate to starlight

DefinitelyMaybe:
Most important things the change gives us:

- correct all internal links. invalid internal links will error and fail future builds.
- Search is built-in and created during the build process
  - removes our preact deps
  - the `data-pagefind-ignore` html attribute has been added to example components so that their content isn't indexed

Noteable improvements:
- Mobile navigation improvements

### DIFF
--- a/apps/docs/src/examples/core/three-arcade-game/game/scenes/Intro.svelte
+++ b/apps/docs/src/examples/core/three-arcade-game/game/scenes/Intro.svelte
@@ -42,7 +42,7 @@
     showPressSpaceToStart = true
   }, showPressSpaceToStartAfter)
 
-  let intervalHandler = window.setInterval(() => {
+  let intervalHandler = setInterval(() => {
     if (!showPressSpaceToStart) return
     blinkClock = blinkClock ? 0 : 1
   }, 500)
--- a/apps/docs/src/examples/core/three-arcade-game/Scene.svelte
+++ b/apps/docs/src/examples/core/three-arcade-game/Scene.svelte
@@ -7,7 +7,7 @@
   import CustomRendering from './Renderer.svelte'
 
   $effect(() => {
-    const intervalHandler = window.setInterval(() => {
+    const intervalHandler = setInterval(() => {
       game.blinkClock = game.blinkClock === 0 ? 1 : 0
     }, 96)
     return () => clearInterval(intervalHandler)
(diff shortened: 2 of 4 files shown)


apps/docs/src/examples/core/three-arcade-game/game/scenes/Intro.svelte
```diff
@@ -42,7 +42,7 @@
     showPressSpaceToStart = true
   }, showPressSpaceToStartAfter)
 
-  let intervalHandler = setInterval(() => {
+  let intervalHandler = window.setInterval(() => {
```
DefinitelyMaybe: So these changes happened because the `setInterval` call happened to create an ci/cd check error because it thought the return type wasn't just a number 🤷‍♂️ I guess somewhere i didn't wrangle the types correctly. this is an fix till then.

apps/docs/src/layouts/MainLayout.astro
```diff
@@ -65,19 +64,6 @@ const isBlog = Astro.url.pathname.startsWith('/blog')
     </div>
   </header>
 
-  <!-- Mobile Nav -->
```
DefinitelyMaybe, line 68: this was an erronous removal. I'll add it back in soon.

DefinitelyMaybe:
on second thought. lemme try reduce it further by tackling these two items.
> - added a title yaml to example pages
> - changed current yaml name to title

DefinitelyMaybe:
hmm on further thought.

* adjusted href links -> for some piece of mind I could add the link validation now and create redirects for old links
* The installation page's install code doesn't have a copy code button atm -> needs to be sorted or left as it was

also there are some other trivial changes which could be done at a later date i.e. file renames, yaml changes.

apps/docs/src/content/docs/learn/getting-started/introduction.mdx
```diff
@@ -0,0 +1,35 @@
+---
```
DefinitelyMaybe, line 1: for some reason the git diff didn't connect that this file was moved. the only changes here are to the yaml and urls

apps/docs/src/examples/core/three-arcade-game/game/scenes/Intro.svelte
DefinitelyMaybe: reverted these changes.

Issue #1471, opened 2025-04-21
Using `"moduleResolution": "NodeNext"` in `tsconfig.json` creates type errors when using Threlte


dslatkin:
### Summary

Basically if you have `"moduleResolution": "NodeNext"` in your `tsconfig.json`, then you get a type error importing from Threlte. In JS projects using `jsconfig.json`, you won't get a type error, but Intellisense doesn't work so you can't see type hints, do function lookups, etc. Using `"moduleResolution": "bundler"` makes the error go away, but shouldn't be needed anymore as-of #985.

Using `npx sv create` with library as the project type uses `NodeNext` module resolution by default for increased compatibility but then doesn't work for people using Threlte in their documentation site at `src/routes`, for example.

### Possible fix

With the repro below, editing `node_modules/@threlte/core/dist/types.d.ts` and changing `export * from './index'` to `export * from './index.js'` makes the issue go away. I _think_ if the Threlte packages use `NodeNext` module resolution instead and update internal imports to use use explicit `.js` file paths (as required/standard for `.ts` imports with this setting), this would fix the issue and make Threlte more compatible across the ecosystem?

### Reproduction

Easiest way is to clone the repo https://github.com/dslatkin/threlte-issue and run `npm run check` to see the issue.

To recreate for a new project:

1. `npx sv create`
   - "Svelte library" for template
   - "Use TypeScript syntax" for type checking
2. `npm install three @types/three @threlte/core @threlte/extras`
   - Not sure if you need all these packages, but just in case...
3. Replace `src/routes/+page.svelte` with the following and `npm run check`

```html
<script lang="ts">
    import { Canvas } from '@threlte/core';
</script>

<Canvas>
    Inside canvas
</Canvas>
```

```
Error: Module '"@threlte/core"' has no exported member 'Canvas'. (ts)
<script lang="ts">
    import { Canvas } from '@threlte/core';
</script>


====================================
svelte-check found 1 error and 0 warnings in 1 file
```

joshwashywash:
https://discord.com/channels/985983540804091964/1363895441573413066 - discord discussion for reference

DefinitelyMaybe:
Hey @benmccann , if you've got some spare time, could we please have your thoughts on this issue.

benmccann:
If you create a new library with `npx sv create` you will get the following `tsconfig.json`:
```
{
	"extends": "./.svelte-kit/tsconfig.json",
	"compilerOptions": {
		"allowJs": true,
[...]

PR #1634, opened 2025-11-03
Update `tsconfig.json` of all Svelte libraries to match recommended defaults: `"moduleResolution": "NodeNext"`

helgee:
Fixes: #1471

### DIFF
--- a/packages/extras/CHANGELOG.md
+++ b/packages/extras/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @threlte/extras
 
+## 9.6.1
+
+### Patch Changes
+
+- 642343b: [core] Fix: adjust type export
+
 ## 9.6.0
 
 ### Minor Changes
--- a/packages/extras/package.json
+++ b/packages/extras/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@threlte/extras",
-  "version": "9.6.0",
+  "version": "9.6.1",
   "author": "Grischa Erbe <hello@legrisch.com> (https://legrisch.com)",
   "license": "MIT",
   "description": "Utilities, abstractions and plugins for your Threlte apps",


helgee:
`camera-controls` is currently incompatible and the usual workarounds don't help. I will submit a fix upstream.

EDIT: See here https://github.com/yomotsu/camera-controls/pull/612

DefinitelyMaybe:
Great stuff. I've subscribed to the PR you made there. I'd consider removing camera-controls temporarily as an intermediary solution to get this merged but will bring it to the teams attention first in case someone has a better idea.

DefinitelyMaybe:
@helgee I got the tests passing using some casting to `any` to literally just ignore the error in the mean time. Little bit less drastic solution than removing camera-controls all together. And once the upstream issue is resolved we can come back to this.

helgee:
Oh yeah, that does work 😅 Thanks!

Using pnpm's package patching functionality would also be an option.

helgee:
I remember now why I tried to avoid casting to `any` in this case. The `any` will propagate and also make the props of the components that use `CameraControls` untyped.

helgee:
I think I will go ahead and use `pnpm patch` and then you folks can decide if this is the way or not. Upgrading to a new `camera-controls` version might also be undesirable due to this: https://github.com/yomotsu/camera-controls/issues/609

helgee:
Finally, I am not sure what kind of version bump this warrants. Let me know and I will add a changeset.

DefinitelyMaybe:
> I remember now why I tried to avoid casting to `any` in this case. The `any` will propagate and also make the props of the components that use `CameraControls` untyped.

Ah, my bad.

TIL about `pnpm patch`, damnnnnnnn 👍

DefinitelyMaybe:
oh! the changeset. one moment.

michealparks:
Reverted this because there are a few areas of uncertainty that need to be checked first. For example, should `svelte.ts` files also have extensions added? If so this PR doesn't yet include them:

```ts
// should this be './utils/useCamera.svelte.js' ?
import { useCamera } from './utils/useCamera.svelte'
```

If I get a moment this week I'll look into this more, since this solution seems to be the way to go!

DefinitelyMaybe:
My bad @helgee ! Little too trigger happy on my side.

helgee:
No worries. Happy to see progress 👍

helgee:
> For example, should `svelte.ts` files also have extensions added?

Yes, IMHO. Because otherwise it is not clear whether the imported file is a component or not.

helgee:
New version in #1648.

PR #1641, opened 2025-11-07
Fix isInstanceOf for Light

zadinvit:
The Light type cannot be used in the isInstanceOf() function at the moment. The problem was with the ThreeClassTypes type because it only checks for class constructors, not abstract class constructors (THREE.Light is an abstract class). I just added the abstract class type as well. 

This should work now: 
```
function is_light(object: THREE.Object3D){
	if (isInstanceOf(object, 'Light')) {
		// Do something specific for lights
	}
}
```

### DIFF
--- a/.changeset/yellow-doors-switch.md
+++ b/.changeset/yellow-doors-switch.md
@@ -1,0 +1,5 @@
+---
+"@threlte/core": minor
+---
+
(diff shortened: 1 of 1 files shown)


michealparks (approved):
Thanks for catching this!

PR #1657, opened 2025-11-23
Update camera-controls and remove patch

helgee:
The latest `camera-controls` version includes my patch and thus it is not needed anymore.


michealparks (approved):
Thank you for following up!

PR #1662, opened 2025-11-25
Patch `@theatre/studio` and remove workaround

### LINKED ISSUE #1659: Regression: crash when importing `@threlte/theatre` + broken demo due to import changes in PR #1648
As of `@threlte/theatre` 3.1.0, importing `@threlte/theatre` crashes the app. The [demo for `<SoftShadows>`](https://threlte.xyz/docs/reference/extras/soft-shadows) is broken [(link to StackBlitz project)](https://stackblitz.com/edit/gn2xz5ew?file=src%2Fexample%2FApp.svelte).

In the demo project, the following error is printed to the console:
```
InnerStudio-R666PK4S.js?v=dc8aa5e3:41 Uncaught (in promise) TypeError: Studio.initialize is not a function
```

Digging into this error, I can see that in `@threlte/theatre`, `InnerStudio.svelte` dynamically imports `@theatre/studio`, and then assumes the result of the dynamic import is the default export itself, but this assumption does not hold in either the demo project nor a freshly created Svelte project using `npx sv create` (version 0.10.4).

Making this local patch in `node_modules/@threlte/theatre/dist/studio/InnerStudio.svelte` (using eg `bun patch`) fixes the problem for me:

```diff
   const studioModule = await import('@theatre/studio')
-  const Studio = studioModule as unknown as typeof import('@theatre/studio').default
+  const Studio = studioModule.default
   import { studio } from '../consts.js'
 
   Studio.initialize()
```

It seems this regression was caused by https://github.com/threlte/threlte/pull/1648, in particular this commit https://github.com/threlte/threlte/pull/1648/commits/4856710019119c704a23227db3c6313be44ba5af.

Perhaps @helgee could confirm whether the assumption _should_ hold (and we're looking at possibly tsconfig/bundler misconfiguration), or if the assumption should not hold, and the patch above points towards the ideal fix.

helgee:
Fixes #1659

### DIFF
--- a/.changeset/thick-mammals-fix.md
+++ b/.changeset/thick-mammals-fix.md
@@ -1,0 +1,5 @@
+---
+'@threlte/theatre': patch
+---
+
+patch @theatre/studio and remove workarounds
--- a/package.json
+++ b/package.json
@@ -31,6 +31,9 @@
   "pnpm": {
     "overrides": {
       "vite": "catalog:"
+    },
+    "patchedDependencies": {
+      "@theatre/studio": "patches/@theatre__studio.patch"
     }
   },
   "engines": {
--- a/packages/theatre/src/lib/studio/InnerStudio.svelte
+++ b/packages/theatre/src/lib/studio/InnerStudio.svelte
@@ -2,8 +2,7 @@
   lang="ts"
   module
 >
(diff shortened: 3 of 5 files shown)


DefinitelyMaybe (approved):
Thank you!

PR #1668, opened 2025-12-15
[docs] simplify webgpu slice material example tsl code

let-josh:
nodes can be "floating" (not sure if that's that's the best word) but essentially they don't need to be *Fn*ed, they can just be out and about like regular variables.

for such a simple example, i think the tsl.ts file is just extra noise so i removed that and just inlined most of the code and simplified the shader nodes.


DefinitelyMaybe (approved):

DefinitelyMaybe:
Nice 👍 LGTM

PR #1670, opened 2025-12-15
remove "updating uniforms" content

### LINKED ISSUE #1669: [docs] remove "webgpu updating uniforms" example and content
it's not a very good example since you normally wouldn't update emissive properties the way that it's showing. instead you'd simply update the emissive intensity like this:

```svelte
<script>
  let emissiveIntensity = $state(0);
</script>

<T.MeshStandardMaterial
  {emissiveIntensity}
>
</T.MeshStandardMaterial>
```

It's making something into TSL that doesn't really need to be TSL and making a uniform out of something that is already (probably) a uniform under the hood.

Aside from that, updating uniforms is no different than how you would update them normally with a WebGLRenderer. if you had a uniform node, you'd just update its **.value** in a callback or effect somewhere and the change will get picked up on the next render.

```svelte
<script>
  const intensity = uniform(0)
</script>

<button onclick={() => {
  intensity += 1;
}}>
```

i think this issue should come after #1668 gets merged.

let-josh:
closes #1669

### DIFF
--- a/apps/docs/src/examples/shaders/webgpu/slice/App.svelte
+++ b/apps/docs/src/examples/shaders/webgpu/slice/App.svelte
@@ -2,7 +2,7 @@
   import Scene from './Scene.svelte'
   import { Canvas, extend } from '@threlte/core'
   import { Checkbox, Color, Folder, Pane, Slider } from 'svelte-tweakpane-ui'
-  import { MathUtils, Color as ThreeColor } from 'three'
+  import { MathUtils } from 'three'
   import {
     DirectionalLight,
     MeshPhysicalNodeMaterial,
@@ -62,7 +62,7 @@
   <Scene
     {rotate}
     {arcAngle}
-    sliceColor={new ThreeColor(sliceColor)}
+    {sliceColor}
     {startAngle}
   />
 </Canvas>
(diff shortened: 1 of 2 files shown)


michealparks:
Weird that no docs environment triggered here, @grischaerbe did something happen with our deployment setup?

DefinitelyMaybe:
looks like a ton are queued and one is hanging in-progress. Their are some server updates that might need to happen to clear this but I broke things when trying it earlier.

DefinitelyMaybe:
So with regards to _italics_ and **bold** in docs, my opinion is simply to follow astro's style. highlighting `code` like things (components, props, config variable names) with back ticks helps distinguish them for the reader.

There is a balance somewhere between visually distinguished things and a chaotic mess.... and some pages may be more chaotic than others atm.

apps/docs/src/content/learn/advanced/webgpu.mdx
```diff
@@ -246,44 +273,38 @@ This is because the exports from `three/webgpu` are different than those in `thr
...
-The `outputNode` is set up in such a way that it discards any fragments that are
-outside the permitted `startAngle` and `arcAngle`. If a fragment is not discarded and
-it is not front-facing, it is assigned the color in the `color` uniform. The material
-needs its `side` set to `THREE.DoubleSide` otherwise three.js will cull them out if they are facing away
-from the camera,
-
-Any fragment that is discarded in the shadowNode will not cast shadows.
-
-### Updating uniforms
+Or can create the material in the script tag and use `<T>`'s `is` prop to
```
DefinitelyMaybe, line 289: Nit pick. Say less. No big worry either way tho

PR #1682, opened 2026-01-30
Bake shadows hook

let-josh:
moves the logic of \<BakeShadows> into a separate hook that the user can import instead of having to import the component.

i feel that if a component doesn't really use \<T>, it's probably better as a hook.


apps/docs/src/examples/extras/bake-shadows/basic/Scene.svelte
```diff
@@ -14,6 +14,10 @@
   useTask((delta) => {
     mesh.rotation.y += delta
   })
+
+  $effect(() => {
+    if (bake) useBakeShadows()
+  })
```
michealparks, line 20: This is an antipattern when designing hooks. As a general rule they should only be called during component setup. Making that a mandatory part of their design provides a good contract about how they generally behave. If we start calling hooks inside of effects, we break that contract and end up causing performance problems, errors, or large design limitations with all our hooks

apps/docs/src/examples/extras/bake-shadows/spaceships/Scene.svelte
```diff
@@ -25,6 +25,8 @@
     { name: 'Striker', position: [8, -1, -10] },
     { name: 'Zenith', position: [-1, 0, 13] }
   ]
+
+  useBakeShadows()
```
michealparks, line 29: I think that the changes in this example shows that component is more flexible and a better match. With the hook you can't do an inline suspense here anymore, and I think users would end up re-creating the component in order to easily toggle and / or throw it into a suspense group

DefinitelyMaybe:
:thinking:  sometimes hook, sometimes not hook. Always hook is guaranteed to not be best solution for all cases.

Thoughts or good to close?

let-josh:
i dunno. i'm still thinking that if something doesn't add anything to the scene or doesn't produce something, it should be implemented as a function or hook or otherwise. baking shadows, controls properties of the renderer. maybe a good compromise is turning it into just a plain function or removing it altogether and just making it an example of how to bake shadows.

i'm also think calling it "BakeShadows" is kind of a misnomer since the shadows aren't truly "baked" into the scene in the same way that lighting information is baked into a model. all the component/hook does is just disable the renderer's shadow map autoupdate.

i think it's probably best suited as a "performance optimization technique" example along the lines of *"if your scene is static and you have complex materials and/or geometries with lots of lights, here's how you can easily get a performance gain"*


```svelte
<!-- Scene.svelte -->
<script>

// turn off the renderer's shadow map's autoupdate

</script>

<LotsOfLights />
<LotsOfComplexMeshes />
```

so i think for now this can be closed. No one has any problems with `<BakeShadows>` so probably best to leave it like it is and if there's a shift in how we treat components/hooks/functions and their conventions, then we can come back to it.

PR #1751, opened 2026-04-04
[docs] Generate component signatures data

DefinitelyMaybe:
wip

notes

1. the switch away from ts-morph came down to [this getType() call](https://github.com/threlte/threlte/blob/9f2324f14e19115b618741bfeb2abc0118d305cb/apps/docs/scripts/generate.ts#L290). I asked ai to rework the script to use typescript directly which reduced the script time from 37 seconds to 2. If future self cannot make enough sense of the direct usage of typescript then the linked script state is where to start from.

2. I considered adding `"erasableSyntaxOnly": true,` to the tsconfig but that can be done in a different PR. keeping the new top level tsconfig barebones simply so that the types are correctly picked up with the `apps/docs/generate.ts` file.

3. Some current component signature types are the following:

```json
"type": {
  "name": "THREE.ToneMapping",
  "url": "https://threejs.org/docs/index.html#api/en/constants/Renderer"
},
```

which i have taken the easy road and instead only done:

```json
{
  "type": "ToneMapping"
},
```

In order to maintain the original, we'd need to add that url into the types.ts files. I leave that to some future work instead of now. Also adding the `THREE.` part would mean the following import into the types file: `import type * as THREE from 'three'`

4. tricky cases arise when trying to gathering type information because I've assumed that the `types.ts` file within a folder corresponds to a single component but thats not true within some folders i.e. InstancedMeshes, InstancedSprite, Portal.

5. patterns like extracting common props from somewhere: `import type { AudioProps as CommonAudioProps } from '../utils/useAudio.js'` and then ` & CommonAudioProps & { ... }` are hard to unpack. Conditional types, in the case of Grid are similarly difficult.

6. The easiest way to get svelte exports was via sveld but required that type/description info be present where the export is written i.e. 

```
/** doc comment of the thing */
export const thing: () => void = () => {}
```


packages/extras/src/lib/components/Align/types.ts
```diff
@@ -32,31 +32,25 @@ export type AlignProps = Props<Group, [{ align: () => void; ref: Group }]> & {
...
+    width: number
+    height: number
+    depth: number
+    boundingBox: Box3
+    center: Vector3
+    boundingSphere: Sphere
+    verticalAlignment: number
+    horizontalAlignment: number
+    depthAlignment: number
+  }) => void
```
DefinitelyMaybe: Other thing I could use some help on would be this kind of situation. It'd be easier to keep the `AlignEventData` type however we'd need to find a way to lookup & unpack that data type.

PR #1762, opened 2026-04-07
<Resize>: stop resizing every frame

let-josh:
fixes #1753

> [!NOTE]
> some of this was written prior to pulling the current code and seeing how the updated `useTask` api is done. so some of this may be refer to the old api but the incoming code is pretty easy to follow so i'll leave it like it is.

- sets `autoInvalidate = false` to prevent the resize calculation every frame. only recalc when a prop updates, a child is added/removed, or a child's reference update.
- adds missing `@default` docs and rewords some things.
- the `axis` prop now accepts `'min'` and `'max'` as values. `max` is the default and resizes along the maximum axis, `min` is the same way but for the `minimum` axis.

breaking changes:

- sets `auto`'s default value to `true` meaning that by default, a resize will happen when a child is added or removed and when a child reference updates. i think this the expected behavior and people can turn it off by setting `auto` to `false`.

```svelte
<!-- only resize when props update and once during component initialization -->
<Resize auto={false} />
```

- moves the "isInstanceOf" check in the plugin into the `$effect` so that `args.ref` is always tracked. it's a rare case of a child reference changing from something that is in the scene to something that is not. for example a mesh becoming a texture. it's something that should never happen but to be safe, `args.ref` needs to be read inside the `$effect`


let-josh:
i gotta fix some stuff that i forgot to pull down locally. stand by.

DefinitelyMaybe:
i like the work on the defaults and types here, good stuff 👍

let-josh:
okay i didn't realize how much useTask had changed but i think it's a lot simpler now. have a look at the updated code. i removed the `stage` prop cuz useTask doesn't accept a closure/getter of the stage.

DefinitelyMaybe (commented):
the best components are re-written 😄

packages/extras/src/lib/components/Resize/Resize.svelte
```diff
@@ -2,98 +2,70 @@
...
-    useStage,
-    useTask,
-    useThrelte,
-    type Plugin
-  } from '@threlte/core'
-  import { Box3, Group } from 'three'
+  import { isInstanceOf, observe, T, useStage, useTask, useThrelte } from '@threlte/core'
+  import { type Plugin } from '@threlte/core'
+  import { Box3, Group, Vector3 } from 'three'
   import InjectPlugin from '../InjectPlugin/InjectPlugin.svelte'
```
DefinitelyMaybe: I wonder where the `../InjectPlugin/InjectPlugin.svelte` came from, `injectPlugin` should probably come from `@threlte/core`.

can also fold that `type Plugin` import into the same statement.

packages/extras/src/lib/components/Resize/Resize.svelte
```diff
@@ -2,98 +2,70 @@
...
   const plugin: Plugin = (args) => {
-    if (!isInstanceOf(args.ref, 'Object3D')) return
-    observe.pre(
-      () => [args.ref],
-      () => {
-        if (auto) scheduleResizing()
-        return () => {
-          if (auto) scheduleResizing()
-        }
+    $effect(() => {
```
DefinitelyMaybe, line 55: cool. My query here would be what happens when a bunch of components enter/leave the sub-tree. Is there suddenly a big pile of effects calling resize (on the one frame or would it be multi?) or did the recent changes to tasks resolve this in a nicer way.

My understanding of plugins is that this effect would be added to each child <T> component which could be quite a few in some cases.

let-josh:
somehow there's a bunch of changes on this PR that don't pertain to the problem it tries to solve. so i'm gonna close this and open a new PR that only has the relevant changes on it.

PR #1778, opened 2026-04-16
pass stencil props through mask children snippet

### LINKED ISSUE #1777: <Mask> does not update material props if the mesh's material is updated
in a case like this:

```svelte
<Mask>
  {#if condition}
    <T is={material1} />
  {:else}
    <T is={material2} />
  {/if}
</Mask>
```

when `condition` updates the props of the "current" material are not updated.

right now the "stencil props" are spread onto the mesh's material once when the component is created and then whenever the relevant component props update. they are not spread when the entire mesh's material updates.

https://github.com/threlte/threlte/blob/a1c757e910d2861b6783d0736ccac851337cea4c/packages/extras/src/lib/components/Mask/Mask.svelte#L21

## Possible solution

i think we can solve this by passing the props that need to be spread onto the material through the children snippet

```svelte
<Mask>
  {#snippet children({ ref, getSpread })}
    {#if condition}
      <T is={material1} {...getSpread()} />
    {:else}
      <T is={material2} {...getSpread()} />
    {/if}
  {/snippet}
</Mask>
```

in the Mask component, that might look like this:

```svelte
<script>
  // props() and other stuff

  const spread: MaterialParameters = $derived({
    colorWrite,
    depthWrite,
    stencilWrite: true,
    stencilRef: id,
    stencilFunc: AlwaysStencilFunc,
    stencilFail: ReplaceStencilOp,
    stencilZFail: ReplaceStencilOp,
    stencilZPass: ReplaceStencilOp,
  });
</script>

<!-- <T is={mesh} {...otherStuff}> -->
{@render children?.({ ref, getSpread: () => spread })}
<!-- <T /> -->
```

> [!IMPORTANT]
> `spread` needs to be passed inside a closure so that reactivity is maintained.

let-josh:
fixes #1777 as described in that issue

i will add this functionality to the docs if we're cool with this change.

### DIFF
--- a/packages/extras/src/lib/components/Mask/Mask.svelte
+++ b/packages/extras/src/lib/components/Mask/Mask.svelte
@@ -5,7 +5,7 @@
 		*/
   import { T } from '@threlte/core'
   import { ReplaceStencilOp, AlwaysStencilFunc, Mesh } from 'three'
-  import type { MaskProps, Spread } from './types.js'
+  import type { MaskProps, MaskSpread } from './types.js'
 
   let {
     id = 1,
@@ -18,7 +18,7 @@
 
   const mesh = new Mesh()
 
-  const spread: Spread = $derived({
+  const spread: MaskSpread = $derived({
     colorWrite,
     depthWrite,
     stencilFail: ReplaceStencilOp,
--- a/packages/extras/src/lib/components/Mask/Mask.svelte
+++ b/packages/extras/src/lib/components/Mask/Mask.svelte
@@ -5,7 +5,7 @@
 		*/
   import { T } from '@threlte/core'
   import { ReplaceStencilOp, AlwaysStencilFunc, Mesh } from 'three'
-  import type { MaskProps } from './types.js'
+  import type { MaskProps, Spread } from './types.js'
 
   let {
     id = 1,
@@ -18,19 +18,20 @@
 
   const mesh = new Mesh()
 
+  const spread: Spread = $derived({
+    colorWrite,
+    depthWrite,
+    stencilFail: ReplaceStencilOp,
+    stencilFunc: AlwaysStencilFunc,
+    stencilRef: id,
+    stencilWrite: true,
+    stencilZFail: ReplaceStencilOp,
+    stencilZPass: ReplaceStencilOp
+  })
+
   $effect(() => {
-    const { material } = mesh
-
-    if (Array.isArray(material)) return
-
-    material.colorWrite = colorWrite
-    material.depthWrite = depthWrite
-    material.stencilWrite = true
-    material.stencilRef = id
-    material.stencilFunc = AlwaysStencilFunc
-    material.stencilFail = ReplaceStencilOp
-    material.stencilZFail = ReplaceStencilOp
-    material.stencilZPass = ReplaceStencilOp
+    if (Array.isArray(mesh.material)) return
+    Object.assign(mesh.material, spread)
   })
 </script>
 
@@ -40,5 +41,5 @@
   renderOrder={-id}
   {...props}
 >
-  {@render children?.({ ref: mesh })}
+  {@render children?.({ ref: mesh, getSpread: () => spread })}
 </T>
--- a/packages/extras/src/lib/components/Mask/types.ts
+++ b/packages/extras/src/lib/components/Mask/types.ts
@@ -1,7 +1,19 @@
 import type { Props } from '@threlte/core'
-import type { Mesh } from 'three'
+import type { Material, Mesh } from 'three'
 
-export type MaskProps = Props<Mesh> & {
+export type Spread = Pick<
+  Material,
+  | 'colorWrite'
+  | 'depthWrite'
+  | 'stencilWrite'
+  | 'stencilRef'
+  | 'stencilFunc'
+  | 'stencilFail'
+  | 'stencilZFail'
+  | 'stencilZPass'
+>
+
+export type MaskProps = Props<Mesh, [{ ref: Mesh; getSpread: () => Spread }]> & {
   /**
    * Each mask must have an id, you can have compound masks referring to the same id
    * @default 1
--- a/packages/extras/src/lib/components/Mask/types.ts
+++ b/packages/extras/src/lib/components/Mask/types.ts
@@ -1,7 +1,7 @@
 import type { Props } from '@threlte/core'
 import type { Material, Mesh } from 'three'
 
-export type Spread = Pick<
+export type MaskSpread = Pick<
   Material,
   | 'colorWrite'
   | 'depthWrite'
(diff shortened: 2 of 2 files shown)


michealparks:
What about instead deprecating `<Mask>` and replacing it with `<MaskMaterial>`?

To me this is another case of Drei providing an abstraction over the wrong object. `<Mask>` concerns itself entirely with the mesh's material with the exception of `id` being mapped to `renderOrder`, which is weird to begin with since it introduces the overhead of another name. It would be just as reasonable to mention `renderOrder` in the `<MaskMaterial>` docs and example.

What do you think?

let-josh:
i like that idea. `Mask` is all about the material. the only thing i can think of that we would miss out on is allowing people to choose which material they use. 99% of the time you're gonna set the `colorWrite = false` and it doesn't matter which  material type you use, in that case the ref would just be a `MeshBasicMaterial`. i think that's a fair limitation though. you get a `MeshBasicMaterial`, if you need something more, you can make your own mask implementation.

another idea i had was to return a pair from `useMask`.

```svelte
<script>
  const { maskProps, maskedOutProps } = useMask(); // don't really like these names though
</script>

<T.Mesh>
  <T.CircleGeometry />
  <T.MeshNormalMaterial {...maskProps} />
</T.Mesh>

<T.Mesh>
  <T.BoxGeometry />
  <T.MeshBasicMaterial {...maskedOutProps} />
</T.Mesh>
```

here you could apply the props to any kind of material

```svelte
<T.MeshNormalMaterial {...maskedOutProps} />
```

and it might be easier to create a material that you can reuse.

```svelte
<script>
  const { maskProps, maskedOutProps } = useMask();
  const maskMaterial = new MeshBasicMaterial(maskProps);
  const maskedOutMaterial = new MeshBasicMaterial(maskedOutProps);
</script>

<!-- Masks -->
<T.Mesh>
  <T.RingGeometry />
  <T is={maskMaterial} />
</T.Mesh>

<T.Mesh position.x={5}>
  <T.CircleGeometry />
  <T is={maskMaterial} />
</T.Mesh>

<!-- Masked-out Meshes -->
<T.Mesh>
  <T is={maskedOutMaterial} />
  <T.BoxGeometry />
</T.Mesh>

<T.Mesh>
  <T is={maskedOutMaterial} />
  <T.SphereGeometry />
</T.Mesh>

otherwise you have to bind to get a reference to the material and deal with `undefined` cases

<script lang="ts">
  let maskMaterial: MeshBasicMaterial = $state();
</script>

<MaskMaterial bind:ref={maskMaterial}/>

<!-- somewhere later in the scene i want to use the MaskMaterial on another mesh -->
{#if maskMaterial !== undefined}
  <T.Mesh>
    <T is={maskMaterial} />
  </T.Mesh>
{/if}

**Comment by DefinitelyMaybe** [comment]
wow, yeah, both `<mask>` and `useMask` are just about the props on the materials. definitely a case of Drei providing an abstraction over the wrong object.

**Comment by michealparks** [comment]
@let-josh I like the hook-only direction, what about:

```svelte
<!-- Mask.svelte -->

<script lang='ts'>
  
  const { renderOrder, mask } = useMask(1)
</script>

<T.Mesh {renderOrder}>
  <T.MeshStandardMaterial {...mask} />
</T.Mesh>
```

and then components using the mask:

```svelte
<script lang="ts">
  import { useMask } from '@threlte/extras'
  
  const { stencil } = useMask(1)
</script>

<T.Mesh>
  <T.TorusKnotGeometry />
  <T.MeshStandardMaterial {...stencil} />
</T.Mesh>
```

let-josh:
the hook-only implementation should be super easy to implement then. one thing i'm thinking is that `mask` needs to be derived so you can pass in `invert` as a getter to a `$state`.

something like this:

```ts
const useMask = (invert = () => false, id = 0) => {
  const stencil = {
    stencilRef: id,
    /* ...rest */
  };
  
  const stencilFunc = $derived(invert() ? NotEqualStencilFunc : EqualStencilFunc);
  
  const mask = $derived({
    stencilRef: id,
    stencilFunc,
    /* ...rest */
  });

  return {
    renderOrder: id,
    stencil,
    mask,
  };
}
```

and then you can invert the mask as needed

```svelte
<script>
  let invert = $state(false);
  const {
    stencil,
    mask,
    renderOrder
  } = $derived(useMask(() => invert));
</script>

<input type="checkbox" bind:checked={invert} />

<!-- mesh and material usage all the same -->
```

you can tell it'd be trivial to make `id` reactive and, although i can't think of a reason to, maybe we should. then both `mask` and `stencil` would be derived.

---

i can already tell that this is much simpler than `<Mask>`.

let-josh:
here's an example of what i'm thinking:

https://svelte.dev/playground/bdc04f96d6f44c87b74effa6c676ddc7?version=5.55.4

i made Mask into a class so that you don't have derive as you destructure.

```ts
const { mask, stencil, renderOrder } = $derived(useMask(/* ... */));
```

without deriving, the reactivity is broken. details [here](https://svelte.dev/docs/svelte/$derived#Destructuring). classes get around this because they turn deriveds into getters.

DefinitelyMaybe:
```svelte
<script>
  let invert = $state(false);
  const {
    stencil,
    mask,
    renderOrder
  } = $derived(useMask(() => invert));
</script>
```

looks like just a smidge too much imo. how about the variables you get back being reactive? or are we stuck with that `$derived`?

```svelte
<script>
  let invert = $state(false);
  const {
    stencil,
    mask,
    renderOrder
  } = useMask(1, invert);
</script>
```

let-josh:
Oh yeah you’re right. It returns an object *with* reactive props **not** a reactive object. I’ll update everything tomorrow.

let-josh:
as far as i know, we've got two options

1. use a class which will turn reactive states into getters

```svelte
<script>
  const mask = new Mask(() => invert, () => id);
</script>

<T.MeshBasicMaterial {...mask.stencil} />

<T.Mesh renderOrder={mask.renderOrder}>
  <T.MeshBasicMaterial {...mask.mask} />
</T.Mesh>
```

note that you can not do this:

```ts
const { stencil, mask, renderOrder} = new Mask();
```

because you "force" the value. this is like reading the `.current` prop of a `Tween` or `Spring`. you can't do

```ts
const { current } = new Tween();
$inspect(current);
```

2. we have to destructure a closure of the derived state to preserve reactivity

```svelte
<script>
  const {
    getMask,
    getStencil,
    getRenderOrder,
  } = useMask(() => invert, () => id);
</script>

<T.MeshBasicMaterial {...getStencil()} />

<T.Mesh renderOrder={getRenderOrder()}>
  <T.MeshBasicMaterial {...getMask()} />
</T.Mesh>
```

as soon as you "force" the value, the reactivity breaks.

i prefer the first option because it's most similar to what svelte does for its reactive classes such as Spring and Tween. Even internally they use a [ReactiveValue](https://github.com/sveltejs/svelte/blob/6a149666c1772b443c0db9385ee7521275b27487/packages/svelte/src/reactivity/reactive-value.js#L6) class for things like [reactive window props](https://github.com/sveltejs/svelte/blob/6a149666c1772b443c0db9385ee7521275b27487/packages/svelte/src/reactivity/window/index.js). but only if all the other reactive hooks in threlte are gonna follow the same direction.

PR #1784, opened 2026-04-19
[docs] add `llm.txt`

michealparks:
Adds `llm.txt` for improved LLM docs parsing and fixes a number of docs errors

### DIFF
--- a/packages/extras/src/lib/audio/AudioListener/autoResume.ts
+++ b/packages/extras/src/lib/audio/AudioListener/autoResume.ts
@@ -1,0 +1,43 @@
+/**
+ * Browser autoplay policies keep an `AudioContext` in `suspended` state until
+ * the user has interacted with the page. THREE's `AudioContext` is a module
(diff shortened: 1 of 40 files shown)


packages/extras/src/lib/audio/AudioListener/autoResume.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
DefinitelyMaybe, line 1: nice little extra feature here. how did this come about?

DefinitelyMaybe:
where is the llm.txt file? i can't see it in this PR

michealparks:
@DefinitelyMaybe 
* https://threlte.xyz/llms.txt
* https://threlte.xyz/llms-full.txt
* https://github.com/threlte/threlte/blob/main/apps/docs/src/pages/llms-full.txt.ts

DefinitelyMaybe:
`<Example path="plugins/lookAt" />` Phew! I was worried about the examples being expanded into their full html, all good.

PR #1785, opened 2026-04-19
[docs] update plugin page

DefinitelyMaybe:
I just happen to be looking at our [docs page for plugins](https://threlte.xyz/docs/learn/advanced/plugins/#bvh-raycast-plugin) and thought maybe we didn't need as many details on bvh on that page. Like, there's the simple case above with `lookAt`, and then heres a link to a more detailed plugin we have.


apps/docs/src/content/learn/advanced/plugins.mdx
```diff
@@ -182,51 +182,7 @@ This is en example implementation that adds the property `lookAt` to all `<T>` c
...
-
-  bvhRaycasting()
-</script>
-
-<T.Mesh>
-  <T.MeshBasicMaterial />
-  <T.BoxGeometry />
-</T.Mesh>
-```
+Another good example is our [bvh plugin](/docs/reference/extras/bvh) that implements [BVH raycasting](https://github.com/gkjohnson/three-mesh-bvh) on all child meshes and geometries.
```
let-josh:
```suggestion
Threlte's [bvh plugin](/docs/reference/extras/bvh) replaces three's default raycasting with a [bounding-volume heirachy](https://github.com/gkjohnson/three-mesh-bvh) implementation.
```

let-josh (commented):
that example code is a little extraneous.

good catch.

PR #1786, opened 2026-04-20
simplify lookAt plugin example code

let-josh:
got bored before calling it a night so i simplified some code.

### DIFF
--- a/apps/docs/src/examples/plugins/lookAt/Scene.svelte
+++ b/apps/docs/src/examples/plugins/lookAt/Scene.svelte
@@ -2,12 +2,17 @@
   import { MathUtils } from 'three'
   import { T, useTask } from '@threlte/core'
   import { injectLookAtPlugin } from './lookAtPlugin.svelte'
+  import type { LookAt } from './lookAtPlugin.svelte'
 
-  const cubePos = $state<[number, number, number]>([0, 0.8, 0])
+  let cubePos: LookAt = $state.raw([0, 0.8, 0])
 
-  useTask(() => {
-    cubePos[0] = Math.sin(Date.now() / 1000) * 2
-    cubePos[2] = Math.cos(Date.now() / 1000) * 2
+  let time = 0
+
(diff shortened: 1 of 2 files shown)


apps/docs/src/examples/plugins/lookAt/Scene.svelte
```diff
@@ -2,12 +2,17 @@
   import { MathUtils } from 'three'
   import { T, useTask } from '@threlte/core'
   import { injectLookAtPlugin } from './lookAtPlugin.svelte'
+  import type { LookAt } from './lookAtPlugin.svelte'
 
-  const cubePos = $state<[number, number, number]>([0, 0.8, 0])
+  let cubePos: LookAt = $state.raw([0, 0.8, 0])
```
michealparks:
```suggestion
  const cubePos = $state<[number, number, number]>([0, 0.8, 0])
```
Would keep as is due to better readability as a docs example. `LookAt` requires a lookup for a simple type.

apps/docs/src/examples/plugins/lookAt/lookAtPlugin.svelte.ts
```diff
@@ -1,8 +1,11 @@
 import { injectPlugin, isInstanceOf, useThrelte } from '@threlte/core'
+import type { Object3D } from 'three'
+
+export type LookAt = Parameters<Object3D['lookAt']>
 
 export const injectLookAtPlugin = () => {
   injectPlugin<{
-    lookAt: [number, number, number]
+    lookAt: LookAt
```
michealparks: Same suggestion, `LookAt` is too simple and the tradeoff cost is indirection and mentally noting the meaning of additional var names

michealparks (changes requested):
Looks good pending a few comments!

let-josh:
i removed the replaced the type alias with its real type. should be good to go now.

michealparks (approved):

PR #1804, opened 2026-05-07
fix: reactivity involving CubeCamera, Environment, VirtualEnvironment, CubeEnvironment

let-josh:
in preparation for implementing [Fisheye](https://drei.docs.pmnd.rs/portals/fisheye#fisheye) which is kinda sorta related to CubeCamera and Environment, i wanted to revisit these and clean them up and bring them up to spec with other reactive systems.

## updates

there are a number of changes that i will try to list here:

### useCubeCamera

1. `useCubeCamera` no longer creates its own render target. instead you must pass in a closure to the render target.

```ts
const target = new WebGLCubeCameraTarget(512); // or since its passed as a closure, it could be derived.
const near = $state(0.1);
const far = $state(1000);
const { camera } = useCubeCamera(() => target, () => near, () => far);
```

this is because you may want to use the target for other things and useCubeCamera's usage does not imply that a render target is created but rather *used*. in this way `useCubeCamera` is more like a reactive wrapper around three's `CubeCamera` that has a couple of internal effects that update properties of the `CubeCamera` instance.

2. the returned value of `useCubeCamera` is now an object with a closure to the created `CubeCamera` instance.

this is for three reasons. Firstly, because `useCubeCamera` no longer creates a render target, there's no need to return one. Secondly, if you need the cube camera's current render target, it's already available through `camera.renderTarget`. And lastly, because the cube camera instance is derived, a closure needs to be returned so that the value can be read in effects and is not stale when returned from `useCubeCamera`.

The last point has a slight downside in that you have to call the returned value when you want to use it. for example:

```ts
const { camera } = useCubeCamera(/* */);
$effect(() => {
  camera().update(renderer, scene);
})
```

but this is like other apis that we already have, such as `shouldRender` from `useThrelte`.

### \<CubeCamera>

1. the `CubeCamera` component no longer passes a render target to its children. this is again because `CubeCamera` uses `useCubeCamera` internally and the current render target is available already as a property of the camera. you'd also probably already have access to the render target in the script anyway. a good example of how all this comes together is in the updated `VirtualEnvironment` component and its examples.

### useEnviroment

1. the effects that make `useEnvironment` work such as restoring previous backgrounds and environment maps have seriously been cleaned up but the api is also different. this is again to be more inline with other apis in threlte. `useEnvironment` now accepts a closure to the environment texture, a closue to the scene to modify, and a closure to a boolean indicating if the scene's background should be updated to the environment texture. these are all in order to make `useEnvironment` fully reactive.

usage will typically look like this:

```ts
let isBackground = $state(true);
const { scene } = useThrelte();
const environment = await (new HDRLoader()).loadAsync('some-hdr-file.hdr');
useEnvironment(
  () => scene,
  () => environment,
  () => isBackground
);
```

`useEnvironment` is heavily used in all the `*Environment` components so you can see how its being used there.

### miscellaneous

- all the examples have been updated to use the new apis
- RGBELoader is deprecated and has been replaced with HDRLoader
- in `*Environment` components, texture props default to null if they are not provided. this avoids any `if (texture === undefined)` checks since `null` is a valid value for `scene.environment` and `scene.background`


apps/docs/src/examples/extras/cube-camera/Scene.svelte
```diff
@@ -77,8 +78,6 @@
 >
   <OrbitControls
     enableDamping
-    enablePan={false}
-    enableZoom={false}
```
michealparks, line 81: I'd keep these unless there's a good reason, this stops scrolling through examples, giving a worse user experience when people are trying to get information

apps/docs/src/examples/extras/virtual-environment/manual-updates/App.svelte
```diff
@@ -4,7 +4,7 @@
 </script>
 
 <div>
-  <Canvas>
+  <Canvas renderMode="on-demand">
```
michealparks: This is the default, why make it explicit?

packages/extras/src/lib/components/CubeCamera/CubeCamera.svelte
```diff
@@ -18,10 +18,19 @@
     ...props
   }: CubeCameraProps = $props()
 
-  export const { camera, renderTarget } = useCubeCamera(
+  const renderTarget = $derived(new WebGLCubeRenderTarget(resolution))
```
michealparks:
```suggestion
  const renderTarget = new WebGLCubeRenderTarget(untrack(() => resolution))
  $effect(() => renderTarget.setSize(resolution, resoution))
```
I think this should work, and in this version we're not rebuilding a class for every resolution change

packages/extras/src/lib/components/environment/Environment/Environment.svelte
```diff
@@ -4,7 +4,7 @@
 >
   const loaders: {
     exr?: EXRLoader
-    hdr?: RGBELoader
+    hdr?: HDRLoader
```
michealparks: We shouldn't switch over to HDRLoader until RGBELoader is close to being removed. The issue is that we'll have to bump Three's peer dependency to r184 which puts a big burden on users to upgrade immediately to latest Three if they want to use the newest extras.

packages/extras/src/lib/components/environment/utils/useEnvironment.svelte.ts
```diff
@@ -1,68 +1,33 @@
...
-  scene: Scene
-  texture?: Texture | undefined
-  isBackground?: boolean
-}
-
-export const useEnvironment = (options: EnvironmentOptions) => {
+import { useThrelte } from '@threlte/core'
+import type { Scene } from 'three'
+
+export const useEnvironment = (
```
michealparks, line 4: I like this signature much more 💯

packages/extras/src/lib/hooks/useCubeCamera.svelte.ts
```diff
@@ -3,30 +3,25 @@ import { isInstanceOf } from '@threlte/core'
...
-    const _resolution = resolution()
-    renderTarget.setSize(_resolution, _resolution)
-  })
+  const camera = $derived(new CubeCamera(DEFAULT_NEAR, DEFAULT_FAR, renderTarget()))
 
   $effect(() => {
     const _near = near()
     for (const child of camera.children) {
-      if (isInstanceOf(child, 'PerspectiveCamera')) {
+      if (isPerspectiveCamera(child)) {
```
michealparks: I'd keep as isInstanceOf. Not much readability gained here and we pay a tax of an additional function call.

packages/extras/src/lib/hooks/useCubeCamera.svelte.ts
```diff
@@ -36,22 +31,15 @@ export const useCubeCamera = (
...
-    return () => {
-      renderTarget.dispose()
-    }
-  })
-
   return {
-    camera,
-    renderTarget
+    // this has to be a getter or closure since cube the camera is derived
+    camera: () => camera
```
michealparks:
```suggestion
    camera: {
      get current() {
        return camera
      }
    }
```
This is the idiomatic pattern for reactive props returned from hooks

let-josh:
@michealparks i updated the code with the comments you made but i have two errors that i'm having trouble tracking down. they're both in `VirtualEnvironment.svelte` and `CubeCamera.svelte` from what i can tell. the error is:

```
Reactive `$state(...)` proxies and the values they proxy have different identities. Because of this, comparisons with `===` will produce unexpected resultshttps://svelte.dev/e/state_proxy_equality_mismatch
```

and it errors so much that it triggers this error:

```
chunk-UGNWK5NI.js?v=842a2d1f:382 Uncaught (in promise) Svelte error: effect_update_depth_exceeded
Maximum update depth exceeded. This typically indicates that an effect reads and writes the same piece of state
```

which causes a crash.

i'll have to debug it when i have some more time.

Whatever the error is, it's something that both those components share or do similarly.

once those two things are fixed, i'll post an update here.

let-josh:
i was able to narrow it down `useCubeCamera`.

it doesn't throw an error if you return this kind of object:

```ts
const useCubeCamera =() => {
  const camera = $derived(new CubeCamera());
  return {
    camera: () => camera
  }
}
```

but it does if you return a getter to the camera:

```ts
const useCubeCamera = () => {
  const camera = $derived(new CubeCamera());
  return {
    get current() {
      return camera;
    }
  }
}
```

and currently, updating the resolution does not cause the cube camera's render target to be updated so i gotta investigate that as well. - it works if we derive a new target each time `resolution` updates but i agree that it would be nice and better to reuse the target and update its size as `resolution` changes rather than make a new one each time.

i don't foresee people updating `resolution` often, but maybe they can render to a lower resolution target if the mesh's material that uses the target's texture is far away from the camera. kinda like a level-of-detail approach for render target textures.

let-josh:
ah okay. so the actual problem was this:

```html
<T is={camera} />
```

in those examples needed to be this:

```html
<T is={camera.current} />
```

that was a tough one to track down. now i just have to fix rerendering to the texture when the resolution updates and check that all the examples still work and we're ready to go i think.

let-josh:
i can't figure out why this doesn't work:

```ts
const resolution = $state(512) // actually from props() in CubeCamera.svelte
const target = new WebGLCubeRenderTarget();
$effect(() => {
  target.setSize(resolution, resolution);
});
```

it should update the target's texture when resolution updates. the texture should get more fidelity when resolution increases and less when it decreases but nothing seems to happen.

note that if you pass an explicit size in the render target's constructor, it can update the texture correctly for sizes <= resolution. so this:

```ts
// same
const target = new WebGLCubeRenderTarget(512);
// same
```

works for resolutions of 32, 64, 128, 256, and 512 but not 1024, for example.

---

i wanted to test whether it was threlte or three.js so i made a little example without threlte [here](https://svelte.dev/playground/921caebf2b474221bad19ace624ad227?version=5.55.5) and it appears to work properly. the cube gets more detailed as the resolution goes up

so i'm not sure what the problem is.

i'm gonna push up what i have so far and ask for some help to narrow this one down.

michealparks (approved):
🚀

PR #1819, opened 2026-05-12
[@threlte/extras] Add `<Wobble>`

michealparks:
Time to wobble it up!

Docs page: https://threlte.xyz/previews/pr-1819/docs/reference/extras/wobble/

`<Wobble>` is a material modifier - like `<Wireframe>` or `<ShadowAlpha>` - that wobbles any material.

wobble


packages/extras/src/lib/components/Wobble/Wobble.svelte
```diff
@@ -0,0 +1,403 @@
...
+      return sum / 1.875;
+    }
+
+    // Rodrigues' rotation formula: rotate v around unit-vector k by angle.
+    vec3 wobbleRotate(vec3 v, vec3 k, float angle) {
+      float c = cos(angle);
+      float s = sin(angle);
+      return v * c + cross(k, v) * s + k * dot(k, v) * (1.0 - c);
+    }
+  `
```
DefinitelyMaybe, line 109: Where did all of this come from? The [OG three.js example](https://github.com/mrdoob/three.js/blob/master/examples/webgl_materials_modified.html#L94) doesn't have this much going on, nor [drei's component](https://github.com/pmndrs/drei/blob/master/src/core/MeshWobbleMaterial.tsx)

michealparks: a little shader research, a little LLM chatting, a little of me wanting more features than what I saw 😎

PR #1822, opened 2026-05-14
[extras/interactivity] feat: add `eventOptions` option

### LINKED ISSUE #1821: [extras] Interactivity breaks default scrolling behavior
In Safari, when the `html` element has `overscroll-behavior: none` CSS, a scene which calls `interactivity()` will break default scroll behavior when the cursor is over the canvas. 

Could this be due to the `onwheel` handler interfering with the default scrolling? Is it possible to disable listening to certain events entirely if we don't need them for our scene? 

**To reproduce:**
1. In Safari, visit the [example in the docs](https://threlte.xyz/docs/reference/extras/interactivity).
2. Inspect the page and add `overscroll-behavior: none` to the `<html>` element.
3. Using a scroll wheel or trackpad, scroll the page while the cursor is over the canvas. The input is ignored; the page does not scroll.

**Notes:**
- When testing locally, removing the `interactivity()` call fixes the issue
- I cannot reproduce this issue in Chrome

**Environment** 
- macOS 26.4.1 
- Safari 26.4 
- MBP M2 Max
- threlte/core 8.5.13
- threlte/extras 9.16.0

ypessoa:
Adds `eventOptions` for `interactivity`:
- `passive` allows overriding default `passive` option for a given DOM event
- `enabled` allows disabling certain DOM events that are not needed by the user, preventing their listeners from being added

This PR also adds an "Event options" section to the interactivity doc. 

Fixes https://github.com/threlte/threlte/issues/1821.

#### Usage example: 
```
interactivity({ 
  eventOptions: {
     wheel: { passive: true },
     click: { enabled: false },
  }
})
  ```

### DIFF
--- a/.changeset/weak-colts-exercise.md
+++ b/.changeset/weak-colts-exercise.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": minor
+---
+
+Add eventOptions option to interactivity plugin
--- a/apps/docs/src/content/reference/extras/interactivity.mdx
+++ b/apps/docs/src/content/reference/extras/interactivity.mdx
@@ -216,6 +216,23 @@ You can filter and sort events by passing a `filter` to the `interactivity` plug
 </script>
 ```
 
+## Event options
+
+Certain configuration options are available for DOM events via `eventOptions`. Setting `enabled: false` for a specified event will prevent its respective listener from being added.
+
+```svelte title="Scene.svelte"
+<script>
+  import { interactivity } from '@threlte/extras'
+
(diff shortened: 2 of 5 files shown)


michealparks:
Thanks for updating this @ypessoa! I removed the `enabled` option for now until someone presents a use case for it, especially since it could end up producing weird behavior (disabling `pointermove` would make `onpointerenter` / `leave` and `useCursor` not work, etc). Merging this now!

michealparks (approved):

PR #1823, opened 2026-05-15
[@threlte/rapier] Reactivity fixes, performance improvements, new docs examples

michealparks:
### Changelog

* Fix all reactivity issues with prop updates on Rapier components.
* Performance improvements based on Chrome Devtools profiling
* New examples for all joint-related hooks
* New example on the getting started page attempting to showcase and stress test most @threlte/rapier features

game


packages/rapier/src/lib/components/Colliders/AutoColliders/AutoColliders.svelte
```diff
@@ -105,25 +108,21 @@
...
   export const refresh = () => create()
 
-  onMount(() => {
-    create()
+  $effect(() => {
+    return untrack(() => {
+      create()
+      return cleanup
+    })
   })
```
DefinitelyMaybe, line 136: here @michealparks, could I check my understanding of this one. Is an $effect preferred here because its a small simplification that does the same thing but we don't need any imports to accomplish - or is there other nuances here?

packages/rapier/src/lib/components/Colliders/AutoColliders/AutoColliders.svelte
```diff
@@ -133,6 +132,16 @@
...
+
+  $effect(() => {
+    if (colliders) {
+      return untrack(() => {
+        if (colliders) {
+          return oncreate?.(colliders)
+        }
+      })
+    }
+  })
```
DefinitelyMaybe, line 156: wait what? was the cleanup functions returned by oncreate events not being called before?

packages/rapier/src/lib/components/Colliders/Collider/Collider.svelte
```diff
@@ -47,93 +46,107 @@
   }: ColliderProps<TShape, TMassDef> = $props()
 
   const object = new Object3D()
-
-  const { updateRef } = useCreateEvent(oncreate)
   const rigidBody = useRigidBody()
   const parentRigidBodyObject = useParentRigidbodyObject()
-  const hasRigidBodyParent = !!rigidBody
+  const hasRigidBodyParent = $derived(rigidBody.current !== undefined)
```
DefinitelyMaybe, line 52: oh hell yeah! good improvement here 👍

packages/rapier/src/lib/components/Colliders/Collider/Collider.svelte
```diff
@@ -47,93 +46,107 @@
...
+    return () => {
+      rapierContext.removeColliderFromContext(currentCollider)
+      collisionGroups.removeColliders([currentCollider])
+      world.removeCollider(currentCollider, true)
+      collider = undefined
     }
   })
 
-  $effect.pre(() => {
+  $effect(() => {
```
DefinitelyMaybe, line 126: do tell, how was `.pre` not working quite right

packages/rapier/src/lib/components/Colliders/Collider/Collider.svelte
```diff
@@ -47,93 +46,107 @@
...
      * Add collider to context
      */
-    rapierContext.addColliderToContext(collider, object, events)
+    rapierContext.addColliderToContext(currentCollider, object, {
+      oncollisionenter,
+      oncollisionexit,
+      oncontact,
+      onsensorenter,
+      onsensorexit
+    })
```
DefinitelyMaybe: ah! did these references change over a components lifecycle hence couldn't be static references?

packages/rapier/src/lib/components/Colliders/AutoColliders/AutoColliders.svelte
michealparks: They were, it was just a bit more confusing to read before since it was in an `onDestroy`

packages/rapier/src/lib/components/Colliders/Collider/Collider.svelte
michealparks: Exactly, if someone swapped out this callback it wouldn't be registered

packages/rapier/src/lib/components/Colliders/Collider/Collider.svelte
michealparks: I'm slowly cleaning up $effect.pre's that don't make sense. We shouldn't use them unless we have an explicitly good reason - and that reason is that some other effect needs to read the updated value of something this effect is writing. That is something that we should rarely do since it's fragile anyways. For most true effects we should use $effect since `.pre` is invoked eagerly and doesn't get batched updates

packages/rapier/src/lib/components/Colliders/AutoColliders/AutoColliders.svelte
michealparks: I've been staying away from onMount / onDestroy since they occasionally cause reactivity issues

michealparks: This also just matches exactly how the T component handles `oncreate` (and is way less code than what we had previously)

PR #1827, opened 2026-05-23
improve UvMaterial docs

let-josh:
adds a little more clarity to the UvMaterial docs page

### DIFF
--- a/apps/docs/src/content/reference/extras/uv-material.mdx
+++ b/apps/docs/src/content/reference/extras/uv-material.mdx
@@ -20,7 +20,6 @@ A material that visualizes the uvs of a geometry. Uvs are typically used for
 texture sampling in the shader. `UvMaterial` is meant to give you a sense of
 how textures will be mapped onto a mesh.
 
-
 <Example path="extras/uv-material" />
 
 ## Usage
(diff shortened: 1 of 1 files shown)


apps/docs/src/content/reference/extras/uv-material.mdx
```diff
@@ -38,12 +40,23 @@ Simply import the component and add it as a child to a mesh
 
 ## Caveats
 
-The geometry must have a "uv" attribute in order for this material to work. If you're creating your own geometry, be sure to assign the "uv" attribute.
+The mesh's geometry must have a "uv" attribute in order for `UvMaterial` to work properly. If
```
michealparks, line 43: I'm curious, why not just add a uv attribute to the geometry if it doesn't exist and then remove it on unmount in this case? that seems pretty safe and non-invasive.

let-josh: i dunno if there's a good way to generate missing uv data cuz uvs could be any set of numbers that are in that [0, 1] range. so maybe we'd look at the `position` attribute of the geometry and go off that somehow. but i could see it being difficult for complex geometries.

let-josh: but you know, if they're using texture maps, they're gonna have to have uv data anyway. i don't think this:

`mesh.material.map = texture;`

will "work" unless your geometry has a uv attribute.

michealparks: Yeah, good points. I'll hold off on anything like this since it seems complex / error prone

michealparks (approved):

PR #1832, opened 2026-06-04
[@threlte/extras] Add hooks for accessing all controls

### LINKED ISSUE #1385: feature proposal - controls as part of the context
R3F exposes the current controls in a context. Now that three.js has a [base class]((https://threejs.org/docs/index.html?q=Controls#api/en/extras/Controls)) for all controls. i think we should do something where we have controls as part of the threlte context. That way we can enable/disable them as needed and facilitate when two controls need to be aware of each other.

We have a problem now that TransformControls needs to disable OrbitControls/TrackballControls upon interaction so that the camera doesn't spin around as you're trying to transform a mesh.

https://github.com/threlte/threlte/blob/94082bf1ba88d3a59e32a8ef70d7b0afc13b38ad/packages/extras/src/lib/components/controls/TransformControls/TransformControls.svelte#L32-L60

this doesn't really scale well. if we added `controls` to the context then we could do something like this:

```ts
// inside TransformControls.svelte
const { controls } = useThrelte()

// Three.TransformControls
const tc = new TransformControls()

let dragging = $state(false)

// these apis doesn't exist but for sake of argument pretend that they did
tc.on('change', () => {
  dragging = tc.dragging 
})

// turn off controls when dragging
$effect(() => {
  if (dragging) {
    $controls?.enabled = false
  }
  return () => {
    $controls?.enabled = true
  }
})
```

we currently have a [useControlsContext](https://github.com/threlte/threlte/blob/main/packages/extras/src/lib/components/controls/useControlsContext.ts) that has entries for all the different controls but we can probably consolidate this to the following

```ts
import type { Controls } from 'three'

type ControlsContext = {
  controls: Writable<Controls | null>
}

export const useControlsContext = (): ControlsContext => {
  return useThrelteUserContext<ControlsContext>('threlte-controls', {
    controls: writable(null)
  })
}
```

and made available on the object returned by `useThrelte`

```ts
const { controls } = useThrelte();
```

`useThrelte` would most likely utilize useControlsContext in it's function body, etc. etc. I think we'd keep it a store for now or a CurrentWritable in case it needs to be used in a callback or `useTask`.

---

This approach should make it much easier to disable controls when something needs to take priority of the event. for example if we wanted a port of [drei's DragControls](https://drei.docs.pmnd.rs/gizmos/drag-controls#dragcontrols), having `controls` in the context would make it much easier to disable OrbitalControls or TrackballControls at the start beginning of dragging. and then re-enable them at the end. This is evident in [drei's DragControls source code](https://github.com/pmndrs/drei/blob/eb449ed100078f9df7747c0672c8dba644849fd3/src/web/DragControls.tsx#L71). 

It really gives the user more say-so in when and how to enable/disable controls.

michealparks:
### Overview

This PR adds hooks to access the ref for `cameraControls`, `trackballControls`, `orbitControls` and `transformControls` extras components.

Fixes https://github.com/threlte/threlte/issues/1385

### DIFF
--- a/packages/extras/src/lib/components/CameraControls/useCameraControls.ts
+++ b/packages/extras/src/lib/components/CameraControls/useCameraControls.ts
@@ -1,0 +1,36 @@
+import type CameraControls from 'camera-controls'
+import { fromStore } from 'svelte/store'
+import { useControlsContext } from '../controls/useControlsContext.js'
+
+/**
+ * ### `useCameraControls`
+ *
+ * Access the `CameraControls` instance created by the `<CameraControls>`
+ * component mounted within the current `<Canvas>`. Useful for imperatively
+ * driving the camera (`setLookAt`, `dolly`, `rotate`, …) from sibling or
+ * parent components without a `bind:ref`.
+ *
+ * `current` is `undefined` until `<CameraControls>` has mounted, and again
+ * after it unmounts, so reads are always safe.
+ *
+ * ```svelte
+ * <script>
+ *   import { useCameraControls } from '@threlte/extras'
+ *
+ *   const controls = useCameraControls()
+ *
+ *   const frameScene = () => {
+ *     controls.current?.setLookAt(5, 5, 5, 0, 0, 0, true)
+ *   }
+ * </script>
+ * ```
+ */
+export const useCameraControls = (): { readonly current: CameraControls | undefined } => {
+  const { cameraControls } = useControlsContext()
+  const controls = fromStore(cameraControls)
+  return {
+    get current() {
+      return controls.current
+    }
+  }
+}
--- a/.changeset/controls-accessor-hooks.md
+++ b/.changeset/controls-accessor-hooks.md
@@ -1,0 +1,5 @@
+---
+"@threlte/extras": minor
+---
+
+Add `useCameraControls`, `useOrbitControls`, `useTrackballControls` and `useTransformControls` hooks
--- a/apps/docs/src/content/reference/extras/camera-controls.mdx
+++ b/apps/docs/src/content/reference/extras/camera-controls.mdx
@@ -86,6 +86,23 @@ A camera can also optionally be passed to the controls as a prop.
 Finally, if the component is created without an attached camera
 it will use the scene's default camera as provided by `useThrelte`.
 
+## `useCameraControls`
+
+To access the underlying `CameraControls` instance from another component —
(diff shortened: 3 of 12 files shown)


packages/extras/src/lib/components/CameraControls/useCameraControls.ts
```diff
@@ -0,0 +1,36 @@
...
+ *     controls.current?.setLookAt(5, 5, 5, 0, 0, 0, true)
+ *   }
+ * </script>
+ * ```
+ */
+export const useCameraControls = (): { readonly current: CameraControls | undefined } => {
+  const { cameraControls } = useControlsContext()
+  const controls = fromStore(cameraControls)
+  return {
+    get current() {
```
let-josh, line 32: i think `fromStore`'s `.current` is [already a getter](https://github.com/sveltejs/svelte/blob/3d5c4ab620a597dfa296486e96642cff88bf90ec/packages/svelte/src/store/index-client.js#L153-L167) so maybe we can just do

```ts
return fromStore(cameraControls)
```

for all of these

let-josh (approved):
looks good to me. i'd like to have just a single `useControls` hook but it might be bothersome to have to narrow its type everywhere. although i wonder if most of the time you'd only need to access/call methods that live on [Controls](https://threejs.org/docs/#Controls), for example to enable/disable the "global" controls. then in rare cases, where you need to do something specific to a certain kind of controls, say orbit controls, then you'd have a good reason to narrow its type or pass the "global" orbit controls around some other way.

it kinda feels like there should be some distinction between controls that control the camera (OrbitControls, TrackballControls, ArcballControls, etc.) and controls that control specific objects in the scene like TranslateControls, and drei's DragControls, drei's PivotControls, etc. but that'd have to be change in three.js.

probably best to leave it like this for now.

packages/extras/src/lib/components/CameraControls/useCameraControls.ts
michealparks: correct, but putting `fromStore` in the getter will cause the store to getter translation to occur every time the getter is accessed. more efficient to cache it

michealparks:
@let-josh Yeah, I struggled with this too. At the end of the day I don't want to bloat `@threlte/core` and I think that users can easily compose their own `useControls` hook if it makes sense for them in the context of just using something like the `enabled` flag.

PR #1838, opened 2026-07-18
docs: add Theatre.js animation tutorial

### LINKED ISSUE #549: Introductory tutorial: `@threlte/theatre`
Each package should have a solid introductory tutorial, analogous to `@threlte/core`'s [Your First Scene](https://threlte.xyz/docs/learn/getting-started/your-first-scene).

`@threlte/theatre` has a small enough API that it mostly should be possible to cover it in one tutorial. The tutorial should have a coherent narrative: the learner should be making a scene that makes sense, and learning about `@threlte/theatre` as they build it.

If you consider contributing this, please discuss your plan here before submitting a PR.

ALPaseeno:
Adds an introductory tutorial for `@threlte/theatre` under Learn → Advanced
(`/docs/learn/advanced/animating-with-theatre`).

The tutorial builds one small scene and animates it end to end, with a runnable
example embedded at each step:

1. a plain Threlte scene — box on a floor, one light, fixed camera
2. wrapping the app in `<Theatre>` and making the box editable with `<SheetObject>` + `Transform`
3. exposing a non-transform property with `<Sync color />`
4. keyframing in the Studio, then exporting `state.json` and loading it back via `config`
5. playing it back with `<Sequence>` (`autoplay`, looping) — plus pointers for driving playback manually and hiding the Studio in production

It also fixes one link along the way: the `<Sync>` mention in the text now points
at the `<Sync>` reference page instead of the `<SheetObject>` page.

Audio coverage is left for a follow-up, per the issue thread.

Docs-only, no changeset.

Closes #549

### DIFF
--- a/apps/docs/src/content/learn/advanced/animating-with-theatre.mdx
+++ b/apps/docs/src/content/learn/advanced/animating-with-theatre.mdx
@@ -114,7 +114,7 @@ On its own, `<Theatre>` only opens the editor — nothing is animatable yet. To 
 The Theatre.js Studio now overlays the scene, with a **Box** entry in the outline on the left. Select it and drag its transform — you're editing the scene live.
 
 <Tip type="info">
-  `<Theatre>` is the convenient all-in-one wrapper. When you need more control — multiple sheets, a custom project name, or showing the studio only in development — reach for [`<Project>`](/docs/reference/theatre/project), [`<Sheet>`](/docs/reference/theatre/sheet) and [`<Studio>`](/docs/reference/theatre/studio) directly. We'll use those in the last step.
+  `<Theatre>` is the convenient all-in-one wrapper. When you need more control — multiple sheets, a custom project name, or showing the studio only in development — reach for [`<Project>`](/docs/reference/theatre/project), [`<Sheet>`](/docs/reference/theatre/sheet) and [`<Studio>`](/docs/reference/theatre/studio) directly — see [Where to go next](#where-to-go-next).
 </Tip>
 
 ## Syncing more properties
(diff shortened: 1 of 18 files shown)


DefinitelyMaybe:
good start!

- The animation I saw didn't rotate 360 degrees but you've said `spin a full turn`, could I has full turn :smile: 
- I was looking further into the theatre doc's because I was thinking "whats the keyboard shortcuts for play, pause, add keyframe etc" and I'm realizing a couple of things:
  - A lot of folks answers will be in their docs so nudging folks that way is in our best interests. Find a couple of points where likely initial questions are going to come up within the readers mind, i.e. "but how do i _" and add a link to the theatre.js docs for said answer. "how do i [add keyframes](https://www.theatrejs.com/docs/latest/manual/sequences#addingremoving-keyframes)" is one such place that I'd like.
  - There is a couple of different versions of the theatre docs. One that we've linked to at the bottom of our `Getting Started` page is the [0.5](https://www.theatrejs.com/docs/0.5/manual/Studio). Because we don't pin a version in the `@threlte/theatre` package we should update those to `latest`.
- I was wrong about placement of this tutorial, I'm now thinking it should be within the theatre section and put as `Your First Animation` right under `Getting Started`

Side note:
I see [AriaMinaei](https://github.com/AriaMinaei) has been doing a bunch of contributing to private repos so I can only imagine that he's still cooking the next iteration of theatre.js. Does anyone have any word about how thats going?

ALPaseeno:
Thanks for the review — all four points make sense, and I'll get them in.

- Full turn: agreed. I'll fix the keyframes so the rotation completes a proper 360° and re-export the state.
- Doc links at the "how do I ___" moments: good idea. I'll add links to the theatre.js docs at those points, starting with adding keyframes.
- I'll update the theatre.js doc links to `latest`, including the one at the bottom of the Getting Started page.
- The placement makes sense too — I'll move it into the theatre section as "Your First Animation", right under Getting Started.

No word on my side about the next theatre.js iteration either.

Thanks again for the quick feedback — I'll push the changes shortly.

ALPaseeno:
All four are in:

- **Full turn** — found the cause: the rotation keyframe had been authored in radians (`6.283…`), but `<Transform>` exposes rotation to Theatre.js in degrees, so the box was only turning 6.28°. The final keyframes in both state files are now `360` and the box completes a proper full turn. I also scanned the rest of the state for other radian-scale values — those two were the only ones.
- **Doc links at the "how do I ___" moments** — added three: adding/removing keyframes (end of the keyframing paragraph), keyboard shortcuts (inside the "edits are saved locally" tip), and projects/state (where the export workflow is introduced).
- **0.5 → latest** — updated all nine links across four files, including the one at the bottom of Getting Started. One anchor changed between versions: `api/core#sheet.object` is `api/core#object` on latest, so that link points at the renamed section.
- **Moved and renamed** — the tutorial now lives in the theatre section as "Your First Animation", directly under Getting Started, and the intro note is reworded since Getting Started is now the page right above it.

Verified locally against all four points. Marking this ready for review.

DefinitelyMaybe:
nice. pretty much there. Your choice on the following:

- the AI has written "but we want to animate the box’s colour too" however none of the examples animate colour.

I've pushed some changes to the PR too so do a quick sync before continuing.

ALPaseeno:
Synced with your changes — thanks, the tighter wording and the tween editor mention both read well.

On the colour line: rather than softening the text, I went the other way and made the example live up to it — the animation now keyframes the box's colour too (orange → violet → orange across the same 4s loop as the rotation). It gives `<Sync>` a proper payoff in the final result, and the cube looks nicer doing it. Authored in the Studio and re-exported, so the rotation track is untouched.

DefinitelyMaybe (approved):
nice! I like it. good start 👍

PR #1841, opened 2026-07-27
docs: update asset URLs to new domain

DefinitelyMaybe:
doesn't currently work locally for some reason.


DefinitelyMaybe:
well, just redeployed via vercel again. i dont like it but this is no longer an issue.

PR #1843, opened 2026-08-25
docs: correct stale prop shapes in the rapier reference

### LINKED ISSUE #1842: [docs] rapier reference pages document removed `Position`/`Rotation` types and object shapes that no longer work
Several props in the `@threlte/rapier` reference are documented with types that were deleted from the codebase, and with object defaults that break the simulation if you pass them literally. Everything affected is a tuple now.

## Verified at runtime

These three are read positionally, so an object leaves every component `undefined` and the body ends up with non-finite transforms. It fails silently — the body simply disappears, fixed bodies stay put, and nothing is logged. I checked each one on its own.

| Page | Prop | Documented | Actual type |
| --- | --- | --- | --- |
| `world.mdx` | `gravity` | `Position`, default `{ y: -9.81 }` | `Parameters<Vector3['set']>` |
| `rigid-body.mdx` | `linearVelocity` | `Position`, default `{}` | `Parameters<Vector3['set']>` |
| `rigid-body.mdx` | `angularVelocity` | `Rotation`, default `{}` | `Parameters<Euler['set']>` |

## Verified by reading the source

| Page | What the docs say | What the source has |
| --- | --- | --- |
| `collider.mdx`, `auto-colliders.mdx` | a nested `massProperties` object with `centerOfMass: Position`, `principalAngularInertia: Position`, `angularInertiaLocalFrame: Rotation` | no `massProperties` prop exists. `mass`, `centerOfMass`, `principalAngularInertia` and `angularInertiaLocalFrame` are **flat top-level props** in a `MassDef` union, typed `Parameters<Vector3['set']>` / `Parameters<Euler['set']>` |
| `use-rope-joint.mdx`, `use-revolute-joint.mdx`, `use-fixed-joint.mdx`, `use-prismatic-joint.mdx`, `use-spherical-joint.mdx` | signature comments mark `anchorA` / `anchorB` as `// Position` and `axis` / `frameA` / `frameB` as `// Rotation` | `Parameters<Vector3['set']> \| Vector3` — a tuple or a `Vector3`, never a plain object |

## Why the docs say this

They were correct when they were written. `Position` was a real exported type that accepted exactly the documented shape:

```ts
export type Position =
  | Vector3
  | {
      x?: number
      y?: number
      z?: number
    }
```

Two commits moved the ground underneath them:

- **[`0a6e828f5`](https://github.com/threlte/threlte/commit/0a6e828f5de6d443dc6ea81a605b256c95d5c71e) — "Adapted Types" (2022-12-22)** switched these props to tuples and, in the same commit, dropped the `positionToVector3()` conversion in favour of raw index access:

  ```diff
  -  export let gravity: Props['gravity'] = { y: -9.81 }
  +  export let gravity: NonNullable<$$Props['gravity']> = [0, -9.81, 0]
  -    positionToVector3(gravity),
  +    { x: gravity[0], y: gravity[1], z: gravity[2] },
  ```

  That helper is why the object shape used to work, and why it now silently yields `undefined`.

- **[`771fd4c39`](https://github.com/threlte/threlte/commit/771fd4c397e0ef92bd69ad926c5754e4f3e21aa0) — "Remove custom Position, Rotation, LookAt types" (2023-04-23)** deleted the types, so `Position` and `Rotation` in the docs no longer refer to anything.

The reference pages were never re-synced after either change.

## Steps to reproduce

1. Add this as an example and open it:

```svelte
<script lang="ts">
  import { Canvas, T } from '@threlte/core'
  import { Collider, RigidBody, World } from '@threlte/rapier'
</script>

<Canvas>
  <!-- the shape world.mdx gives as the default -->
  <World gravity={{ y: -9.81 }}>
    <T.PerspectiveCamera
      makeDefault
      position={[6, 4, 8]}
      oncreate={(ref) => {
        ref.lookAt(0, 1, 0)
      }}
    />
    <T.DirectionalLight position={[5, 10, 5]} />
    <T.AmbientLight intensity={1} />

    <RigidBody type="fixed">
      <T.Group position.y={-0.5}>
        <Collider
          shape="cuboid"
          args={[10, 0.5, 10]}
        />
      </T.Group>
      <T.Mesh position.y={-0.5}>
        <T.BoxGeometry args={[20, 1, 20]} />
        <T.MeshStandardMaterial color="#4E5A46" />
      </T.Mesh>
    </RigidBody>

    <T.Group position={[0, 3, 0]}>
      <RigidBody>
        <Collider
          shape="cuboid"
          args={[0.5, 0.5, 0.5]}
        />
        <T.Mesh>
          <T.BoxGeometry args={[1, 1, 1]} />
          <T.MeshStandardMaterial color="#9C6B4E" />
        </T.Mesh>
      </RigidBody>
    </T.Group>
  </World>
</Canvas>
```

2. **Expected:** the box falls and comes to rest on the ground.
3. **Actual:** the box never appears. The ground renders normally and nothing is logged.
4. Change `gravity={{ y: -9.81 }}` to `gravity={[0, -9.81, 0]}` — the box falls and rests as expected.

`linearVelocity` and `angularVelocity` behave the same way. Sending `[2, 6, 0]` to `linearVelocity` launches the box as you'd expect; sending the identical value as `{ x: 2, y: 6, z: 0 }` makes `body.linvel()` read `NaN, NaN, NaN` and the box never appears.

Worth noting how far the corruption spreads in each case, since it is not identical. Broken `gravity` takes `linvel()` down with it — gravity is an acceleration applied to linear velocity each step — but leaves `angvel()` untouched, exactly as the physics would predict. Broken `angularVelocity` is narrower still: the body keeps falling normally and `linvel()` stays healthy, and only the orientation goes non-finite. The box disappears anyway, because the `NaN` quaternion poisons the mesh's world matrix, so the rendered position is `NaN` even though Rapier's translation is fine. Something vanishing therefore does not tell you which prop is at fault, which is what makes this slow to diagnose.

If it's useful I can include the A/B example I built while tracking this down in the fix PR, as a docs example: pick a prop, toggle between the tuple and its documented object shape, and it reads `world.gravity`, `linvel()` and `angvel()` back out of Rapier each frame and highlights whichever has gone non-finite — so all three props can be checked against their documented shapes at a glance.

## Suggested fix

If this is welcome I'd like to pick it up, and I'd cover every occurrence found in the audit above rather than just the gravity row:

- replace `Position` / `Rotation` with the real tuple types and correct the defaults on `world.mdx` and `rigid-body.mdx`
- correct `collider.mdx` and `auto-colliders.mdx` to show the flat mass props instead of a nested `massProperties` object
- fix the `// Position` / `// Rotation` signature comments on the five joint hook pages

I have a small example that toggles each prop between the documented shape and the real one, which I can include if that's useful for regression.

ALPaseeno:
Closes #1842.

Three commits, one per category of error:

- **Tuple props** — `gravity`, `linearVelocity` and `angularVelocity` are read positionally from an array. The documented `Position` / `Rotation` object shapes stopped working when those types were removed, and passing one yields `undefined` components with nothing logged.
- **`massProperties`** — there is no such prop. `MassDef` exposes `mass`, `centerOfMass`, `principalAngularInertia` and `angularInertiaLocalFrame` as flat, mutually exclusive props on both `<Collider>` and `<AutoColliders>`. The pretext on both pages is corrected to match.
- **Joint hooks** — anchors and frames documented as the tuple-or-instance unions they actually are. This also corrects `axis` on the revolute and prismatic pages, which was labelled `Rotation` but is a direction vector.

Everything was checked against the source rather than inferred. Prettier and the docs build are green.

I have a small A/B example that toggles each prop between its documented and real shape and reads the values back out of Rapier each frame — happy to add it as a docs example or regression aid if that's useful, but I've left it out to keep this PR to the correction itself.

### DIFF
--- a/apps/docs/src/content/reference/rapier/auto-colliders.mdx
+++ b/apps/docs/src/content/reference/rapier/auto-colliders.mdx
@@ -6,7 +6,7 @@ sourcePath: 'packages/rapier/src/lib/components/Colliders/AutoColliders/AutoColl
 type: 'component'
 componentSignature:
   {
-    pretext: "If a <code>&lt;AutoColliders&gt;</code> component is not a child of a <code>&lt;RigidBody&gt;</code> component, the transform properties are reactive.<br /><br />If you don't provide any of the properties <code>density</code>, <code>mass</code> or <code>massProperties</code>, Rapier will figure that out for you.<br /><br />You can provide either a property <code>density</code>, <code>mass</code> <strong>or</strong> <code>massProperties</code>.",
+    pretext: "If a <code>&lt;AutoColliders&gt;</code> component is not a child of a <code>&lt;RigidBody&gt;</code> component, the transform properties are reactive.<br /><br />If you don't provide any of the properties <code>density</code>, <code>mass</code> or the full mass properties, Rapier will figure that out for you.<br /><br />Provide <code>density</code> on its own, <code>mass</code> on its own, <strong>or</strong> <code>mass</code> together with <code>centerOfMass</code>, <code>principalAngularInertia</code> and <code>angularInertiaLocalFrame</code>.",
     props:
       [
         {
@@ -16,12 +16,9 @@ componentSignature:
         },
         { name: 'density', type: 'number' },
         { name: 'mass', type: 'number' },
-        { name: 'massProperties', type: '{
-            mass: number,
-            centerOfMass: Position,
-            principalAngularInertia: Position,
-            angularInertiaLocalFrame: Rotation,
-            }' },
+        { name: 'centerOfMass', type: '[number, number, number]' },
+        { name: 'principalAngularInertia', type: '[number, number, number]' },
+        { name: 'angularInertiaLocalFrame', type: '[number, number, number]' },
         { name: 'restitution', type: 'number' },
         { name: 'restitutionCombineRule', type: 'CoefficientCombineRule' },
         { name: 'friction', type: 'number' },
--- a/apps/docs/src/content/reference/rapier/collider.mdx
+++ b/apps/docs/src/content/reference/rapier/collider.mdx
@@ -6,7 +6,7 @@ title: '<Collider>'
 type: 'component'
 componentSignature:
   {
-    pretext: "If a <code>&lt;Collider&gt;</code> component is not a child of a <code>&lt;RigidBody&gt;</code> component, the transform properties are reactive.<br /><br />If you don't provide any of the properties <code>density</code>, <code>mass</code> or <code>massProperties</code>, Rapier will figure that out for you.<br /><br />You can provide either a property <code>density</code>, <code>mass</code> <strong>or</strong> <code>massProperties</code>.",
+    pretext: "If a <code>&lt;Collider&gt;</code> component is not a child of a <code>&lt;RigidBody&gt;</code> component, the transform properties are reactive.<br /><br />If you don't provide any of the properties <code>density</code>, <code>mass</code> or the full mass properties, Rapier will figure that out for you.<br /><br />Provide <code>density</code> on its own, <code>mass</code> on its own, <strong>or</strong> <code>mass</code> together with <code>centerOfMass</code>, <code>principalAngularInertia</code> and <code>angularInertiaLocalFrame</code>.",
     props:
       [
         {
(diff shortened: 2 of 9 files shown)


michealparks (commented):
Thank you!

michealparks (approved):

PR #1844, opened 2026-08-25
docs: add rapier physics tutorial

### LINKED ISSUE #548: Introductory tutorial: `@threlte/rapier`
Each package should have a solid introductory tutorial, analogous to `@threlte/core`'s [Your First Scene](https://threlte.xyz/docs/learn/getting-started/your-first-scene).

`@threlte/rapier` allows users to enable physics in their apps, and the tutorial should clarify the key elements needed to achieve this. The tutorial should have a coherent narrative: the learner should be making a scene that makes sense, and learning about `@threlte/rapier` as they build it.

If you consider contributing this, please discuss your plan here before submitting a PR.

ALPaseeno:
Closes #548. Opening as a draft so you can steer it, same as last time.

Six steps building one scene — a cannon and a brick wall — each a runnable `<Example>`, laid out as the plan in #548 describes. The spine is step 2: the reader's first physics code produces a spectacular failure, because the ground looks solid but was never given a body, and steps 3–5 are them fixing it.

The commits follow the tutorial's own order, so they can be reviewed as a walkthrough rather than as one diff.

On your two notes — I've extended `Where to go next` with a step that builds on the step 6 sensor toward game logic, pointing at `usePhysicsTask` for rules that depend on simulation state and at fixed framerate for repeatability. I also matched the bullet style to the theatre tutorial's version of that section.

Placed at `reference/rapier/` under Getting Started, mirroring where Your First Animation ended up. Prettier, `svelte-check` and the docs build are green.


apps/docs/src/content/reference/rapier/your-first-simulation.mdx
```diff
@@ -0,0 +1,246 @@
...
+simulation steps — it's on one side at the end of one step and the far side at the
+end of the next, having never touched anything. `restitution` is bounciness, `0`
+meaning none; `friction` lets the shot bite instead of skating; and the damping
+values bleed off the energy that a clipped corner would otherwise turn into a wild
+ricochet.
+
+<Example path="building-a-cannon/step-5" />
+
+Try setting `restitution` to `2` and firing again. Physics parameters are much
+easier to understand once you've abused them.
```
DefinitelyMaybe: Instead of "abused" I'd prefer something like "played with the extremes"

DefinitelyMaybe (commented):
Nice 👍 small bit of wording but good enough otherwise

apps/docs/src/content/reference/rapier/your-first-simulation.mdx
ALPaseeno: My bad, going with "played with the extremes", it's definitely the better fit here. That was gamer vocab leaking in: you abuse a Vandal in Valorant and it's a compliment. Doesn't travel into a tutorial though! 😄

apps/docs/src/examples/building-a-cannon/shared/Cannon.svelte
```diff
@@ -0,0 +1,150 @@
+<script lang="ts">
```
michealparks, line 1: The cannon is identical in all scenes, can we have one shared cannon component for all examples?

apps/docs/src/examples/building-a-cannon/step-3/Scene.svelte
```diff
@@ -0,0 +1,99 @@
...
+  import { Collider, Debug, RigidBody } from '@threlte/rapier'
+  import Cannon from './Cannon.svelte'
+
+  let { debug }: { debug: boolean } = $props()
+
+  const WALL_X = 7
+  const ROWS = 4
+  const COLUMNS = 3
+
+  const bricks = Array.from({ length: COLUMNS * ROWS }, (_, i) => {
```
michealparks: Same with bricks, seems like something that we could wrap into a shared component

apps/docs/src/examples/building-a-cannon/step-4/App.svelte
```diff
@@ -0,0 +1,70 @@
...
+    label="Aim"
+    min={0}
+    max={85}
+    step={1}
+  />
+  <Slider
+    bind:value={power}
+    label="Power"
+    min={2}
+    max={60}
```
michealparks: can we clamp power from 0.1-2? anything stronger is almost imperceptible

apps/docs/src/examples/building-a-cannon/step-4/Scene.svelte
```diff
@@ -0,0 +1,156 @@
...
+<T.DirectionalLight
+  castShadow
+  intensity={2}
+  position={[8, 20, 6]}
+  shadow.camera.top={20}
+  shadow.camera.bottom={-20}
+  shadow.camera.left={-20}
+  shadow.camera.right={20}
+  shadow.mapSize.width={2048}
+  shadow.mapSize.height={2048}
```
michealparks: I'd scale the map size down to 1024 since there are multiple scenes on this page

apps/docs/src/examples/building-a-cannon/step-1/Scene.svelte
```diff
@@ -0,0 +1,72 @@
...
+    }
+  })
+</script>
+
+<T.PerspectiveCamera
+  makeDefault
+  position={[-9, 5, 6]}
+  fov={45}
+>
+  <OrbitControls
```
michealparks, line 15: Can we add `enableZoom={false}` to all the OrbitControls? that'll allow users to more easily scroll through these examples.

michealparks (commented):
Cool example! A few preliminary thoughts

apps/docs/src/examples/building-a-cannon/shared/Cannon.svelte
ALPaseeno: Done. The catch was that the example loader only bundles files inside each step's own directory, so a ../shared import would vanish from the code tabs and break the StackBlitz export — that's also why the theatre and first-scene tutorials duplicate their scenes per step. I added shared/ folder support to Example.astro (StackBlitz path mapping handled), so there's now one shared/Cannon.svelte for all six steps. The fixed body around it stays in each step's scene from step 3 on, so the lesson is still visible where it's taught.

apps/docs/src/examples/building-a-cannon/step-3/Scene.svelte
ALPaseeno: The wall layout now lives in shared/wall.ts, one createWall() used by all six steps, parameterized so step 6 can still take rows/columns as props. I kept the per-brick markup in each scene though: that's the part that actually changes per step (no physics in step 1, colliders in step 2, userData tagging in step 6), so wrapping it up would hide the thing each step is teaching. Also moved the cannon's pivot geometry into shared/cannon.ts while I was in there, so the muzzle spawn can't drift from the barrel length anymore.

apps/docs/src/examples/building-a-cannon/step-4/Scene.svelte
ALPaseeno: Done, 1024 on all six scenes.

apps/docs/src/examples/building-a-cannon/step-1/Scene.svelte
ALPaseeno: Done, added to all six. Scrolling through the page is much nicer for it.

apps/docs/src/examples/building-a-cannon/step-4/App.svelte
ALPaseeno: Went with a multiplier on a tuned base impulse rather than a raw clamp, raw 0.1–2 left the shot too weak to reach the wall at all. It ended up clamped 0.1–1 rather than 0.1–2: with the base tuned so a full-power shot arrives with real momentum, anything past 1 crossed too fast to actually watch. Where the shot lands matters now, low punches a hole through the wall, high topples it, and the ricochet and tunnelling chaos stays in step 4 on purpose, since that's exactly what step 5 fixes. I think that's also why extra power read as imperceptible before.

ALPaseeno: Also added space to fire and arrow keys to aim while I was in there, mentioned in the prose as `<kbd>` keycaps. `markdown.css` had no rule for the element, so I added one matching the keycap treatment the pachinko example's HUD already uses.

apps/docs/src/examples/building-a-cannon/shared/Cannon.svelte
DefinitelyMaybe: Is this what you were meaning @michealparks . I'm just not sure about changing the Examples component for it.

apps/docs/src/examples/building-a-cannon/step-1/Scene.svelte
DefinitelyMaybe: it's also the camera orbital rotation that makes these hard to scroll past for me or have i been doing it wrong :sweat_smile:  (This is also a issue for many other examples but thats a separate issue)

If not, choosing some fixed camera angle that looks good enough on both mobile and desktop wouldn't be too hard imo

PR #1852, opened 2026-08-30
[@threlte/extras] Add WebGPU export

michealparks:
### Overview

Adds a `@threlte/extras/webgpu` export. 

Begins the migration work by porting `MeshDiscardMaterial` and `Grid` to TSL. 

Grid example: https://threlte.xyz/previews/pr-1852/docs/reference/extras/grid/

Screenshot 2026-08-30 at 18 44 45


.changeset/hungry-poems-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@threlte/extras": minor
+---
+
+[@threlte/extras] Add a `@threlte/extras/webgpu` entrypoint that omits GLSL-based exports
```
DefinitelyMaybe, line 5: `GLSL-based exports`? TSL based exports right? 😅

apps/docs/src/content/reference/extras/grid.mdx
```diff
@@ -58,7 +59,14 @@ componentSignature:
...
 
-<Example path="extras/grid" />
+<Tabs>
+  <Tab name="WebGL">
+    <Example path="extras/grid" />
+  </Tab>
+  <Tab name="WebGPU">
+    <Example path="extras/grid-webgpu" />
+  </Tab>
+</Tabs>
```
DefinitelyMaybe, line 69: yeah so we could avoid this by creating a `<Tip type="info">` at the top of the page that says something like :

"this component is has the equivilent webgpu export from `@threlte/extras/webgpu`"

apps/docs/src/pages/docs/reference/[...slug]/index.astro
```diff
@@ -32,9 +33,21 @@ const { Content } = await render(entry)
...
+      {title}
+    </h1>
+
+    {
+      webgpu && (
+        <Pill
+          label="webgpu"
+          href={`${import.meta.env.BASE_URL}docs/learn/advanced/webgpu`}
+          tooltip={`${title} works with WebGPU and is available from the ${category}/webgpu export.`}
+        />
```
Comment (DefinitelyMaybe) [inline apps/docs/src/pages/docs/reference/[...slug]/index.astro:47]: ah, lol here we go, you were thinking pretty much the same thing.

apps/docs/src/content/reference/extras/grid.mdx
DefinitelyMaybe: if the example stays the same, why create a new one?

Screenshot_20260902_155136

apps/docs/src/pages/docs/reference/[...slug]/index.astro
Comment (DefinitelyMaybe) [reply apps/docs/src/pages/docs/reference/[...slug]/index.astro:47]: oh... I like that we're pretty much on the same page here, I was thinking that it'd be less of a tooltip popup and more of a usual `<Tip>` component.

the webgpu Pill gives me the impression that its only for webgpu and it's not until I hover the pill that I realize thats not whats being communicated.

Might be just me, no?

.changeset/hungry-poems-shake.md
michealparks: `omits` is the keyword, but i'll make it more clear

DefinitelyMaybe: omgosh. yup, missed that somehow :sweat_smile:

apps/docs/src/pages/docs/reference/[...slug]/index.astro
Comment (michealparks) [reply apps/docs/src/pages/docs/reference/[...slug]/index.astro:47]: I was on the fence too, but i think most people can infer that this just means "now supported in webgpu renderer".

apps/docs/src/content/reference/extras/grid.mdx
michealparks: I think it's useful 1. to show people how to use the new exports and 2. to compare the two scenes to avoid visual regressions

DefinitelyMaybe: ok, I can 100% get behind point number 2. Having a visual check within the docs is a much better idea than manually creating a setup to check.

apps/docs/src/pages/docs/reference/[...slug]/index.astro
Comment (DefinitelyMaybe) [reply apps/docs/src/pages/docs/reference/[...slug]/index.astro:47]: I guess. Having the combination of both the pill and the tabbed examples might be enough.

apps/docs/src/components/Tabs/Tabs.astro
```diff
@@ -0,0 +1,7 @@
+---
+import TabsIsland from './Tabs.svelte'
+---
+
+<TabsIsland client:load>
+  <slot />
+</TabsIsland>
```
DefinitelyMaybe, line 7: oh, other question about the tabs, are we doubling up on running examples within the browser? if so anything we could do regarding performance there or its just gonna be what its gonna be?

apps/docs/src/content/reference/extras/webgpu.mdx
```diff
@@ -0,0 +1,67 @@
...
+  makeDefault
+  position={[10, 10, 10]}
+>
+  <OrbitControls />
+</T.PerspectiveCamera>
+
+<Grid />
+```
+
+Mixing the two entrypoints is what to avoid. A component imported from
```
DefinitelyMaybe, line 27:
```suggestion
Avoid mixing the two entrypoints. A component imported from
```

apps/docs/src/components/Tabs/Tabs.astro
michealparks: Only the active example runs now!

